PR #521, opened 2015-01-07
Adds error tracking through analytics to static

roc:
- We don't track frontend javascript errors at all at the moment, this should be remedied quickly.
- There are about a thousand ways to slice this issue, and there are lots of factors that complicate the way we deliver tracking (do we run our own error server? do we pay for a service? What about ajax events?) but a quick bootstrap solution would be to use google analytics, as we are already using it to track non-error events.
- Not sure what support < ie8 is like for error events, happy to take some advice here

### DIFF
--- a/app/assets/javascripts/analytics/error-tracking.js
+++ b/app/assets/javascripts/analytics/error-tracking.js
@@ -1,0 +1,24 @@
+// Extension to track errors using google analytics as a data store.
+(function() {
+
+    "use strict";
+    var trackJavaScriptError = function (e) {
+      var errorSource = e.filename + ': ' + e.lineno;
+      _gaq.push([
+        '_trackEvent',
+        'JavaScript Error',
+        e.message,
+        errorSource,
+        true //nonInteractive so bounce rate isn't affected
+      ]);
+    };
+
+    if (window.addEventListener) {
+      window.addEventListener('error', trackJavaScriptError, false);
+    } else if (window.attachEvent) {
+      window.attachEvent('onerror', trackJavaScriptError);
+    } else {
+      window.onerror = trackJavaScriptError;
+    }
+
+}());
--- a/app/assets/javascripts/analytics/error-tracking.js
+++ b/app/assets/javascripts/analytics/error-tracking.js
@@ -9,7 +9,8 @@
         'JavaScript Error',
         e.message,
         errorSource,
-        true //nonInteractive so bounce rate isn't affected
+        1, // Set our value to 1, though we could look to tally session errors here
+        true // nonInteractive so bounce rate isn't affected
       ]);
     };
 
--- a/app/assets/javascripts/analytics/error-tracking.js
+++ b/app/assets/javascripts/analytics/error-tracking.js
@@ -1,25 +1,25 @@
 // Extension to track errors using google analytics as a data store.
 (function() {
 
-    "use strict";
-    var trackJavaScriptError = function (e) {
-      var errorSource = e.filename + ': ' + e.lineno;
-      _gaq.push([
-        '_trackEvent',
-        'JavaScript Error',
-        e.message,
-        errorSource,
-        1, // Set our value to 1, though we could look to tally session errors here
-        true // nonInteractive so bounce rate isn't affected
-      ]);
-    };
+  "use strict";
+  var trackJavaScriptError = function (e) {
+    var errorSource = e.filename + ': ' + e.lineno;
+    _gaq.push([
+      '_trackEvent',
+      'JavaScript Error',
+      e.message,
+      errorSource,
+      1, // Set our value to 1, though we could look to tally session errors here
+      true // nonInteractive so bounce rate isn't affected
+    ]);
+  };
 
-    if (window.addEventListener) {
-      window.addEventListener('error', trackJavaScriptError, false);
-    } else if (window.attachEvent) {
-      window.attachEvent('onerror', trackJavaScriptError);
-    } else {
-      window.onerror = trackJavaScriptError;
-    }
+  if (window.addEventListener) {
+    window.addEventListener('error', trackJavaScriptError, false);
+  } else if (window.attachEvent) {
+    window.attachEvent('onerror', trackJavaScriptError);
+  } else {
+    window.onerror = trackJavaScriptError;
+  }
 
 }());
--- a/app/assets/javascripts/analytics.js
+++ b/app/assets/javascripts/analytics.js
@@ -4,3 +4,4 @@
 //= require analytics/print-tracking
 //= require analytics/print-intent
 //= require analytics/scroll-tracker
+//= require analytics/error-tracking


app/assets/javascripts/analytics/error-tracking.js
```diff
@@ -0,0 +1,24 @@
...
+
+    "use strict";
+    var trackJavaScriptError = function (e) {
+      var errorSource = e.filename + ': ' + e.lineno;
+      _gaq.push([
+        '_trackEvent',
+        'JavaScript Error',
+        e.message,
+        errorSource,
+        true //nonInteractive so bounce rate isn't affected
```
rboulton: You're missing the 4th event parameter, which is an integer "value" (https://developers.google.com/analytics/devguides/collection/gajs/eventTrackerGuide#Values).  The nonInteractive flag comes after that.

Elsewhere we've just set this to 1: eg https://github.com/alphagov/static/blob/master/app/assets/javascripts/browser-check.js#L23

app/assets/javascripts/analytics/error-tracking.js
```diff
@@ -0,0 +1,24 @@
+// Extension to track errors using google analytics as a data store.
+(function() {
+
+    "use strict";
+    var trackJavaScriptError = function (e) {
+      var errorSource = e.filename + ': ' + e.lineno;
+      _gaq.push([
+        '_trackEvent',
+        'JavaScript Error',
```
rboulton: Might be more consistent with other common events we have if this was 'javascript_error'.

That said, we've not got much consistency in event names: most use _ instead of space, though some use - and a few do use spaces.  Some are all lower-case, others are TitleCase.

For comparison some of our existing events are:

 - 'MS_guide'
 - 'IG_detailed_guidance'
 - 'Print Intent'
 - 'ScrollTo'
 - 'user_satisfaction_survey'
 - 'startpages'
 - 'multivariatetest_cohort_tax-disc-beta'

This is definitely a non-blocking comment, though. It would be good to set out some guidelines for event names in future (perhaps during the transition to universal analytics).

rboulton:
I think this is a great idea and will help us massively.

dsingleton:
👍 for the idea, huge step in the right direction

I there is a maximum number of events/views we can track in GA in a month,
and this could add a lot, with a risk of going over in some circumstances

You might want to start with a sampled version and possibly chat with
someone like John.

On Wednesday, January 7, 2015, Richard Boulton <notifications@github.com>
wrote:

> I think this is a great idea and will help us massively.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/alphagov/static/pull/521#issuecomment-69108023>.
>


-- 
*David Singleton*

Cabinet Office, Government Digital Service
Technical Lead, User Formats Team
david.singleton@digital.cabinet-office.gov.uk

app/assets/javascripts/analytics/error-tracking.js
roc: Cool, I'm the one responsible for 'Print Intent' so I guess this is my second motion for Title Cased Names

bradleywright:
Not sure we really want to put errors into Analytics. It's accessed by a lot more people than just GDS/GOV.UK staff, and I'm concerned about the impression given out by errors appearing in that system.

app/assets/javascripts/analytics/error-tracking.js
roc: Good spot, I think the universal analytics syntax half crept in whilst I was writing this.

roc:
@bradleywright yeah I thought this might come up as an issue. I want to take a sort of cottaged moral view and say 'well surfacing those errors is exactly what we should be doing to keep us honest' but I totally get the point of drawing unnecessary fire.

At the moment we're almost completely in the dark though, could we run it for a number of days or to a select group and just see what we end up with? Communicating to arms length GA account users that we're running an experiment to help us tune the site shouldn't be too hard, right? ;-)

rboulton:
One thing to note is that this will only add "events" to analytics, and I strongly suspect most account users won't look at events at all.  ie, while these errors would be visible if an arms length GA account user looked for them, I doubt most would notice their existence, since they're in a separate part of the GA user interface.

app/assets/javascripts/analytics/error-tracking.js
annapowellsmith: Just ran this past Ash, he's happy with both the event name (title cased or not) and with experimenting with the idea in principle.

annapowellsmith:
Re @dsingleton's comment, just spoke to Ash, he thinks we are unlikely to hit the limit (it's a billion per month). But if we let him know when this goes live then he can monitor the rate at which these arrive.

dsingleton:
> At the moment we're almost completely in the dark though, could we run it for a number of days or to a select group and just see what we end up with?

:+1: to this. At the least we should collect a sample, to even a hint of whats going on at the moment, which might help us decide how to capture errors in a more appropriate way.

dsingleton:
> Re @dsingleton's comment, just spoke to Ash, he thinks we are unlikely to hit the limit (it's a billion per month). But if we let him know when this goes live then he can monitor the rate at which these arrive.

It's a high limit, but I don't think it's infeasible for some pathological situation to flood GA with errors, but if we're doing a sample (either by time or %) then I'm not too worried.

roc:
So to summarise:
- Probably no issues with rate limits (1 billion per month (assuming it's 1,000,000,000 here still p good))
- Annexed to an event layer within the analytics UI, so mitigates bad press issues
- Running it for a short while should help us with our implementation reckons

# ... ?

rboulton:
I think we should limit it to one javascript error per page view.  That way, we can at most double the number of events we're sending (and we can currently afford to much more than double the rate).

app/assets/javascripts/analytics/error-tracking.js
```diff
@@ -0,0 +1,25 @@
...
+    };
+
+    if (window.addEventListener) {
+      window.addEventListener('error', trackJavaScriptError, false);
+    } else if (window.attachEvent) {
+      window.attachEvent('onerror', trackJavaScriptError);
+    } else {
+      window.onerror = trackJavaScriptError;
+    }
+
```
edds: This method seems to be double indented. Mind removing one layer of indents so that it is only 2 space indents.

edds:
> Not sure what support < ie8 is like for error events, happy to take some advice here

Without wrapping all our code in try catch blocks it is mostly impossible to catch errors there. With the current browser stats I don't think it is needed earlier.

We are all working on reckons with the events at the moment. If we are worried about using our quota I vote for merging this and watching the real time events once deployed. We should be able to roll back pretty fast if we see it causing problems.

app/assets/javascripts/analytics/error-tracking.js
roc: Oh yeah! 
:+1:

bradleywright:
As long as we're all on the understanding that this is a fixed term thing only I'm happy to merge this. We need a better plan for logging these things soon though.

roc:
Sweet

PR #522, opened 2015-01-08
Remove old analytics JavaScript

edds:
This analytics was added for the performance platform to create an
engagement score of pages. They have since stopped using these events
and they are no longer needed. Remove them so they stop sending lots of
events to Google Analytics.

Tidy up the function in `tracking.js` which is still needed so that it
is clearer as to what is going on.

--- 

This should enable us to remove GDS_successEvents and GDS_analyticsTokens from the cookie page as there will no longer be any code setting or getting those cookies.


bradleywright:
So much :red_circle:

alexmuller:
This is brilliant, thanks @edds. We'll provide :cake: at some point.

alexmuller:
> This should enable us to remove GDS_successEvents and GDS_analyticsTokens from the cookie page

I've opened a [ticket](https://govuk.zendesk.com/tickets/1059762) to do that because the expiration date for those cookies has passed.

fofr:
:heart: :+1:

PR #525, opened 2015-01-08
Use jasmine rails

alicebartlett:
Came to this repo to add tests for a new component but found that the tests were running but then erroring and I was completely unable to run them in browser.

This PR updates the JS test suite to use rails-jasmine rather than the jasmine gem. The tests can now be run by going to /specs or using `bundle exec rake spec:javascript` or `bundle exec rake` or `bundle exec rake tests` depending on how much of the test suite you want to run.

I hope this is an improvement, using jasmine-rails is the way all of the newer GOV.UK repos use jasmine.


fofr:
The admin tools are mostly using the official jasmine gem for tests, rather than jasmine-rails but there is only one clear reason I can think of for this. In the jasmine gem there's a configuration option that prevents jasmine from automatically downloading and installing a version of PhantomJS (by the phantomjs gem). @alext originally highlighted the automatic install as a cause for concern in a managed environment and we patched the jasmine gem to make it configurable.

For more context, here's the issue I raised on jasmine-rails: https://github.com/searls/jasmine-rails/issues/133 and the PR to fix this issue on the jasmine gem: https://github.com/jasmine/jasmine-gem/pull/213 (I fixed it here as that was the gem being used at the time)

Here is where static is preventing the download:
https://github.com/alphagov/static/blob/master/spec/javascripts/support/jasmine_helper.rb

alicebartlett:
To summarise the conversation @fofr and I just in meatspace:

There are three options here
1) Revert the change, things are still a bit broken but phantomjs isn't being downloaded automatically.
2) Keep the change, live with the problem being re-introduced. @fofr and I aren't really sure if this is still as bigger problem as it was when we fixed it last time. This is something we'd like to talk to @alext about.
3) Keep the change and patch jasmine-rails to not automatically download phantomjs. @fofr would quite like to do this as a community spirited patch for rails-jasmine users as well as fixing this for us (here and in the other places we're using jasmine-rails)

For now, @fofr is going to spend 20 minutes looking at how difficult patching jasmine-rails would be
@alext if you have any opinions about option 2 here then please share them.
I'm no longer blocked and the tests I'm going to write now for my new component will work with jasmine or jasmine-rails so I'll continue as I was and leave it up to you to merge or not merge.

edds:
The branch was accepted and made it into `jasmine-rails 0.10.6` so this doesn't need to point to a github branch anymore.

PR #527, opened 2015-01-09
Add an alpha label component

edds:
The same as the beta, but alpha. For things that are newer and haven't
yet proven their worth.

screen shot 2015-01-09 at 15 54 48


app/views/govuk_component/docs.yml
```diff
@@ -1,3 +1,10 @@
+- id: "alpha_label"
+  name: "Alpha Banner"
+  description: "A banner than indicates content is in a alpha phase with an optional explanation"
```
rjc123: 'an alpha'

PR #529, opened 2015-01-12
Option select component

alicebartlett:
This PR introduces the scrollable list format for checkboxes (now renamed "option-select"). The option-select is used on finders (http://github.com/alphagov/finder-frontend), site search (https://www.gov.uk/search?q=tax) and in the Digital Marketplace. I have mostly ported the code from http://github.com/alphagov/finder-frontend but have taken the opportunity to improve the code (with hindsight) as I moved it.

**About this component**
There are three features that are turned on/off depending on the user's setup:

| Feature       | no js | <ie8  | mob/tab | other
| --------------- |:-------:|-------:|------------:|-------
| scrollable    | ✓      | ✓      | ✗            | ✓    |
| clear link     | ✗       | ✓      | ✓           | ✓    |
| collapsable | ✗       | ✗       | ✓           | ✓    |   

No JavaScript (doesn't open/close):
screen shot 2015-01-12 at 18 22 55

IE6/7 (not collapsable)
ie7

Mobile / tablet (no scroll)
screen shot 2015-01-12 at 18 48 55

Fully featured (scrollable/clear link/collapsable)
screen shot 2015-01-12 at 18 26 41

Here is a GIF of the fully featured thing working:
pull_request_overkill

Here is the component in the components guide (https://github.com/alphagov/govuk-component-guide)
screen shot 2015-01-23 at 14 10 29

**Commit structure of this pr**
The commits work through building the component from scratch. An un-enhanced version first, then a commit for each JS enhancement (complete with tests and SCSS additions).

**Feedback**
I haven't used this in anger yet, so it's likely when I come to replace http://github.com/alphagov/finder-frontend with it I'll find loads of embarrassing mistakes, but I wanted to get feedback on it before I undertook that piece of work.

I'm interested in the following from reviewers:
- As this code is going to be used (hopefully!) by quite a few people, is it easy to read and understand?
- Is the SCSS too specific in any areas? I tried not to use super specific selectors so they could be easily overridden by implementors but I've never written a component like this before so its likely I haven't got it totally right.
- In the commit about keyboard accessibility(#20fb54a) I discuss a design compromise between keyboard users and screen reader users (though many users fall into both camps). Can you think of a better way to do this?
- Other smaller bits of feedback too, obviously.


app/assets/javascripts/govuk-component/option-select.js
```diff
@@ -0,0 +1,137 @@
...
+    return this.$optionSelect.find('.js-clear-selected');
+  };
+
+  OptionSelect.prototype.resetOptions = function resetOptions(){
+    this.$options.prop({
+      indeterminate: false,
+      "checked": false
+    }).trigger("change");
+    this.$clearingLink.addClass('js-hidden');
+    return false;
```
edds: Is there a need for this event to not bubble? If so can you add a comment explaining why, otherwise rather returning false on the event can this call `event.preventDefault()`. That will let us add other listeners (like analytics for example).

app/assets/javascripts/govuk-component/option-select.js
```diff
@@ -0,0 +1,137 @@
+(function ($) {
+  "use strict";
+  window.GOVUK = window.GOVUK || {};
+
+  function OptionSelect(options){
+    /* This JavaScript provides two fucntional enhancements to option-select components:
```
edds: s/fucntional/functional/

app/assets/javascripts/govuk-component/option-select.js
```diff
@@ -0,0 +1,137 @@
...
+
+  OptionSelect.prototype.close = function close(){
+    this.$optionSelect.addClass('js-closed');
+  };
+
+  OptionSelect.prototype.isClosed = function isClosed(){
+    return this.$optionSelect.hasClass('js-closed');
+  };
+
+  OptionSelect.prototype.ensureOptionSelectIsOpen = function ensureOptionSelectIsOpen(){
```
edds: Is it worth adding the `if (this.isClosed()) {` to the contents of `open` and removing this method? It would make for a simpler API and anything wanting to ensure the OptionSelect is open can just call open and not have to worry about any side effects.

app/assets/javascripts/govuk-component/option-select.js
```diff
@@ -0,0 +1,137 @@
...
+    this.$clearingLink = this.attachClearingLink();
+    this.updateClearingLink();
+
+    // Attach event listeners for clearing
+    this.$clearingLink.on('click', this.resetOptions.bind(this));
+    this.$options.on('click', this.updateClearingLink.bind(this));
+
+    // Performance in ie 6/7 is not good enough to support animating the opening/closing
+    // so do not allow option-selects to be collapsible in this case
+    var allowCollapsible = (typeof ieVersion == "undefined" || ieVersion > 7) ? true : false;
```
edds: We haven't defined `ieVersion` globally only on an app by app basis. If you want to use it in this I would suggest adding them to [the template slimmer uses for all pages](https://github.com/alphagov/static/blob/master/app/views/root/_stylesheet.html.erb) in the [same way they have been added to finders](https://github.com/alphagov/finder-frontend/blob/master/app/views/layouts/application.html.erb#L6-L8)

edds:
The code to me looks really good and I didn't struggle to understand it at all. :+1: Thank you for getting this ready to go into components!

One thing I would say you need to look at is your the reference Sass selectors. From our styleguide ["Always define parent reference selectors before defining child selectors"](https://github.com/alphagov/styleguides/blob/master/css.md#sass-nesting). There are two cases where you have got a non-reference selector before a reference selector and trying to work out what the selector will compile to is a little complicated.

> In the commit about keyboard accessibility (20fb54a) I discuss a design compromise between keyboard users and screen reader users (though many users fall into both camps). Can you think of a better way to do this?

Have you asked Leonie about how she expects to use the dropdowns? Based entirely on reckons (and slightly informed by the way that accessible tabs work for keyboard users) I would have thought that hiding the contents which isn't visible would have been the most appropriate. That would make pages with lots of filters still usable by keyboard users. As at the moment if there are lots of options it will be very slow to use these pages if you have to tab through every input. But as I say that is based on reckons without having used it so feel free to ignore it.

Would it be possible to include a screenshot of the component guide app for this page to the pull request please.

Also A+++ on the Git and GitHub work. Good commits and pull request.

app/assets/javascripts/govuk-component/option-select.js
```diff
@@ -0,0 +1,137 @@
...
+    if(allowCollapsible){
+
+      // Add js-collapsible class to parent for CSS
+      this.$optionSelect.addClass('js-collapsible');
+
+      // Add the little arrow toggle image
+      this.attachOpenCloseToggle();
+
+      // Add open/close listeners
+      this.$optionSelect.find('.container-head').on('click', this.toggleOptionSelect.bind(this));
```
jonnywyatt: Is it safer to reference a js- class here? https://github.com/alphagov/styleguides/blob/master/js.md#html-class-hooks

app/assets/javascripts/govuk-component/option-select.js
```diff
@@ -0,0 +1,137 @@
...
+
+      // Add a listener to the checkboxes so if you navigate to them with the keyboard you can definitely see them
+      this.$options.on('focus', this.ensureOptionSelectIsOpen.bind(this));
+
+      this.close();
+    }
+  }
+
+  OptionSelect.prototype.attachClearingLink = function attachClearingLink(){
+    this.$optionSelect.find('.container-head').append('<a class="js-clear-selected">clear</a>');
```
jonnywyatt: I can't see this covered in the JS styleguide but is a <button> more appropriate than a link? eg. do your users expect to activate control buttons with space or enter.

edds: @jonnywyatt: things should look like what they are. As it looks like a link it should be a link. Otherwise assistive technology won't pick it up when you try and find the links saying clear.

spec/javascripts/govuk-component/option-select-spec.js
```diff
@@ -0,0 +1,281 @@
+describe('GOVUK.OptionSelect', function() {
+
+  var $optionSelectHTML, optionSelect;
+
+  beforeEach(function(){
+
+    optionSelectFixture = '<div class="govuk-option-select" tabindex="0">'+
```
jonnywyatt: This is probably outside the scope of this PR but could the HTML be loaded from an external file? A step beyond that could be to generate it by putting some data into app/views/govuk_component/option_select.raw.html.erb
Just a suggestion, feel free to ignore.. :)

app/assets/javascripts/govuk-component/option-select.js
jonnywyatt: @edds - that's a very good point

joelanman:
very nice, tiny thing: do we standardise on whitespace in `if` statements?

`if(allowCollapsible){`

`if (this.isClosed()) {`

tombye:
Brilliant pull request, especially in baking so much detail into the commit messages.

Some mostly un-evidenced opinion (reckon) for the question about keyboard accessibility:

I think the main question is whether screenreader users benefit from being able to limit the number of inputs onscreen. 

Visual users not using a keyboard currently get value from being able to control the number of inputs they see onscreen at once and therefore the cognitive load of performing a search. Hiding the contents would pass this benefit on to keyboard users.

I've been told (in an accessibility review) that screenreader users can also benefit from being able to filter the amount in the page. It is definitely something to ask Léonie about though as it's very much about what is and isn't important if you use a screenreader all the time.

edds:
@joelanman: no we aren't strict on whitespace around `if` statements. We haven't found that having different whitespace decreases comprehension or understanding. There is also little point in stating a preference unless there is a test suite enforcing it as, from experience, it will be haphazardly followed.

bradleywright:
JSLint would complain about that normally if `white` isn't turned on  ('Tolerate messy whitespace')

dsingleton:
> Is the SCSS too specific in any areas? I tried not to use super specific selectors so they could be easily overridden by implementors but I've never written a component like this before so its likely I haven't got it totally right.

Just chatted about this with @alicebartlett offline, summary:

We shouldn't worry about avoiding super specific selectors to enable implementors (apps calling components), as implementors should be treating components as a black-block, and never hanging their own styling (or behaviour) off the markup of existing styling of a component. 

Apps calling components should work on the assumption that the implementation may change at any time, without an update to that app itself. For example:
- Internal markup changing and invalidating selectors, eg, `span` to `strong`
- Expectations of what styles do don't need to be overridden

If apps calling components are closely coupled to the implementation of the components then it makes iterating them without breaking things a lot harder, potentially having to deploy many apps at the same time to make changes to a component. This is the world we're in now with poorly-scoped selectors on `static`, so we're keen to avoid that :)

In a situation where a calling app does want to override a component the suggested way (we've not had a real example of this yet) would be to add a parameter to the component for the different styling/behaviour, forcing it to be standardised. This might be awkward in some situations (we'll see), but it reinforces the idea that components should be entirely standardised and there needs to be a good reason for different behaviour in a single app.

That said, having non-super specific selectors is still good practice, and makes developing the component itself easier in the future :+1: 

(I hope that makes sense to everyone else)

app/assets/javascripts/govuk-component/option-select.js
alicebartlett: Hmm this comment looks familiar ^_^
https://github.com/alphagov/finder-frontend/pull/12#discussion-diff-10792502 

So basically return false is here to stop the event bubbling up because there is a listener on the parent container to open/close the option-select. You're right about the comment thing though, if I'd added that in the first place you wouldn't be repeating yourself :)

dsingleton:
Rest of the PR looks :+1: to me. 

Running it locally against the component-guide also looks good - though I think the compontent-guide will need some improving to handle bigger/more complex components like this in the future.

app/assets/javascripts/govuk-component/option-select.js
alicebartlett: Do you mean instead of `.container-head` ?

spec/javascripts/govuk-component/option-select-spec.js
alicebartlett: Initially I had the fixtures in a separate file being loaded in, but currently chrome blocks file:// URIs from reading other file:// URIs, which means they can't be loaded in chrome unless you switch --allow-file-access-from-files on. (more on this here: https://github.com/velesin/jasmine-jquery#cross-domain-policy-problems-under-chrome) Introducing this felt like a major gotcha for developers using chrome coming to the project so I put the html in the test to avoid future confusion.

jonnywyatt: Fair enough. A framework like Karma can make this easy to do eg will load all HTML fixtures and make them available to the tests - https://github.com/karma-runner/karma-html2js-preprocessor
Although you may not want to add another component to your stack..

app/assets/javascripts/govuk-component/option-select.js
jonnywyatt: Yes

app/assets/javascripts/govuk-component/option-select.js
jonnywyatt: @edds, thought a bit more about this - agreed, if it looks like a link it should be a link... but should it look like a link at all if it works like a button? ie. it's not taking you anywhere
Again, maybe a design issue and outside the scope of this PR

app/assets/javascripts/govuk-component/option-select.js
alicebartlett: OK cool. Yeah (having thought about it) I think you're right. I didn't originally because I read that original bit in the style guide as "if you add a class with javascript, or if your class is only used for javascript stuff then make it js-" but I see now that wasn't quite the right interpretation.

jonnywyatt: The test I'd apply is, if a developer thinks about changing or removing that class for styling purposes, can they know for certain they won't break any Javascript. So any hook for Javascript should probably be separate and prefixed with js-

alicebartlett: That's a good way around to think about it, thanks.

app/assets/javascripts/govuk-component/option-select.js
alicebartlett: Thanks for explaining this so clearly edds. I've made the change as part of this pull request.

spec/javascripts/govuk-component/option-select-spec.js
alicebartlett: Ah cool, hadn't heard of Karma, I'll check it out. It might be worth adding this in if future work involves a lot more Javascript testing in this way, but for now I'm going to leave the tests as they are.

app/assets/javascripts/govuk-component/option-select.js
```diff
@@ -0,0 +1,136 @@
...
+  "use strict";
+  window.GOVUK = window.GOVUK || {};
+
+  function OptionSelect(options){
+    /* This JavaScript provides two functional enhancements to option-select components:
+      1) A link that clears all of the checkboxes (referred to here as the "clearing link")
+      2) Open/closing of the list of checkboxes - this is not provided for ie6 and 7 as the performance is too janky.
+    */
+
+    this.$optionSelect = options.el;
```
joelanman: since the prefix of $ means the variable is a jQuery object, shouldn't `options.el` be `options.$el`? And obviously the same naming when it's called on line 134

alicebartlett:
# Thank you!
OK! Finally gotten around to addressing all of these comments now. Thanks for all of your feedback, it was extremely helpful.
For the comments not made inline, here are the things I've done:

@edds:
- rearranged the Sass selectors into the style guide recommended ordering.
- Added a screenshot from the components guide

## On keyboard users vs screen readers
I'm going to throw together an example for Léonie and ask her for feedback on accessibility. Specifically, is it helpful to be able to hide some options from the list for screen reader users and is there a way to allow keyboard users to do this without confusing screen reader users.

I'll write up Léonie's response in a blog post, if it's interesting enough and doesn't turn out to be hugely obvious.

Thanks all!

app/assets/javascripts/govuk-component/option-select.js
```diff
@@ -0,0 +1,136 @@
...
+
+    if (anyOptions && clearingLinkHidden) {
+      this.$clearingLink.removeClass('js-hidden');
+    } else if (!anyOptions && !clearingLinkHidden) {
+      this.$clearingLink.addClass('js-hidden');
+    }
+  };
+
+  OptionSelect.prototype.attachOpenCloseToggle = function attachOpenCloseToggle(){
+    this.$optionSelect.find('.js-container-head').append('<div class="js-toggle"></div>');
```
edds: Seeing as this element is only for displaying the state of the collapse do you think this element might be better named as something like `toggle-indicator`? Being called just `toggle` I was just trying to work out where the event listener to go with it was.

app/assets/javascripts/govuk-component/option-select.js
```diff
@@ -0,0 +1,136 @@
...
+
+  OptionSelect.prototype.attachClearingLink = function attachClearingLink(){
+    this.$optionSelect.find('.js-container-head').append('<a class="js-clear-selected">clear</a>');
+    return this.$optionSelect.find('.js-clear-selected');
+  };
+
+  OptionSelect.prototype.resetOptions = function resetOptions(){
+    this.$options.prop({
+      indeterminate: false,
+      "checked": false
```
edds: Can you remove the quotes from `checked` so it matches `indeterminate`.

alicebartlett: Yes that's a good idea.

app/assets/javascripts/govuk-component/option-select.js
alicebartlett: Yes good point, this is clearer. Thanks Edds.

app/views/govuk_component/option_select.raw.html.erb
```diff
@@ -0,0 +1,31 @@
+<div class="govuk-option-select" tabindex="0">
+  <div class="container-head js-container-head">
+    <div class='option-select-label'><%= title %></div>
+  </div>
+  <div class='options-container'>
+    <div class='js-auto-height-inner'>
+      <% options.each do | option |%>
+
+      <label for='<%= option['id'] %>'>
```
edds: I am slightly concerned about the potential for an attack on this `label` and the `input` below. In theory if an application allowed the options to be user generated I think user could enter a string like `' onclick="nasty"'` which wouldn't get encoded as part of the standard rails string escaping. Using something like `content_tag(:label, for: option["id"])` would add the extra escaping to protect the page.

I could well be wrong here though and am happy to be shown that the standard escaping works.

bradleywright: These are safe - unless a string is explicitly marked as `html_safe` it'll get escaped automatically.

edds: Didn't realise that is auto escaped quote marks. That is cool then.

edds:
Good work on this @alicebartlett. This is really great! Thanks for following up with all the feedback.

dsingleton:
:+1: this is a fantastic addition to components.

tombye:
This is great work and a brilliant resource. Thanks for taking the time to investigate my reckons with @LjWatson, not to mention all the extra work writing that up.

app/views/govuk_component/option_select.raw.html.erb
bradleywright: See https://github.com/rails/rails/blob/3-2-stable/activesupport/lib/active_support/core_ext/string/output_safety.rb#L6

PR #531, opened 2015-01-15
Remove old unused Sass

edds:
I have searched across all our projects and couldn't find anything
using any of this Sass so deleting it.


app/assets/stylesheets/helpers/_core.scss
```diff
@@ -1006,65 +977,3 @@ article {
...
-.linear-promo {
-  position: static;
-  padding-right: 0;
-}
-
-/* extra images for variant testing on driving done pages */
-
-.organ-donor-photo,
-.organ-donor-logo{
-  background-image: image-url("campaigns/organ-donation-photo.jpg");
```
dsingleton: Is it worth deleting the assets too?

dsingleton:
LGTM, if we can safely remote the assets referenced we should do that too.

edds:
@dsingleton thank you, completely overlooked that in my deleting spree. Have also removed the referenced assets as as far as I can tell they are no longer called by anything either.

PR #533, opened 2015-01-22
Use one base template for all HTTP response pages

edds:
Use one base error partial for all response pages rather than two. This
means that we can iterate the look and feel of the one page rather than
having to copy paste code between the two.


app/views/root/_error_page.html.erb
```diff
@@ -0,0 +1,28 @@
+<%
+title = "Sorry, we are experiencing technical difficulties (#{status_code} error)" unless local_assigns.include?(:title)
+heading = "Sorry, we are experiencing technical difficulties" unless local_assigns.include?(:heading)
+intro = "Please try again in a few moments." unless local_assigns.include?(:intro)
```
bradleywright: Is it worth moving these conditionals to the place they're rendered at, for clarity? Not a blocking change.

edds: Good idea. I've moved the heading and intro down to their usages.

bradleywright:
Cool. Is this testable?

bradleywright:
Confirmed these are already tested.

PR #534, opened 2015-01-23
Send HTTP status codes back to Google Analytics.

annapowellsmith:
For 4xx or 5xx responses, send the HTTP status code as a custom variable
to Google Analytics. This will be useful both for the /info pages
application and for search analytics.

Explicitly add status codes as local variable in 4xx responses, since
they weren't recorded before.

### DIFF
--- a/app/views/root/_error_page.html.erb
+++ b/app/views/root/_error_page.html.erb
@@ -26,6 +26,12 @@
       </div>
     </article>
   </div>
+
+<% if status_code %>
+  <script type="text/javascript">
+    window.httpStatusCode = '<%= status_code %>';
+  </script>
+<% end %>
 </main>
 <% end %>
 
--- a/app/views/root/400.html.erb
+++ b/app/views/root/400.html.erb
@@ -1,5 +1,6 @@
 <%= render partial: "error_page", locals: {
     title: "Bad request - 400",
     heading: "Bad request",
-    intro: ''
+    intro: '',
+    status_code: 400
 } %>
--- a/app/views/root/404.html.erb
+++ b/app/views/root/404.html.erb
@@ -3,5 +3,6 @@
     heading: "Page not found",
     intro: '<p>If you entered a web address please check it was correct.</p>
 
-            <p>You can also <a href="/search">search GOV.UK</a> or <a href="/">browse from the homepage</a> to find the information you need.</p>'
+            <p>You can also <a href="/search">search GOV.UK</a> or <a href="/">browse from the homepage</a> to find the information you need.</p>',
+    status_code: 404
 } %>
--- a/app/views/root/406.html.erb
+++ b/app/views/root/406.html.erb
@@ -1,5 +1,6 @@
 <%= render partial: "error_page", locals: {
     title: "Page not found - 406",
     heading: "The page you are looking for can't be found",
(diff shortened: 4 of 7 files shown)


rboulton:
This looks great - should we set the status code for 418 and all the 5xx pages too, for consistency?

annapowellsmith:
Thanks for spotting 418! I've added it. It looks like all the 5xx ones already pass status codes.

zilnhoj:
just a quick q - have you spoken to Ash Chohan to get which CV to use?

On 23 January 2015 at 16:45, Anna Powell-Smith <notifications@github.com>
wrote:

> Thanks for spotting 418! I've added it. It looks like all the 5xx ones
> already pass status codes.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/alphagov/static/pull/534#issuecomment-71223093>.
>



-- 
*John Byrne*

*Performance analyst *Government Digital Service
Cabinet Office
*T:* 07810 814 713
*E:*john.byrne@digital.cabinet-office.gov.uk
6th Floor • Aviation House • 125 - 133 Kingsway • London • WC2 6NH

annapowellsmith:
@zilnhoj yup - he's allocated the CV name and number and confirmed that it's a page-level variable.

zilnhoj:
cool - thanks

On 23 January 2015 at 16:49, Anna Powell-Smith <notifications@github.com>
wrote:

> @zilnhoj <https://github.com/zilnhoj> yup - he's allocated the CV name
> and number and confirmed that it's a page-level variable.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/alphagov/static/pull/534#issuecomment-71223721>.
>



-- 
*John Byrne*

*Performance analyst *Government Digital Service
Cabinet Office
*T:* 07810 814 713
*E:*john.byrne@digital.cabinet-office.gov.uk
6th Floor • Aviation House • 125 - 133 Kingsway • London • WC2 6NH

app/views/root/_error_page.html.erb
```diff
@@ -26,6 +26,12 @@
       </div>
     </article>
   </div>
+
+<% if status_code %>
+  <script type="text/javascript">
+    GOVUK.httpStatusCode = '<%= status_code %>';
```
rboulton: Could this be just "window.httpStatusCode" instead of GOVUK.httpStatusCode?

(I'm only suggesting this because @edds was suggesting using "window" when we spoke offline.  I'm don't have any particular idea why GOVUK or window might be better.)

annapowellsmith: I just used `GOVUK` out of a reflexive dislike of adding anything to `window` - but perhaps `window` is better here if we can't assume that `GOVUK` is available. Will defer to @edds on this.

edds: This code as is will cause `Uncaught ReferenceError: GOVUK is not defined`. If we want to attach the status code onto the `GOVUK` object you will need the line 

```javascript
window.GOVUK = window.GOVUK || {};
```

before this. Alternatively for simplicity I can't see any problem with setting this onto the window object.

annapowellsmith: Thanks @edds, I've amended this to use `window`.

rboulton:
:+1: :heart:

PR #537, opened 2015-01-27
Remove alpha and beta notices

edds:
These used to be used by slimmer to insert a standardised alpha and beta
onto pages. However, that functionality has now all been replaced by the
components. Remove this unused code.

This needs this other pull request to be merged and deployed before this: https://github.com/alphagov/frontend/pull/735


app/controllers/root_controller.rb
```diff
@@ -31,10 +31,7 @@ def govuk_locales
   end
 
   NON_LAYOUT_TEMPLATES = %w(
-    alpha_label
     barclays_epdq
```
jamiecobbett: This might belong in a separate PR, but this also seems to be no longer used either. I think it was in transaction-wrappers, which have moved to [gds-attic](https://github.com/gds-attic/transaction-wrappers/search?utf8=%E2%9C%93&q=epdq)

Search of github.com/alphagov:
https://github.com/search?q=user%3Aalphagov+barclays_epdq&type=Code&utf8=%E2%9C%93

jamiecobbett:
I am not an expert, but aside from a minor comment this looks good to me - I couldn't find any uses of the old style.

dsingleton:
alphagov/frontend#735 got deployed, so i'm happy with this being merged and `barclays_epdq ` being removed in a follow up.

PR #538, opened 2015-01-27
Rename the stylesheets from application to static

edds:
When debugging CSS using browser based debugging tools the styles are
shown listed next to the file name of the CSS file they were found in.
Having the styles which are inherited from static listed as coming from
`application.css` can get confusing as that is the standard name for
CSS files in new Rails projects. This can cause you to look in the
wrong place to find a selector.

This will let us very obviously be able to tell where a CSS rule is
coming from in the same way we can tell `govuk_template` styles because
they are defined in `govuk-template.css`.


dsingleton:
:kiss: :kiss: :kiss:

PR #539, opened 2015-01-29
Pass default value to GA httpStatusCode variable.

annapowellsmith:
We track the HTTP status code using the GA custom variable
httpStatusCode for all 4xx and 5xx responses. However, it's
remarkably difficult to filter for "custom variable not set"
in GA [1], and actually we're usually interested in getting
all successful responses, rather than unsuccessful ones.

So, if this is not a 4xx or 5xx response, pass a default value
of 'unset' to GA, to simplify filtering.

[1] http://stackoverflow.com/questions/7972978/google-analytics-exclude-empty-custom-variable-in-a-custom-report

### DIFF
--- a/app/views/root/_google_analytics.html.erb
+++ b/app/views/root/_google_analytics.html.erb
@@ -14,10 +14,12 @@
   if (window.devicePixelRatio) {
     _gaq.push(['_setCustomVar', 11, 'Pixel Ratio', String(window.devicePixelRatio), 2 ]);
   }
-  // Track status code for failed responses
-  if (window.httpStatusCode) {
-    _gaq.push(['_setCustomVar', 15, 'httpStatusCode', String(window.httpStatusCode), 3 ]);
+  // Track status code
+  if (!window.httpStatusCode) {
+    window.httpStatusCode = '200';
   }
+  _gaq.push(['_setCustomVar', 15, 'httpStatusCode', String(window.httpStatusCode), 3 ]);
+
   // Search result placement tracking, set custom var and destroy the cookie
   if(GOVUK.cookie && GOVUK.cookie('ga_nextpage_params') !== null){
     var customVar = GOVUK.cookie('ga_nextpage_params').split(',');


annapowellsmith:
I don't know enough about static to know whether we can correctly set '200' rather than 'unset' at this point. If we can, then great.

rboulton:
@edds - do you think we can set this to 200?  I suspect it would be correct to do so.

One possible reason that we might not be able to set it to 200 is if we are serving pages with other 2xx status codes sometimes (eg, 201, 202).  If we're doing this, or might be in future, perhaps setting the value to 2xx would be best.

If we're going to set it to a special value, we should consult with Ashraf to make sure we use a value that will be consistent across the other custom variables.

rboulton:
PS: thanks for picking this up, @annapowellsmith - I'd have found myself extremely frustrated when I came to make use of these custom variables in future.  I'm very strongly in favour of setting the variable to *some* value on non-error pages.

app/views/root/_google_analytics.html.erb
```diff
@@ -14,9 +14,11 @@
   if (window.devicePixelRatio) {
     _gaq.push(['_setCustomVar', 11, 'Pixel Ratio', String(window.devicePixelRatio), 2 ]);
   }
-  // Track status code for failed responses
+  // Track status code
   if (window.httpStatusCode) {
     _gaq.push(['_setCustomVar', 15, 'httpStatusCode', String(window.httpStatusCode), 3 ]);
+  } else {
+    _gaq.push(['_setCustomVar', 15, 'httpStatusCode', 'unset', 3 ]);
```
edds: Rather than having making a different call in the conditional it would be safer to being something like:

```javascript
if (!window.httpStatusCode) {
  window.httpStatusCode = 'unset';
}
_gaq.push(['_setCustomVar', 15, 'httpStatusCode', String(window.httpStatusCode), 3 ]);
```

That would reduce the chance of a developer accidentally only updating one of the setCustomVar calls. If for example we wanted to use a different custom variable number.

edds:
I can't think of any situation where the unset value would be 200. I think it is probably safe to make that assumption, as you will be making that assumption in any code which consumes this data anyway. It then moves the assumptions into the data gathering rather than data analysis.

rboulton:
Excellent - this will make data analysis much simpler for this, and avoids having to work out what the correct convention to use for unset custom variables is :)

annapowellsmith:
Thanks, I've pushed @edds's suggested changes and updated the commit message.

rboulton:
Excellent work :+1: :star:

PR #540, opened 2015-01-30
Remove dependency on <article> tags

edds:
Currently to introduce a standardised styling on blocks of content you have to wrap that content in an article tag. This removes the dependancy on an article tag and instead lets you wrap the section in an element with the class `content-block`.

There is also some misc tidying up which was revealed when looking for instances of article tags. Check the commits for more details.


fofr:
I like this a lot :smile:

dsingleton:
:raised_hands: Huge improvement.

tombye:
Brilliant and long overdue :+1:

edds:
This would enabled us to make changes like this: https://github.com/alphagov/frontend/pull/743

PR #542, opened 2015-02-02
Add an ID to the options-container

alicebartlett:
In the future I'll be adding an `aria-controls={options-container-id}` to the `.container-head` of this component. `aria-controls` exposes the relationship between `.container-head` and `.options-container` to accessibility APIs.

For this reason it is important that the `.options-container` actually has an `#id` so I'm adding it here.

Adding a non-optional value like this to a component would normally be a breaking change, however there are no bits of GOV.UK using this component yet so it's fine.

### DIFF
--- a/app/views/govuk_component/docs.yml
+++ b/app/views/govuk_component/docs.yml
@@ -278,6 +278,7 @@
     default:
       key: "market_sector"
       title: "Market sector"
+      options_container_id: "list_of_sectors"
       options:
         - value: "aerospace"
           label: "Aerospace"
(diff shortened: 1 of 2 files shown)


edds:
This looks fine as is. Just wondering why is adding `aria-controls` a future enhancement rather than part of this?

alicebartlett:
@edds Adding in the `aria-controls` is part of a bigger pull request I'm working on that uses js to replace `div.container-head` with a `button` which is better with assistive tech apis. This PR is the outcome of a conversation with Léonie which is documented here: https://github.com/alphagov/static/pull/529#issuecomment-71441290. This work is not big but it's quite hard to fit into my head all at once and so I want to get API changes like this (which would be a breaking change) into master ahead of the grunt work.

edds:
Thank you for explaining. That is cool then :+1:

PR #546, opened 2015-02-18
Change URL in humans.txt to one that doesn't 404

tommyp:
The previous URL was presumably the jobs page. This commit changes it to be the current one. 

I also have a bit of an issue with `developers and content designers` but thought I'd see how controversial that might be before changing it to something that reflects all the disciplines.

### DIFF
--- a/public/humans.txt
+++ b/public/humans.txt
@@ -1,1 +1,1 @@
-GOV.UK is built by a team of developers and content designers at the Government Digital Service in London. If you'd like to join us, see http://digital.cabinetoffice.gov.uk/working-for-gds.
+GOV.UK is built by a team of developers and content designers at the Government Digital Service in London. If you'd like to join us, see https://gds.blog.gov.uk/jobs.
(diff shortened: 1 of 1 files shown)


alicebartlett:
I think it would be pretty sensible to update "developers and content designers" to "people"

rboulton:
:+1: to `people`

edds:
The old URL wasn't the jobs page. It was a page about people who worked here which was quickly going out of date. There is even a link to that page from the jobs page this now points to.

tommyp:
@edds When it gets fixed I'll remove the former commit and change this PR to reflect the `people` change.

bradleywright:
I've fixed the jobs page.

bradleywright:
Also suggest `GOV.UK is built by a team at the Government Digital Service in London.`

tommyp:
Changed the latter commit with @bradleywright's suggestion.

PR #548, opened 2015-02-20
[DISCUSS] Begin migration from classic to universal Google Analytics

fofr:
# For discussion, do not merge

* This PR puts a facade in front of the Google Analytics classic and universal APIs (`GOVUK.analytics`), so that we can track events, pageviews and dimensions identically in both (eg `GOVUK.analytics.trackEvent`, `GOVUK.analytics.trackPageview`, `GOVUK.analytics.setDimension`).
* It manages the loading of analytics libraries and configuration of the tracker profiles from a file, rather than a template, avoiding repetition of this code on every page and allowing it to be minified and cached.
* Slimmer is inserting custom variables into a script tag, while this isn't ideal a choice has been made to leave this as-is for now and tackle separately. To account for Slimmer using only the classic API a shim has been put in place.
* Any use of the analytics API within apps that hasn't been updated to use `GOVUK.analytics` should be unaffected, they just won't be tracking to Universal yet. There are stories to update them.
* Updates to the user satisfaction survey and scroll tracking show how the API can be used.

I'd love to have feedback on all of the above as well as some advice on a couple of the TODOs still left:

* Classic analytics uses `_setAllowLinker`: https://developers.google.com/analytics/devguides/collection/gajs/methods/gaJSApiDomainDirectory#_gat.GA_Tracker_._setAllowLinker – this seems to be used for enabling cross domain tracking but the other parts necessary to make this work aren't configured. It's not clear what Universal should do.
* Classic analytics formerly set the domain name (for cookies) to `.www.gov.uk` or document.domain using `_setDomainName` (see https://github.com/alphagov/static/blob/master/app/views/root/_google_analytics.html.erb#L7-L11). https://developers.google.com/analytics/devguides/collection/gajs/methods/gaJSApiDomainDirectory#_gat.GA_Tracker_._setDomainName – should this behaviour be maintained, should it be migrated to Universal, or should it be removed? Currently it has been removed.
* With analytics now within a file this introduces a new problem: if we bundle it with all our existing javascript and that breaks (eg parse error), this would stop our tracking from working. A way around this would be to include analytics as a separate js file, but that's an extra HTTP hit. I'm not sure which is best.


edds:
This is a good move and I generally approve of the approach :+1: Some general comments though:

As mentioned elsewhere it would be really good to change this to use `GOVUK.analytics` as the base object over `GOVUK.Analytics`. This would fit better within our [naming conventions](https://github.com/alphagov/styleguides/blob/master/js.md#naming-conventions) as you can't create a new instance of `GOVUK.Analytics`. You can still leave the creation of `GOVUK.Analytics = {}` in the `ga-params` script tag till slimmer has been updated to remove its usage of it.

Also rather than using `self = this` would it be possible to use [bind (with the polyfill)](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/javascripts/vendor/polyfills/bind.js) to manage your `this` if `this` would otherwise get changed.

I might leave some other comments inline.

alext:
> Classic analytics formerly set the domain name (for cookies) to .www.gov.uk or document.domain using _setDomainName

I believe this was to force the cookies to be set on www.gov.uk.  Without this, I think they get set on gov.uk, and therefore leak across to other service domains etc.

fofr:
@alext: Setting the `.www.gov.uk` bit has been kept, it's the switch that happens if the domain happens to not be `www.gov.uk` (where it sets the cookie domain as document.domain) that I'm confused about.

alext:
Ah, that'll be so that it works correctly on preview/staging etc.

fofr:
Spoke to @edds about `Analytics` namespace. Rather than renaming Analytics I've opted to remove the namespace from the new objects. See https://github.com/alphagov/static/commit/b359ad15486fc17a32a2885d3c59a44520dc9521

`GOVUK.Analytics` itself can be removed later when Slimmer has been updated. (It should probably become the constructor for GOVUK.analytics)

tombye:
This is really nice and feels a lot cleaner already. All I have is a suggestion for the future, that once this has seen some use in live environments we add some deprecation warnings to the existing objects to give the switchover a push. (Something like using `console.assert` against the old objects).

fofr:
Speaking with @tombye and @edds, there isn't much between having a separate .js file (to prevent something like a parse error from causing tracking to fail) and bundling analytics into application.js. Considering only the users, one less http hit is preferable.

fofr:
@edds I've updated the tracker to use `bind` rather than `self`, and the code is more readable for it.

It has been agreed with the analytics team that we don't need an equivalent to `setAutoLink` within Universal. This was the last TODO that needed work. I'm closing this discussion and opening a new PR.

PR #549, opened 2015-02-23
Begin migration from classic to universal Google Analytics

fofr:
This PR puts a facade in front of the Google Analytics classic and universal APIs (`GOVUK.analytics`), so that we can track events, pageviews and dimensions identically in both (eg `GOVUK.analytics.trackEvent`, `GOVUK.analytics.trackPageview`, `GOVUK.analytics.setDimension`).

* It manages the loading of analytics libraries and configuration of the tracker profiles from a file, rather than a template, avoiding repetition of this code on every page and allowing it to be minified and cached.
 * These files are bundled with our other JS and continues to be loaded at the foot of the page (a separate test will be conducted to assess the differences between tracking code placement at the top and bottom)
* Slimmer is inserting custom variables into a script tag, while this isn't ideal a choice has been made to leave this as-is for now and tackle separately. To account for Slimmer using only the classic API a shim has been put in place.
* All custom events, custom variables and pageviews within static are now tracked in both universal and classic
* Any use of the analytics API within apps that hasn't been updated to use `GOVUK.analytics` should be unaffected, they just won't be tracking to Universal yet. There are stories to update them.
* The `GOVUK.sendToAnalytics` method has been removed

An analysis of the state of analytics across GOV.UK is available as a Google Doc:
https://docs.google.com/a/digital.cabinet-office.gov.uk/document/d/1Tga2eA0uP2hiVuLQW6nyWOuOLmhdekGE6fs_DPec0BM/edit?usp=sharing


edds:
Is it worth changing the interface of `GOVUK.analytics.trackEvent` so that it takes an object rather than arguments?

So instead of:

```js
GOVUK.analytics.trackEvent('browser-check', 'prompt-shown', '', 1, true);
```

Something like:

```js
GOVUK.analytics.trackEvent('browser-check', {
  action: 'prompt-shown', 
  label 'thing',
  value: 1,
  noop: true
});
```

And then have the label, value and noop to be optional. That way I can work out what the arguments mean without having to look them up.

fofr:
@edds, yes, it is worth it. Updated in https://github.com/alphagov/static/commit/df477da852d5b69645e623c733486ce6166469b1

The first two parameters are required, category and action. Optional arguments are now passed in through an options object:

```js
GOVUK.analytics.trackEvent('category', 'action', {
  label: 'label',
  value: 1,
  nonInteraction: true
});
```

edds:
Thank you for that @fofr. My only other concern is the interface to the Tracking object and the requirement to have both a universal and classic tracking codes in the initialisation. In theory if we start using this code for other projects that could be an issue. Though its probably beyond the scope of this and as all our instantiation is happening in this project I am happy for this to be merged as is.

Unless @dsingleton, @alicebartlett, @bradleywright et al. have any large concerns I am happy for this to get merged so we can start iterating.

app/assets/javascripts/analytics/tracker.js
```diff
@@ -0,0 +1,118 @@
+(function() {
+  "use strict";
+  window.GOVUK = window.GOVUK || {};
+
+  var Tracker = function(universalId, classicId, cookieDomain) {
```
bradleywright: Would it be better to make these passed as an object? That way turning off classic analytics would just require us to drop the argument from the code where it's initialised rather than rewrite all of this.

bradleywright:
I've commented on one thing, which @edds already referenced. Otherwise :+1:

app/assets/javascripts/analytics/tracker.js
jabley: :+1: please

jabley:
We'll need to update spotlight with a similar change. I spoke to @fofr in the frontend meeting today about it.

fofr:
I've updated the Tracker constructor to take an object with named parameters in ff058c2

benilovj:
Great work!

dsingleton:
Very happy with the approach here. Great work @fofr

PR #552, opened 2015-03-02
Don't minify analytics initialisers in template

fofr:
In IE8 and below the minified inline javascript was inadvertently wiping out the GOVUK object, causing JS errors.

We aren’t sure exactly why it was wiping out the object, but by removing `var` and referring to window the error is eliminated.

Fix discovered by @edds after some excellent sleuthing.


fofr:
This adds some context:
http://tobyho.com/2013/03/13/window-prop-vs-global-var/

PR #553, opened 2015-03-03
Add a component for /government navigation

tommyp:
This commit adds a component for the government_navigation which contains only markup. This can be included anywhere on GOV.UK and will be lifted by Slimmer and placed into the header, appropriately styling it.

It can be passed a string assigned to active which will then mark the matching link as active. See the code for which strings are used.

This component is slightly odd as Slimmer will move it from the page the first instance it is included and it doesn't come with it's own CSS so the first example on the component view in the guide contains nothing and the second example contains an unstyled list.

Screenshot:
screen shot 2015-03-03 at 16 11 24

### DIFF
--- a/app/views/govuk_component/docs.yml
+++ b/app/views/govuk_component/docs.yml
@@ -40,6 +40,13 @@
       from: "<a href='/government/organisations/cabinet-office'>Cabinet Office</a>"
       history: "Updated 2 weeks ago"
       part_of: "<a href='/government/topics/environment'>Environment</a>"
+- id: "government_navigation"
+  name: "Government navigation"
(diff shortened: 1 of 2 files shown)


edds:
Ideally we would have a way of defining markup only components so that the component guide correctly handles them. This looks good to me though and should let us achieve what we need.

dsingleton:
This throws an error if `active` isn't defined, which was 500'ing the component-guide page for this component.

This is a defaults pattern we've been following in components, as itavoids using ERB specific behaviour that couldn't be ported to Mustache (or similar).
```
<%=
    active ||= nil
%>
```

Otherwise LGTM. +1 that we the component guide should handle the preview better, but that can come later, + the description explaining that is good.

PR #554, opened 2015-03-04
Add page navigation component

jackscotti:
Create a new navigation component.

We wanted to reuse the code used to display navigational links like Simple Smart Answers do (ie https://www.gov.uk/new-state-pension/eligibility).
All this component will do is display the labels, linked to their relative URLs and accompanied by the relative icons.

The Sass for this has been lifted straight out of static with some minor alterations (ie removed bottom margin).

The links will look like this:
screen shot 2015-03-04 at 14 17 09

If one of the 2 links is not provided, it simply won't appear:
screen shot 2015-03-04 at 14 52 55

If we provide labels, the links will look like this:
screen shot 2015-03-05 at 10 39 00

Labels can be anything:
screen shot 2015-03-05 at 11 38 39

I worked on this with @alicebartlett .


app/views/govuk_component/previous_and_next_navigation.raw.html.erb
```diff
@@ -0,0 +1,18 @@
+<nav class="pagination govuk-previous-and-next-navigation" role="navigation" aria-label="Pagination">
+  <ul class="group">
+    <% if local_assigns.include?(:previous_element) %>
+      <li class="previous-element">
+        <a href="<%= previous_element['href'] %>" rel="previous" title="<%= previous_element['title'] %>">
```
fatbusinessman: What will this do if, say, you pass in the `href` and `label`, but no `title`? I can imagine, for example, a case where the link text is sufficient on its own:

    {"href" => "/kittens/adopting", "label" => "Next: adopting a kitten"}

It looks to me like that would make the link render with a blank `title` attribute: in this case, would a screen reader still read out the link text?

edds:
@dsingleton and I have opinions on the API that has been picked for this component. It would be good to chat about it

rboulton:
Have you considered how this component might be extended (not necessarily in this PR) to include information on "page numbers" - eg for the pagination links on search result pages, which look like:
screen shot 2015-03-04 at 15 55 06

jackscotti:
@edds @rboulton @dsingleton @fatbusinessman you are all right, I am going to re-do this PR. Thank you for the feedback.

Jack

jackscotti:
@fatbusinessman after a chat with @edds , we decided to remove the screenreader label because it would only create noise.

@edds I've also swapped the labels and titles tags. Now titles are titles and labels are labels? :confounded: 
I've also removed some of the unused html tags I've added when copy-pasting the code.

app/assets/stylesheets/govuk-component/_previous-and-next-navigation.scss
```diff
@@ -0,0 +1,95 @@
+.govuk-previous-and-next-navigation {
+  display: block;
+
+  @include media-down(mobile) {
```
jonnywyatt: Should the default styling be for the narrowest viewport, then the media queries work up from there?

app/assets/stylesheets/govuk-component/_previous-and-next-navigation.scss
```diff
@@ -0,0 +1,95 @@
...
+      content: " ";
+      }
+
+    &.previous-element {
+      float: left;
+      text-align: left;
+    }
+
+    &.previous-element a {
+      padding: 0.75em 0 0.75em 3em;
```
jonnywyatt: Should spacing be in pixels? https://github.com/alphagov/styleguides/blob/master/css.md#dimensions

app/views/govuk_component/previous_and_next_navigation.raw.html.erb
```diff
@@ -0,0 +1,20 @@
+<nav class="govuk-previous-and-next-navigation" role="navigation" aria-label="Pagination">
```
jonnywyatt: Could use schema.org metadata? http://schema.org/SiteNavigationElement

jackscotti:
@jonnywyatt this is more or less my first "frontend" PR. Are you available to guide me through your comments? Thanks.

jonnywyatt:
@jackscotti happy to discuss, they're just suggestions as I'm not 100% familiar with gov.uk frontend standards. I was under the impression gov.uk uses mobile first styling which covers the first comment. The one about spacing references the CSS styleguide, pixels are preferred to ems apparently. The third is just a suggestion to add semantic value to the markup. Would be interested to get eg @edds views on these before you do anything.

dsingleton:
@jonnywyatt While those are all good changes, it's more important to get this into a component and then once it's abstracted we can change the internal implementation, thats the beauty of components ;)

I'd totally support those changes happening in follow up PRs. When we're converting existing (probably quite old) code into components theres always going to be a lot small changes like that, and it'd bog down the initial conversion, where the focus should be getting the naming/interface/fixtures right.

I'll try and write up some of this as guidance about writing/converting/reviewing components, if people think that'd be useful?

evilstreak:
> I'll try and write up some of this as guidance about writing/converting/reviewing components, if people think that'd be useful?

@dsingleton Yes please, that would be great.

:+1: Good work @jackscotti

jackscotti:
I've just updated a typo in the docs.yml file and added a fixture that has a label made only of text (screenshot in the body of this PR).

jackscotti:
@dsingleton thank you for looking into that. I don't mind addressing what Jon raised but I'll need a more experienced frontend dev for doing so. Up to you all.

dsingleton:
I'd suggest doing that in a follow up - and i'm happy to help with that - smaller PRs, and iterative improvements is a good default.

There are points in the creation of a new component that we need to get right the first time, so to speak, as they become hard to change later. Once an app starts using a component it's much harder to change it's name and arguments. It can help to think of this as a method definition, you can refactor the implementation inside the method, but if you change the name or arguments you'll need to update the calling code too.

The name i'm happy with :+1:, the arguments mostly look good, but i'd suggest changing `previous_element` and `next_element` to be just be `previous` and `next`. They still make sense, and they're simpler.

`href` is makes sense, but it's kind of specific to the `<a>` tag, `link` would be less specific. I don't think it make any difference for this component, but I expect some future components to also take an `href`/`link` in the argument. We'll want a standardised term, so the arguments are intuitive someone calling the component, so it's worth thinking a little about now.

Otherwise this looks really good and it's a great bit of UI to turn into a component. Don't be disheartened by my quibbling over naming, it's just because it's hard to change later :)

jackscotti:
@dsingleton 
- renamed `href` to `link`. 
- `previous_element` and `next_element` are now named `previous_page` and `next_page`. I could not rename them `previous` and `next` because `next` is a reserved word in Ruby.

dsingleton:
Just had an offline discussion with @jackscotti and @edds about `link` vs `href` where we decided `url` is a better name.

This was based on the naming conventions in content store, where `url` is used pretty exclusively when dealing with links.

jackscotti:
@dsingleton  `href` is now `url`.

dsingleton:
Happy with the name/arguments. Thanks you for all the tweaks!

:+1: nice one.

PR #556, opened 2015-03-06
Provide additional visual cue that option-select is scrollable

quis:
## The problem

In research on the Digital Marketplace we have seen users not realise that the option-select has more options than those visible.

On certain combinations of browsers/operating system/number-of-lines-per-option it isn't clear that the component is scrollable. This is because the `max-height` of the container lines up with the bottom edge of the last visible option.

## The solution

This pull request adds Javascript to adjust the height of the container so that the first line of the last visible option is always cut exactly in half horizontally.

Before | After
--- | ---
before | after

It also works when the options wrap to multiple lines:
multiline

## Feedback

I'd especially like to know:
- If there's a more thorough way to test this, without going as far as re-writing the logic from the component in the tests
- If you can understand what the code is doing from reading it—the logic is somewhat complex so if it could be written in a more explicit way I'd welcome that

### DIFF
--- a/app/assets/javascripts/govuk-component/option-select.js
+++ b/app/assets/javascripts/govuk-component/option-select.js
@@ -102,13 +102,39 @@
   OptionSelect.prototype.setupHeight = function setupHeight(){
     var optionsContainer = this.$optionSelect.find('.options-container');
     var optionList = optionsContainer.children('.js-auto-height-inner');
+    var options = optionList.children('label');
     var initialOptionContainerHeight = optionsContainer.height();
     var height = optionList.height();
+    var lastVisibleOption;
 
     if (height < initialOptionContainerHeight + 50) {
       // Resize if the list is only slightly bigger than its container
-      optionsContainer.height(optionList.height());
+      optionsContainer.css({
+        'max-height': 'none',
+        'height': optionList.height()
+      });
+      return;
     }
+
+    // Resize to cut last item cleanly in half
+    lastVisibleOption = options.filter(function(index, option) {
+      var distanceFromTopOfContainer = $(option).offset().top - optionList.offset().top;
+      var containerHeight = optionsContainer.height();
+      return distanceFromTopOfContainer < containerHeight
+    }).last();
+
+    var middleOfLastVisibleOption = (
+      lastVisibleOption.position().top +
+      parseInt(lastVisibleOption.css('border-top-width'), 10) +
+      parseInt(lastVisibleOption.css('padding-top'), 10) +
+      (parseInt(lastVisibleOption.css('line-height'), 10) / 2)
+    );
+
+    optionsContainer.css({
+      'max-height': 'none', // Have to clear the max-height set by the CSS
+      'height': middleOfLastVisibleOption
+    });
+
   };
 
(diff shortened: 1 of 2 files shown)


edds:
I haven't looked at the implementation but this is a totally neat change. Working it out so you can chop of half the word is cool!

dsingleton:
This is a wicked tweak. A subtle but huge improvement :+1:

My first thought was could it be done with pure CSS/SASS? It's quite a lot of JS, so if it could be done with clever CSS instead that'd be preferable. I think it'd be doable (though not sure about the multi-line case) but i've not looked in too much detail - If it's only possible in JS, then i'm happy with that.

It'd also be worth including a multi-line label example in one of the existing fixtures. It's a great edge case to document, exactly what we have the fixtures for.

quis:
@dsingleton Good point about the fixtures, changed in f8d3f68. Will have a think about a pure CSS implementation, although my first instinct is that it's not possible…

quis:
OK, I've had a think. These are my thoughts:

1. If you could guarantee that every option would only span one line then you could calculate a suitable max-height for the container in SASS and not use any JS
2. It might still be worth doing _2._ for non-Javascript users in addition to calculating the height for users with Javascript, because _most_ of the time there won't be any options that span multiple lines, ie it would still be an improvement

alicebartlett:
I've left two inline comments which summarise a conversation we @quis  and I had in person.

Overall I'm really happy with this pull request, I can see the effort you've put into writing it in a clear and understandable way, and the change you're proposing is important/useful and backed up by findings from user research.

Great work :+1:

quis:
@alicebartlett:
- 6028d21, cef4e2b and e3c544c refer to your comment on nested functions
- I've rebased to remove the debugging code which I added in one commit and removed in another (which seems to have made your comments disappear :crying_cat_face:)

app/assets/javascripts/govuk-component/option-select.js
```diff
@@ -99,16 +102,50 @@
...
+
+    this.setContainerHeight(
+      lastVisibleLabel.position().top +
+      parseInt(lastVisibleLabel.css('border-top-width'), 10) +
+      parseInt(lastVisibleLabel.css('padding-top'), 10) +
+      (parseInt(
+        "normal" == lastVisibleLabel.css('line-height') ?
+          lastVisibleLabel.css('font-size') : lastVisibleLabel.css('line-height'),
+        10
+      ) / 2)
```
fofr: This block is quite hard to interpret. It might be worthwhile splitting out the various parts, logic and parsing into named variables beforehand, so the arithmetic is understandable at a glance.

Something like `this.setContainerHeight(position + topBorder + topPadding + halfOfLabelHeight)`

fofr:
I've reviewed your changes and added a minor comment. I've tested this locally and it works wonderfully, great work.

app/assets/javascripts/govuk-component/option-select.js
quis: @fofr Thanks for the feedback. Addressed this specific comment in 003426e.

dsingleton:
:+1: thank you for persisting with all the tweaks/changes, this is looking really good :100:

PR #557, opened 2015-03-12
Use frontend_toolkit’s analytics

fofr:
The analytics API from https://github.com/alphagov/static/pull/549 was ported across to the govuk_frontend_toolkit (https://github.com/alphagov/govuk_frontend_toolkit/pull/162) so that non rails apps can use the code (specifically, spotlight).

* Update the frontend toolkit gem and pull in the new analytics API from the gem
* Remove the old files and tests
* Create a `StaticTracker` for doing static specific analytics tasks, such as handling slimmer custom variables, reading analytics cookies and setting certain custom dimensions
* Print intent and JavaScript error trackers now need to be started. Start them after registering the first pageview.

Analytics documentation:
https://github.com/alphagov/govuk_frontend_toolkit/blob/master/docs/analytics.md


benilovj:
:+1:

PR #558, opened 2015-03-13
Set analytics custom dimensions from `govuk:` meta tags

fofr:
Related to https://github.com/alphagov/slimmer/pull/119

* Find `govuk:` namespaced meta tags and use the content of these tags to set dimensions
* Be explicit about what's being read and set
* Works with both old and new slimmer versions but can be simplified once apps are all on version 8.0.0 or greater

### DIFF
--- a/app/assets/javascripts/analytics/static-tracker.js
+++ b/app/assets/javascripts/analytics/static-tracker.js
@@ -19,7 +19,7 @@
     setHTTPStatusCodeDimension();
     shimNextPageParams();
     shimClassicAnalyticsQueue(classicQueue);
-    setDimensionsFromMetaTags.apply(this);
+    this.setDimensionsFromMetaTags();
 
     // Track initial pageview
     tracker.trackPageview();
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/analytics/static-tracker.js
```diff
@@ -78,6 +79,26 @@
       tracker.setDimension(customVar[1], customVar[3], customVar[2], customVar[4]);
     }
 
+    function setDimensionsFromMetaTags() {
```
edds: Is there any reason this method can't be added to the prototype? That would save the awkward `apply(this)` line above and would better follow the approach we suggest in the JavaScript styleguide.

fofr: I saw no reason to add `setDimensionsFromMetaTags` to the public api as its only use is during initialisation.

edds: Adding it to the public api would let us add proper unit tests to just this method. Which if it is private we can't do. Also while there is no need for this to be added to the public api there is also no harm in it being public. I would lean on the side of making JavaScript methods public unless there is a good reason not to.

fofr: Changed in 11f0014

PR #560, opened 2015-03-17
Mechanism for making analytics calls on next page

fofr:
* Based on an existing technique that sets a GOVUK cookie but which needed classic GA parameters to work
* Instead use `callOnNextPage` which sets a method to call on the `analytics` object with an array of parameters to pass to it
* Use JSON stringify and parse to save method name and parameters
* Call the requested method before the initial pageview, allowing for custom variables to be set from a prior page (as is the current use case)

Example of usage replacing old approach in frontend:
https://github.com/alphagov/frontend/compare/next-page-analytics

(Once the frontend change is merged and deployed, the remaining shims within static can be deleted)

### DIFF
--- a/app/assets/javascripts/analytics/static-tracker.js
+++ b/app/assets/javascripts/analytics/static-tracker.js
@@ -20,6 +20,7 @@
     shimNextPageParams();
     shimClassicAnalyticsQueue(classicQueue);
     this.setDimensionsFromMetaTags();
+    this.callMethodRequestedByPreviousPage();
 
     // Track initial pageview
     tracker.trackPageview();
@@ -79,6 +80,33 @@
       tracker.setDimension(customVar[1], customVar[3], customVar[2], customVar[4]);
     }
 
+  };
+
+  StaticTracker.prototype.callOnNextPage = function(method, params) {
+    params = params || [];
+
+    if (!$.isArray(params)) {
+      params = [params];
+    }
+
+    if (GOVUK.cookie && typeof this[method] === "function") {
+      params.unshift(method);
+      GOVUK.cookie('analytics_next_page_call', params.join('|'));
+    }
+  };
+
+  StaticTracker.prototype.callMethodRequestedByPreviousPage = function() {
+    if (GOVUK.cookie && GOVUK.cookie('analytics_next_page_call') !== null) {
+      var params = GOVUK.cookie('analytics_next_page_call').split('|'),
+          method = params.shift();
+
+      if (typeof this[method] === "function") {
+        this[method].apply(this, params);
+      }
+
+      // Delete cookie
+      GOVUK.cookie('analytics_next_page_call', null);
+    }
   };
 
   StaticTracker.load = function() {
--- a/app/assets/javascripts/analytics/static-tracker.js
+++ b/app/assets/javascripts/analytics/static-tracker.js
@@ -91,16 +91,20 @@
 
     if (GOVUK.cookie && typeof this[method] === "function") {
       params.unshift(method);
-      GOVUK.cookie('analytics_next_page_call', params.join('|'));
+      GOVUK.cookie('analytics_next_page_call', JSON.stringify(params));
     }
   };
 
   StaticTracker.prototype.callMethodRequestedByPreviousPage = function() {
     if (GOVUK.cookie && GOVUK.cookie('analytics_next_page_call') !== null) {
-      var params = GOVUK.cookie('analytics_next_page_call').split('|'),
-          method = params.shift();
+      var params, method;
 
-      if (typeof this[method] === "function") {
+      try {
+        params = JSON.parse(GOVUK.cookie('analytics_next_page_call'));
+        method = params.shift();
+      } catch(e) {}
+
+      if (method && typeof this[method] === "function") {
         this[method].apply(this, params);
       }
 
--- a/spec/javascripts/analytics/static-tracker-spec.js
+++ b/spec/javascripts/analytics/static-tracker-spec.js
@@ -50,7 +50,7 @@ describe("GOVUK.StaticTracker", function() {
       expect(GOVUK.analyticsPlugins.error).toHaveBeenCalled();
     });
 
-    describe('when there is a cookie with next page parameters set', function() {
+    describe('when there is an old GA cookie with next page parameters set', function() {
       it('sets them as a dimension', function() {
         window.ga.calls.reset();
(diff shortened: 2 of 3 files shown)


app/assets/javascripts/analytics/static-tracker.js
```diff
@@ -81,6 +82,33 @@
...
+
+    if (GOVUK.cookie && typeof this[method] === "function") {
+      params.unshift(method);
+      GOVUK.cookie('analytics_next_page_call', params.join('|'));
+    }
+  };
+
+  StaticTracker.prototype.callMethodRequestedByPreviousPage = function() {
+    if (GOVUK.cookie && GOVUK.cookie('analytics_next_page_call') !== null) {
+      var params = GOVUK.cookie('analytics_next_page_call').split('|'),
```
rboulton: This means that none of the parameters passed to the method can contain a '|' character, which isn't obvious and seems likely to trip us up in future.  Parameters will also be silently converted to strings, which might trip people up (though, this being javascript, often won't be noticed).

Might it be better to limit the split to split only once, and pass the full remaining cookie data to the provided method, so the method can parse it as it wants. Particularly since our only current use just wants a single string parameter.

rboulton:
This seems mostly reasonable, but I'm wondering if there's actually a need (at present, or anticipated) for the flexibility of allowing a method name to be specified in the cookie?  The method needs to be defined on all pages on our site - is there an easy way to ensure this is done right?  In the example of usage on frontend, I can't see where the method is defined (but it couldn't be defined in frontend, anyway).

One problem we've seen with the existing implementation of this is that sometimes the cookie data gets loaded on the wrong page (presumably due to the target page failing to load, etc).  I've wondered if it would be good to include the expected target page url (or a hash of it, for brevity) in the cookie data, and only perform the action on the target page if the url matches, to strip out this confusing data.

Finally, don't forget to update https://www.gov.uk/help/cookies with the new cookie name.

app/assets/javascripts/analytics/static-tracker.js
edds: I don't know how long these strings are likely to be and how much we can put into a cookie but could we use `JSON.stringify` and `JSON.parse` to manage the encoding and decoding of the data?

rboulton: JSON.* sounds like a reasonable approach if we want to pass multiple parameters.

fofr:
@rboulton Thank you for the review, some excellent points. The method provided needs to be a named method on the analytics object itself. In the frontend example it would call `GOVUK.analytics.setSearchPositionDimension`.

Because this is only being used once, and because of the problems with `|` delimiting and string coercion, I'm tempted to make this a specific solution for setting just the search result custom dimension.

> One problem we've seen with the existing implementation of this is that sometimes the cookie data gets loaded on the wrong page.

I did wonder if that was happening. Even with this bad data, is the technique and custom dimension still useful?

> Finally, don't forget to update https://www.gov.uk/help/cookies with the new cookie name.

Ah yes, good point.

rboulton:
Even with the bad data, this dimension is invaluable for analysing search results.  I previously estimated the bad data was occurring at most 5% of the time, probably less, but I can't remember the methodology I used to get that estimate off the top of my head.

Making this be a specific solution for setting a custom dimension would be reasonable, I think.  The current implementation isn't specific to setting any particular custom dimension, and I think it would be useful to preserve the ability to set other custom dimensions - but I don't think we need more flexibility than being able to set a list of custom dimensions to particular values.

rboulton:
Even with the bad data, this dimension is invaluable for analysing search results.  I previously estimated the bad data was occurring at most 5% of the time, probably less, but I can't remember the methodology I used to get that estimate off the top of my head.

Making this be a specific solution for setting a custom dimension would be reasonable, I think.  The current implementation isn't specific to setting any particular custom dimension, and I think it would be useful to preserve the ability to set other custom dimensions - but I don't think we need more flexibility than being able to set a list of custom dimensions to particular values.

rboulton:
Even with the bad data, this dimension is invaluable for analysing search results.  I previously estimated the bad data was occurring at most 5% of the time, probably less, but I can't now remember the methodology I used to get that estimate.

Making this be a specific solution for setting a custom dimension would be reasonable, I think.  The current implementation isn't specific to setting any particular custom dimension, and I think it would be useful to preserve the ability to set other custom dimensions (or even expand that to allow setting multiple such dimensions at a time) - but I don't think we need more flexibility than being able to set a list of custom dimensions to particular values.  I think @edds suggestion of using JSON should make it easy to avoid the problems with delimiters and coercion.

Finally, if you do include a hash of the target URL in the cookie, I'm not sure whether the best thing to do is to drop data where the URL doesn't match the expected value, or whether we should be including that in the custom dimensions so we can do some analysis of how often the URL will mismatch.  There's lots of edge cases to think about involving redirections, etc.  That probably merits a wider discussion than fits easily into a PR.

rboulton:
Even with the bad data, this dimension is invaluable for analysing search results.  I previously estimated the bad data was occurring at most 5% of the time, probably less, but I can't now remember the methodology I used to get that estimate.

Making this be a specific solution for setting a custom dimension would be reasonable, I think.  The current implementation isn't specific to setting any particular custom dimension, and I think it would be useful to preserve the ability to set other custom dimensions (or even expand that to allow setting multiple such dimensions at a time) - but I don't think we need more flexibility than being able to set a list of custom dimensions to particular values.  I think @edds suggestion of using JSON should make it easy to avoid the problems with delimiters and coercion.

Finally, if you do include a hash of the target URL in the cookie, I'm not sure whether the best thing to do is to drop data where the URL doesn't match the expected value, or whether we should be including that in the custom dimensions so we can do some analysis of how often the URL will mismatch.  There's lots of edge cases to think about involving redirections, etc.  That probably merits a wider discussion than fits easily into a PR.

rboulton:
Apologies for the duplicate/partial responses (now deleted) earlier - github went weird on me.

app/assets/javascripts/analytics/static-tracker.js
fofr: Updated to use JSON in cc1292d

fofr:
Is this in a state good to merge? I think the improvements @rboulton mentioned are worth doing but should be done separately to this. cc @wryobservations

edds:
This LGTM.

rboulton:
It does look good, and I'm happy to leave improvements for later.

PR #561, opened 2015-03-18
Add Analytics dimensions for History Mode

dsingleton:
Documents in Whitehall that are political and published
by a previous government will be marked as historical.

This adds two new dimensions that will be used to track
content related to History mode, but they've been kept
a little more general, so they could be applied more
widely in the future.

The new dimensions are:
 - PublishingGovernment, the government a document was
   first published under (which may not be when it was
   first published to GOV.UK)
 - PoliticalStatus, whether a document is historic
   (political and non current), political (political
   and current) or apolitical

https://trello.com/c/fednKEAR/105-add-analytics-for-history-mode

Note: This re-orders the tests slightly, so the argument IDs and dimension IDs are both increasing. This feels a bit brittle, and generates noise in the diff, but is more readable, I think. 

Longer term it would be nicer if the test didn't need to care about the order dimensions are set in, eg a test helper like: 
```
expect(classicDimensionFor(1, 'Section').toEqual(['section', 3]);
expect(universalDimensionFor('dimension1').toEqual('section']);
```
Thoughts?

### DIFF
--- a/spec/javascripts/analytics/static-tracker-spec.js
+++ b/spec/javascripts/analytics/static-tracker-spec.js
@@ -80,6 +80,8 @@ describe("GOVUK.StaticTracker", function() {
           <meta name="govuk:format" content="format">\
           <meta name="govuk:need-ids" content="1,2,3">\
           <meta name="govuk:search-result-count" content="1000">\
+          <meta name="govuk:publishing-government" content="2005-to-2010-labour-government">\
+          <meta name="govuk:political-status" content="historic">\
           <meta name="govuk:analytics:organisations" content="<D10>">\
           <meta name="govuk:analytics:world-locations" content="<W1>">\
         ');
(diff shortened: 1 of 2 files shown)


spec/javascripts/analytics/static-tracker-spec.js
```diff
@@ -80,6 +80,8 @@ describe("GOVUK.StaticTracker", function() {
           <meta name="govuk:format" content="format">\
           <meta name="govuk:need-ids" content="1,2,3">\
           <meta name="govuk:search-result-count" content="1000">\
+          <meta name="govuk:analytics:publishing-government" content="2005-to-2010-labour-government">\
+          <meta name="govuk:analytics:political-status" content="historic">\
```
fofr: These values aren't specific to analytics, they are human readable and could be consumed by other things. They don't need the analytics name space.

dsingleton: Noted and updated.

PR #562, opened 2015-03-23
Change Cymraeg link label

wryobservations:
Better descriptor of the page the link takes users to.


issyl0:
Looks fine to me, but why is this being changed?

wryobservations:
Research shows users think clicking the ‘Cymraeg’ link will take them to a translated version of the page they are on, rather than a specific list of services GOV.UK has translated into Welsh. The phrase ‘Cynnwys Cymraeg’ better describes the nature of the page users are taken to.

PR #563, opened 2015-03-23
Revert "Change Cymraeg link label"

wryobservations:
Reverts alphagov/static#562

Research was actually inconclusive - awaiting new link label.


issyl0:
Reverting so this doesn't get deployed by accident...

PR #567, opened 2015-04-02
Add EU logo so it can be added to a manual

benilovj:
A new Countryside Stewardship manual from Defra has to display the European Agricultural Fund for Rural Development logo. Making the agreed asset available for use in the manual.

https://govuk.zendesk.com/agent/tickets/996447


edds:
This shouldn't really live in this repository. If it is for use in manuals it should be in the manuals repo so that we know where the content associated to it lives. Putting it in static will make it almost impossible to ever remove as there is no place to point to to check to make sure nothing is using it anymore.

PR #568, opened 2015-04-07
Remove analytics slimmer shims

fofr:
Now slimmer uses meta tags to convey page metadata and all apps have been updated to the latest slimmer (and deployed), the inline `<script>` tag and shims for resetting the original `_gaq` queue can
be removed.

This also removes the last uses of the `GOVUK.Analytics` namespace.

cc @benilovj


edds:
:tada: Good work!

PR #577, opened 2015-04-16
On option select, only set height one time

quis:
Browsers maintain the scroll position of an option select even when they are closed. This confuses the code that tries to calculate the height of the container to cut off the last 'visible' item.

For some scroll positions this causes the container to get progressively shorter each time it is opened.

error-option-select

To fix this the option select should only have its height set once, the first time it is opened.


edds:
:+1: I approve of the animated gif.

quis:
@edds Can you remind me what I have to type to run the tests on the command line? It's not in the readme.

edds:
`bundle exec rake` will run the whole test suite.

quis:
image

(In lieu of Jenkins)

PR #578, opened 2015-04-16
Change the text of the "Cymraeg" link in the footer

issyl0:
- Research shows users think clicking the "Cymraeg" link will take
  them to a translated version of the page they are on, rather than a
  specific list of services GOV.UK has translated into Welsh. The
  phrase "Rhestr o Wasanaethau Cymraeg" better describes the nature of
  the page users are taken to.

Before:

screen shot 2015-04-16 at 15 39 34

After:

screen shot 2015-04-16 at 15 38 20

(The link text wraps "Built by the Government Digital Service" onto a new line, but according to the designers on our team, this is fine.)

### DIFF
--- a/app/views/root/_footer_support_links.html.erb
+++ b/app/views/root/_footer_support_links.html.erb
@@ -3,6 +3,6 @@
   <li><a href="/help">Help</a></li>
   <li><a href="/help/cookies">Cookies</a></li>
   <li><a href="/contact">Contact</a></li>
-  <li><a href="/cymraeg">Cymraeg</a></li>
+  <li><a href="/cymraeg">Rhestr o Wasanaethau Cymraeg</a></li>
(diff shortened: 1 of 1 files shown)


edds:
It would be really good to update the other places we use this footer if we updating it. These are the ones I can remember exist:

- [Performance platform footer](https://github.com/alphagov/spotlight/blob/master/app/server/templates/page-components/footer_links.html)
- [Service manual footer](https://github.com/alphagov/government-service-design-manual/blob/master/_search/includes/layouts/_footer_support_links.liquid)
- [Trade tariff](https://github.com/alphagov/trade-tariff-frontend/blob/master/app/views/layouts/_base.html.erb#L27-L37)

issyl0:
Thanks @edds. Will do - that is part of the story anyway. Thanks for pointing me in the right direction of the others.

ajlanghorn:
For those unsure, as I was, "Rhestr o Wasanaethau Cymraeg" translates to "List of Welsh services".

PR #581, opened 2015-04-22
Report-a-problem JS refactoring

benilovj:
This work is done in preparation to implementing an alternative design.

This was worked on by @binaryberry, @fofr  and myself.


app/assets/javascripts/report-a-problem-form.js
```diff
@@ -0,0 +1,83 @@
...
+    this.$form.find('input#url').val(window.location);
+  }
+
+  ReportAProblemForm.prototype.postFormViaAjax = function() {
+    $.ajax({
+      type: "POST",
+      url: "/contact/govuk/problem_reports",
+      dataType: "json",
+      data: this.$form.serialize(),
+      success: $.proxy(this.triggerSuccess, this),
```
edds: Any reason to use `proxy` over `Function.prototype.bind`? It would be good to move everything to `bind` where possible for consistency. In this case the `bind` polyfill is already loaded by the time this code will be run.

edds:
Added a minor comment about the use of `proxy` over `bind`.

I also think the new structure of ReportAProblem is slightly confusing. It looks like a prototypal object but the only method which is exposed on the created object works like a static method rather than an instance method. The two other methods which, which would be valid instance methods,  `showConfirmation` and `showError` are also hidden away in the constructors closure. I think it would be a good idea to go further so the object is more like:

```javascript
function ReportAProblem(){}
ReportAProblem.prototype.addToggleLink = function(){
  ...
  this.$container...
}
ReportAProblem.prototype.showConfirmation = function(){
  ...
  this.$container...
}
ReportAProblem.prototype.showError = function(){
  ...
  this.$container...
}
```

This would then also better match the structure and interface you have used for the new `ReportAProblemForm` object.

benilovj:
@edds I've addressed your comments with c6bb0cf and 47e1398

app/assets/javascripts/report-a-problem.js
```diff
@@ -1,84 +1,48 @@
...
+        <p class="report-a-problem-toggle">\
+          <a href="">Is there anything wrong with this page?</a>\
+        </p>\
+      </div>');
+
+    $container.before($toggle);
+    $toggle.on("click", ".report-a-problem-toggle a", function(evt) {
+      $container.toggle();
+      evt.preventDefault();
+    });
```
edds: You can loose the argument to this method and make it work like an instance method using something like: 

```javascript
ReportAProblem.prototype.addToggleLink = function() {
  var $toggle = $('\
    <div class="report-a-problem-toggle-wrapper js-footer">\
      <p class="report-a-problem-toggle">\
        <a href="">Is there anything wrong with this page?</a>\
      </p>\
    </div>');

  this.$container.before($toggle);
  $toggle.on("click", ".report-a-problem-toggle a", function(evt) {
    this.$container.toggle();
    evt.preventDefault();
  }.bind(this));
}
```

benilovj: fixed in de79558a8365cf5145c2d6b371108d14b8bed8c2

PR #582, opened 2015-04-24
Remove clear link

alicebartlett:
screen shot 2015-04-24 at 14 36 08

screen shot 2015-04-27 at 12 20 28

This PR removed the clear link from the option select. I don't anticipate this being a problem for the users of finder-frontend but I would like to know about how it might affect our users on digital marketplace and search.

This PR is a feature change that makes some more significant accessibility improvements a lot more straightforward to implement but there are other ways to lay that groundwork without removing this link, they just require a bit more design work.

More detail can be found in the commits.

@quis, @tombye, @edds, @rboulton As people who are using the option-select I'm interested in if you think this is a good idea or not.


tombye:
Makes sense to remove a click target inside another, should give some immediate improvements :+1:

rboulton:
This makes sense from a technical perspective, and I'm in favour of simplifying things so that we can improve it (as long as we do then get round to the "improve it" bit).  However, some thoughts:

 - do we have any analytics on how often this link is being used?
  - if not, how easily could we get some?
 - we should gather opinions on this both from product people, and from user-research people who might have some more insight into use of this.

As I remember, we added this link because we were concerned that people wouldn't be easily able to get out of a "filtered" context, particularly for a filter with lots of options.  However, one thing that makes this relatively easy in the main search interface is that (at least after a page refresh) we sort the "selected" filter options to the top of the list, so it's generally easy to see all the links which are selected.

quis:
Our user researcher is away today, and I'd like to check with her before I dispense reckons. They are forthcoming though.

app/assets/javascripts/govuk-component/option-select.js
```diff
@@ -14,19 +14,15 @@
     this.$optionsContainer = this.$optionSelect.find('.options-container');
     this.$optionList = this.$optionsContainer.children('.js-auto-height-inner');
 
-    // Build clearing link
-    this.$clearingLink = this.attachClearingLink();
-    this.updateClearingLink();
-
-    // Attach event listeners for clearing
-    this.$clearingLink.on('click', this.resetOptions.bind(this));
```
edds: `this.resetOptions` is now no longer called so can also be deleted.

alicebartlett: Oh yeah, I kept meaning to remove that and then forgetting! Thanks Edds.

edds:
I have no strong opinions on if this is a good idea or not. I can see both sides of the argument. However, if people aren't using it, we end up with less code to maintain so happy for this change to happen. I am also happy with the code changes :+1:

quis:
Reckons update: Neither I or our user researcher can recall seeing anyone use that link while we've been observing them. We'd have no issue with it being removed.

evilstreak:
@rboulton We don't have any event tracking analytics for finders, so adding event tracking for this would need a chunk of work.

We haven't observed the clear link being used in research sessions for finders. Since the Digital Marketplace team haven't either, and since the main search interface pulls the selected options to the top, I'm happy to merge this to unblock the further accessibility work @alicebartlett mentioned.

rboulton:
Thanks for the detailed consideration; good to see this code being simplified.

PR #583, opened 2015-04-28
AB test “Report a problem” design

fofr:
On specified pages, create a multivariate test showing a new report a problem design.

The design:
* collects quantitative data about the page's usefulness which gets sent to Google Analytics
* analytics also tracks opening and closing of original form on test pages, and successfully submitted feedback for both variants
* hides the yes/no links and asks for more details about problems
* forms are still submitted to the same endpoint, but the "what doing" field is pre-filled and hidden

screen shot 2015-04-28 at 11 18 43

screen shot 2015-04-28 at 11 18 51

cc @binaryberry @benilovj

### DIFF
--- a/app/assets/javascripts/report-a-problem.js
+++ b/app/assets/javascripts/report-a-problem.js
@@ -21,6 +21,8 @@
           variant_1: {variantId: 1, callback: renderVariant}
         }
       });
+
+      $form.on("reportAProblemForm.success", this.trackSuccessfulFeedback.bind(this));
     } else {
       renderOriginal();
     }
@@ -112,6 +114,13 @@
 
     if (ReportAProblem.isBeingTestedOnThisPage() && typeof GOVUK.analytics === "object") {
       GOVUK.analytics.trackEvent('ab-test-report-a-problem', 'was-page-useful', {label: wasUseful.toLowerCase()});
+    }
+  };
+
+  ReportAProblem.prototype.trackSuccessfulFeedback = function(evt) {
+    var variant = this.multivariateTest.getCohort();
+    if (typeof GOVUK.analytics === "object") {
+      GOVUK.analytics.trackEvent('ab-test-report-a-problem', 'feedback-submitted', {label: variant});
     }
   };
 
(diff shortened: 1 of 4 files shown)


binaryberry:
:+1:

benilovj:
When the test is finished, we'd ideally like to be able to revert this change wholesale.

This branch has some good improvements to the report-a-problem form that we'd like to keep... is it possible to spin those out into a separate PR, leaving just the stuff that we can remove later?

app/assets/javascripts/report-a-problem.js
```diff
@@ -5,26 +5,140 @@
...
+          '/life-in-the-uk-test',
+          '/government/publications/application-to-transfer-or-retain-a-vehicle-registration-number',
+          '/government/publications/income-tax-tax-relief-for-expenses-of-employment-p87',
+          '/calculate-state-pension',
+          '/calculate-your-holiday-entitlement',
+          '/am-i-getting-minimum-wage'
+        ],
+        slug = ReportAProblem.getCurrentSlug();
+
+    return slugsBeingTested.indexOf(slug) > -1;
```
benilovj: I don't think that this will work for subpages of guides and smart answers (eg `/check-uk-visa/y/afghanistan`)

fofr: Update to match any URL that begins with the specified slug. See https://github.com/alphagov/static/commit/1d6e520c98479ff939aab7ac26ba6d86e2c87c86

binaryberry:
A couple of styling points:
- In the original version, the toggle font size is 16px, vs. 19px in the variant version, which I don't think was part of the design
- Also in variant version, the form text area seems really small, cf. screenshot

screen shot 2015-04-28 at 15 51 17

fofr:
@binaryberry Are your styles cached?

binaryberry:
Urgh, yes. Hard refresh solved the form size issue, though not the font size difference between original and variant toggles.

fofr:
Altered "Was this useful" text to be a light grey and smaller, to match existing toggle text in 1c83e91

screen shot 2015-04-28 at 16 18 16

binaryberry:
:+1:

PR #588, opened 2015-05-06
Add cross domain tracking

bradleywright:
This adds the cross domain tracking functionality from https://github.com/alphagov/govuk_frontend_toolkit/pull/185 to static, making it globally available.

### DIFF
--- a/app/assets/javascripts/analytics/static-tracker.js
+++ b/app/assets/javascripts/analytics/static-tracker.js
@@ -111,6 +111,10 @@
     this.tracker.trackShare(network);
   };
 
+  StaticTracker.prototype.addLinkedTrackerDomain = function(trackerId, name, domain) {
+    this.tracker.addLinkedTrackerDomain(trackerId, name, domain);
+  };
+
   StaticTracker.prototype.setSectionDimension = function(section) {
     this.setDimension(1, section, 'Section');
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/analytics/static-tracker.js
```diff
@@ -111,6 +111,10 @@
     this.tracker.trackShare(network);
   };
 
+  StaticTracker.prototype.addLinkedDomain = function(trackerId, name, domain) {
```
fofr: Should this be `addLinkedTrackerDomain`? (Though I prefer `addLinkedDomain`)

bradleywright: On the fence about that, I thought I'd make the public method shorter. Will update for parity.

fofr:
One minor comment, otherwise this and https://github.com/alphagov/business-support-finder/pull/87 look good.

app/assets/javascripts/analytics/static-tracker.js
bradleywright: Updated and force pushed.

bradleywright:
Updated this and https://github.com/alphagov/business-support-finder/pull/87 to match the public API.

PR #592, opened 2015-05-14
Accessibility option select

alicebartlett:
Following on from: https://github.com/alphagov/static/pull/529 and https://github.com/alphagov/static/pull/582, this PR makes the accessibility recommended by @LjWatson. These are:

- Make the .js-container-head a button
- Add ARIA-controls connecting the button to the checkbox container
- Add ARIA-expanded to the button with the state of the container determining true|false

To achieve these things, this PR:
- adds JavaScript that converts the .js-container-head to a button in the DOM (as it shouldn't be a button if the JavaScript has not loaded as the button won't work). 
- adds CSS to style the button
- adds tests to make sure aria-expanded is being updated and that the button is being added

Here are some screen grabs of the component in various browsers:
Chrome:
chrome

ios Safari7

ios5-safari7

ie6
ie6

ie7
ie7

ie8
ie8

android4 (Icecream Sandwich)
android4-htcone

android5 (Lolipop)
android-osv5-nexus-6

Sorry it took me so long to get round to adding these improvements. This component is _still_ not used in anger anywhere in our component. They now contain significant improvements over the versions used in Finder Frontend and on search pages so should be added to those repos as soon as possible.


app/assets/stylesheets/govuk-component/_option-select.scss
```diff
@@ -1,6 +1,7 @@
 .govuk-option-select {
   background-color: $grey-3;
-  padding: 0 5px 5px 5px;
+  padding-bottom:1px;
+  padding: 5px;
```
edds: This padding line makes the one above redundant.

app/assets/stylesheets/govuk-component/_option-select.scss
```diff
@@ -64,58 +64,56 @@
...
     /* styles for collapsibleness. .js-collapsible is added by the javascript if the browser is not ie6/7 in which case these don't collapse */
     &.js-collapsible {
-      &:hover {
-        background-color: $grey-2;
-        cursor:pointer;
-      }
+      button {
+        z-index: 100;
+        position: relative  ;
+        background-color: $grey-3;
```
edds: This background is redefined below, also you have some extra space before your semi-colon in your position on the line above;

app/assets/javascripts/govuk-component/option-select.js
```diff
@@ -62,20 +81,19 @@
...
       this.open();
     } else {
       this.close();
     }
+    e.preventDefault();
   };
 
   OptionSelect.prototype.open = function open(){
     if (this.isClosed()) {
+      this.$optionSelect.find('button').attr('aria-expanded', true);
```
edds: I would be tempted to call `find('.js-container-head')` rather than just `button` so it is slightly clearer as to what element you are looking for. But the way you have done it also works and is fine.

edds:
Added a couple of super minor comments. But this looks really good, well done on pushing this further, just got to get it implemented somewhere now.

tombye:
Some really thorough testing and description. These changes will make it really solid for users.

I'd suggest updating the JS comments to reference the "checked count" rather than the now-defunct "clearing link":

https://github.com/alphagov/static/blob/accessibility-option-select/app/assets/javascripts/govuk-component/option-select.js#L7

alicebartlett:
@edds @tombye thanks both! I've made the changes you suggested.

PR #596, opened 2015-05-26
Remove the multivariate test of the alternate "report-a-problem" design

benilovj:
The test has now run for several weeks and has generated more than enough data to allow us to judge the goodness of the new design.

This is a direct revert of the two PRs that added the new design.

/cc @fofr


benilovj:
@fofr should some of the improvements that you made to the original RaP in these two PRs be carried across?

fofr:
I don't see any improvements to the original in these PRs other than some whitespace in the stylesheets.

PR #597, opened 2015-05-26
Remove Google Analytics classic

fofr:
__To be deployed 2 June.__

Following on from https://github.com/alphagov/govuk_frontend_toolkit/pull/194

On 2 June we plan to deprecate Google Analytics classic on GOV.UK, from then on we will only use Universal analytics. Google Analytics classic data collection techniques are being deprecated by Google. We should no longer support classic and new tracking should use only Universal.

* Bump govuk_frontend_toolkit to pull in latest analytics
* Remove references to classic tracker
* Remove classic specific arguments (eg custom variable name and scope)
* Update references of `GOVUK.Tracker` to `GOVUK.Analytics`
* Rename `StaticTracker` to `StaticAnalytics` to match upstream refactor

Because the StaticTracker and StaticTrackerSpec files have been renamed, when reviewing I suggest looking at the individual commits.


alicebartlett:
:+1: LGTM

fofr:
I've rebased this against master to pick up @edds’s latest change.

PR #601, opened 2015-06-16
Support symbol and string keys for some components

dsingleton:
This is a transitional change to switch components that expect
string key'd objects to handle symbol key'd objects only.

We should only use one style of keys for objects used in components,
to avoid reduce confusion when using components (I spent a while
trying to work out why the arguments I was passing to prev/next
weren't rendering). Consistency is good.

After discussion with @edds we agreed to use symbols consistently,
as this makes more sense in a ruby/rails environment.

The first step is to support both, as clients of those components
will still be sending string key;d objects and can't be updated
at the time as this app, so we have to;

1. Support both kinds of keys on components that expect strings
2. Update [clients sending string keys](https://github.com/search?utf8=%E2%9C%93&q=govuk_component%2Foption_select+OR+govuk_component%2Fprevious_and_next_navigation+repo%3Aalphagov%2Fcalendars+repo%3Aalphagov%2Fcollections+repo%3Aalphagov%2Fdesign-principles+repo%3Aalphagov%2Ffeedback+repo%3Aalphagov%2Fcourts-frontend+repo%3Aalphagov%2Fcalculators+repo%3Aalphagov%2Flicence-finder+repo%3Aalphagov%2Ffrontend+repo%3Aalphagov%2Fgovernment-frontend+repo%3Aalphagov%2Fsmart-answers+repo%3Aalphagov%2Finfo-frontend+repo%3Aalphagov%2Fcontacts-frontend+repo%3Aalphagov%2Fmanuals-frontend+repo%3Aalphagov%2Fspecialist-frontend+repo%3Aalphagov%2Ffinder-frontend+repo%3Aalphagov%2Fbusiness-support-finder+repo%3Aalphagov%2Fwhitehall+repo%3Aalphagov%2Fstatic&type=Code&ref=searchresults) to use symbols instead
3. Remove support for string keys from these components

This PR is step 1., and using `with_indifferent_access` is a quick
hack to support both, which we can remove at step 3.

### DIFF
--- a/app/views/govuk_component/option_select.raw.html.erb
+++ b/app/views/govuk_component/option_select.raw.html.erb
@@ -5,24 +5,24 @@
   <div class='options-container' id='<%= options_container_id %>'>
     <div class='js-auto-height-inner'>
       <% options.each do | option |%>
-
-      <label for='<%= option['id'] %>'>
+      <% option = option.with_indifferent_access %>
+      <label for='<%= option[:id] %>'>
         <input
         name="<%= key %>[]"
-        value="<%= option['value']%>"
-        id="<%= option['id']%>"
+        value="<%= option[:value]%>"
+        id="<%= option[:id]%>"
         type="checkbox"
 
         <% if local_assigns.include?(:aria_controls_id) %>
           aria-controls="<%= aria_controls_id %>"
         <% end %>
 
-        <% if option['checked'].present? %>
+        <% if option[:checked].present? %>
           checked="checked"
         <% end %>
 
         >
-        <%= option['label']%>
+        <%= option[:label] %>
         </label>
       <% end %>
     </div>
--- a/app/views/govuk_component/previous_and_next_navigation.raw.html.erb
+++ b/app/views/govuk_component/previous_and_next_navigation.raw.html.erb
@@ -1,18 +1,20 @@
 <nav class="govuk-previous-and-next-navigation" role="navigation" aria-label="Pagination">
   <ul class="group">
     <% if local_assigns.include?(:previous_page) %>
+      <% previous_page = previous_page.with_indifferent_access %>
       <li class="previous-page">
-        <a href="<%= previous_page['url'] %>" rel="previous" >
-          <span class="pagination-part-title"><%= previous_page['title'] %></span>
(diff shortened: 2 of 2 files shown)


boffbowsh:
I think it'd make more sense to use `symbolize_keys!` rather than `with_indifferent_access`. Under the hood, a `HashWithIndifferentAccess` transparently converts all keys to strings, which is the opposite of what you want. `HashWithIndifferentAccess` would make sense for sending from clients if you didn't control the components, but here you do.

dsingleton:
> I think it'd make more sense to use symbolize_keys! rather than with_indifferent_access. Under the hood, a HashWithIndifferentAccess transparently converts all keys to strings, which is the opposite of what you want. HashWithIndifferentAccess would make sense for sending from clients if you didn't control the components, but here you do.

`symbolize_keys!` does make a bit more sense, thanks. `HashWithIndifferentAccess` still works in this case through right? I'm tempted not to change it as this is a very temporary shim while I get the callers updated, then it can be removed entirely (hopefully this shouldn't take more than a day)

boffbowsh:
Converting them once is a shim that can be left in there forever if needed. `HashWithIndifferentAccess` not so much.

dsingleton:
> Converting them once is a shim that can be left in there forever if needed. HashWithIndifferentAccess not so much.

If I were leaving it in then `symbolize_keys!` is perfect, but I actively _don't_ want to leave this in forever. 

That would introduce a new inconsistency that some components would respond to string keys in some situations and not others. This PR is only dealing with components that are being called with string keys in the wild, `metadata` also takes objects and already assumes symbol keys.

To provide that consistency each component would have to convert every `Hash` argument (however nested, as [deep_symbolize_keys!](http://apidock.com/rails/v4.2.1/Hash/deep_symbolize_keys%21) requires Rails 4+) is noisy and unintuitive to a lot of people writing components, but more importantly I don't think we need to do it. 

The confusion around types of keys was caused by the [component guide trying to support both](https://github.com/alphagov/govuk-component-guide/pull/20) and the copy/pastable fixtures [using string keys in the past](https://github.com/alphagov/govuk-component-guide/pull/15).

Hopefully that explains the reason for not wanting to make this a permanent change?

boffbowsh:
Fair enough :+1:

PR #603, opened 2015-06-16
Add doc note on keys vs string for objects

dsingleton:
Some components templates in static were using symbol keys and 
others strings, which was inconsistent and confusing, by only using
symbols we avoid confusion.

Tools like the component guide have been updated so they will
obviously error when strings are used - previously it silently
corrected any string/key mismatch.

More here: https://github.com/alphagov/govuk-component-guide/pull/20

I just took some time cleaning things up to only use symbols, so it
seemed worth doc'ing somewhere. That said, i'm not sure this is a 
well phrased bit of documentation, would appreciate thoughts, 
especially from those who've made component changes, like 
@jackscotti or @alicebartlett.

### DIFF
--- a/doc/govuk-components.md
+++ b/doc/govuk-components.md
@@ -35,6 +35,8 @@ To match rails view convention the partial itself should use an underscore, rath
 </div>
 ```
 
+_Note_: For consistency with other components, and Rails convention, you should only use `:symbols`, rather than `"strings"`, for object keys.
+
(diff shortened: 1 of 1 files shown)


doc/govuk-components.md
```diff
@@ -35,6 +35,8 @@ To match rails view convention the partial itself should use an underscore, rath
 </div>
 ```
 
+_Note_: For consistency with other components, and Rails convention, you should only use symbols, rather than strings, for object keys.
```
tommyp: Might it be useful to write this as `:symbols` and `"strings"`?

dsingleton: Yep, that's much better. Updated!

PR #605, opened 2015-06-17
Patch Rails security issue not fixed in a 3.2.x

edds:


alext:
Can we wrap this in a conditional that will raise an error if the Rails version is different than expected so that we know when to review/remove the patch.

jamiecobbett:
@alext if we do, @boffbowsh should update the gist: https://gist.githubusercontent.com/boffbowsh/073d4f36f8677304fa9f/raw/c5eebb5b80c5f261917aab94555805ef34db9cc3/cve_2015_3226_fix.rb

edds:
Hows that?

alext:
Looks good.  Possibly worth including `__FILE__` in that error string so people know where to look.

alext:
but that'll be in the backtrace, so ignore me...

boffbowsh:
I'll update the gist, good idea @alext

PR #608, opened 2015-06-23
Fix retina bug for option-select arrows

alicebartlett:
Although the retina file is included in the project already, for some
reason I never actually added it. So, this commit uses the file.

Pivotal: https://www.pivotaltracker.com/story/show/97293160

Before:
screen shot 2015-06-23 at 15 40 09

After:
screen shot 2015-06-23 at 15 40 20

### DIFF
--- a/app/assets/stylesheets/govuk-component/_option-select.scss
+++ b/app/assets/stylesheets/govuk-component/_option-select.scss
@@ -89,6 +89,7 @@
 
         @include device-pixel-ratio() {
           background-size: 32px 50px;
+          background: image-url('govuk-component/option-select-toggle-sprite.png') no-repeat right 5px;
         }
 
         &:hover {


app/assets/stylesheets/govuk-component/_option-select.scss
```diff
@@ -89,6 +89,7 @@
 
         @include device-pixel-ratio() {
           background-size: 32px 50px;
+          background: image-url('govuk-component/option-select-toggle-sprite.png') no-repeat right 5px;
```
edds: As far as I am aware using the `background` shorthand sets the `background-size` to the default `auto` ([from the spec](http://www.w3.org/TR/css3-background/#the-background)). So this new line will cancel out the `background-size: 32px 50px;` and actually makes the whole `device-pixel-ratio` block redundant. Could we instead just remove the block?

PR #612, opened 2015-07-03
Wrap the related sidebar in an aside

edds:
The aside element lets you mark content on the page which is
complementary to the page. This lets screen readers and other assistive
technology announce to the user that the content isn't part of the main
body of the page.

In user research we have seen users accidentally enter the related
sidebar without realising and assuming that they are still in the main
body of the page. This should help those users.

### DIFF
--- a/app/views/root/related.raw.html.erb
+++ b/app/views/root/related.raw.html.erb
@@ -3,7 +3,7 @@
   <!-- related -->
   <div class="related-container">
 
-    <div class="related" id="related">
+    <aside class="related" id="related">
       <% artefact.related_artefacts.group_by(&:group).each do |group, related_artefacts| %>
         <%
           section = case group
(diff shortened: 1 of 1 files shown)


fofr:
Looks good to me :+1:

Is it worth include a "Related" heading and/or prefixing links with "Related: "? Ed H mentioned that users were navigating by links and headings rather than landmark.

edds:
I think that would be to noisy. If you are browsing around the page it will announce that you have left the main region and entered the complimentary region anyway. Also from watching the user research I think the bigger issue on the page they were testing was that there was no good text on the button they were supposed to be using.

edwardhorsford:
I'm not convinced by this. I feel that some addtional text would be helpful.

Also - the button text 'start now' is standard across all start pages. Perhaps that needs fixing, but it's not isolated to the page we were testing.

PR #620, opened 2015-07-10
Trigger draft watermark differently.

elliotcm:
This relies only on the DRAFT_ENVIRONMENT envvar
being set rather than being set to 'true'.

### DIFF
--- a/app/views/root/_base.html.erb
+++ b/app/views/root/_base.html.erb
@@ -1,7 +1,7 @@
 <% content_for :homepage_url, Plek.new.website_root %>
 <% content_for :page_title, content_for?(:title) ? yield(:title) : "GOV.UK - The best place to find government services and information" %>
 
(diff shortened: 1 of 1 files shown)


app/views/root/_base.html.erb
```diff
@@ -1,7 +1,7 @@
 <% content_for :homepage_url, Plek.new.website_root %>
 <% content_for :page_title, content_for?(:title) ? yield(:title) : "GOV.UK - The best place to find government services and information" %>
 
-<% if ENV["DRAFT_ENVIRONMENT"] == 'true' %>
+<% if ENV["DRAFT_ENVIRONMENT"] %>
```
alext: would be better to use `.present?` to cover the case where it's set, but blank.

elliotcm:
@alext Sure, done.

PR #621, opened 2015-07-13
Remove link titles

edds:
Having link text the same as the title makes reading the link out really
repetitive for a screen reader user as it will read the same text twice.

Remove the repeated text as you get all the content you need from the
link text.


fofr:
:+1:

PR #623, opened 2015-07-22
Bump govuk_frontend_toolkit to 4.1.0

tijmenb:
This adds support for setting the page variable in Analytics.

https://github.com/alphagov/govuk_frontend_toolkit/pull/203

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -32,7 +32,8 @@ end
 
 gem 'plek', '1.11.0'
 
-gem 'govuk_frontend_toolkit', '4.0.1'
+gem 'govuk_frontend_toolkit', '~> 4.1.0'
+
 if ENV['GOVUK_TEMPLATE_DEV']
   gem 'govuk_template', :path => "../govuk_template"
 else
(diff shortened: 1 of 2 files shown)


dsingleton:
Is there a particular reason for switching to a the [pessimistic version constraint](http://guides.rubygems.org/patterns/#pessimistic-version-constraint)? I think we've only used exact version in the past.

(i'm not sure it makes a huge amount of difference, and we [recommend](https://github.com/alphagov/styleguides/blob/master/ruby.md#gemfile) pessimism for trustworthy gems)

tijmenb:
Nope, no particular reason to do that now. Suppose I thought it couldn't hurt to make the change now to make future updates easier.

dsingleton:
I don't have a strong opinion, just curious. We should be able to trust ourselves to semver properly here, so :shipit:

alext:
> We should be able to trust ourselves to semver properly here

That's also what the Gem styleguide says to do these days - https://github.com/alphagov/styleguides/blob/master/ruby.md#gemfile

PR #632, opened 2015-08-10
Add rake task for publishing special routes.

elliotcm:
Includes things like /robots.txt

https://trello.com/c/blLdEZN5/292-make-apps-register-special-routes-on-deploy


lib/tasks/router.rake
```diff
@@ -13,33 +13,6 @@ namespace :router do
...
-        raise
-      end
-    end
-
-    @router_api.commit_routes
-  end
-
   desc "Register application and routes with the router"
-  task :register => [ :register_backend, :register_routes ]
+  task :register => :register_backend
```
alext: `register_backend` shouldn't be needed any more now that this isn't registering any routes.  It shouldn't be talking directly to the router at all.

elliotcm: Does the backend get registered each time by the content store?

alext: Yep - https://github.com/alphagov/content-store/blob/master/app/models/registerable_route_set.rb#L76

elliotcm: Cool.  Updated.

app/models/special_route.rb
```diff
@@ -0,0 +1,127 @@
...
+                    such as developers interested in joining GDS.",
+    },
+    {
+      base_path: "/fonts",
+      content_id: "f0115875-227e-4441-a061-3fdc6d79b7e6",
+      title: "Fonts prefix",
+      description: "This location previously held fonts for GOV.UK, but has since been removed.",
+      type: "prefix",
+      rendering_app: "gone",
+    },
```
alext: I'm not sure it's worth including this in here.  It was never actually live on the www.gov.uk domain, and was originally included by mistake (the fonts are all served through the assets domain).

alext:
This will require updates to the hooks in the deploy scripts to stop calling the `router:register` task, and instead to call the `publishing_api:publish_special_routes` task.

elliotcm:
@alext The relevant PR is here, will need modifying now the router task is gone: https://github.gds/gds/alphagov-deployment/pull/1019

app/models/special_route.rb
elliotcm: Removed.

elliotcm:
@alext Updated to use the gem and include an alias avoiding alphagov deployment changes.

test/test_helper.rb
```diff
@@ -6,3 +6,4 @@
 require 'test/unit'
 require 'rails/test_help'
 require 'mocha/setup'
+require 'webmock/minitest'
```
danielroseman: Why is this here? It seems unrelated.

elliotcm: Leftover from the previous implementation.  Removed.

PR #635, opened 2015-08-14
Tidy the print styles

robinwhittleton:
* Add a few more components to hide
* Normalise values to use SASS variables
* Add margins to tables
* Make table right-alignment work in IE7 & 8


russellthorn:
This looks good to me. A nice bit of code tidying :+1:

PR #638, opened 2015-08-19
core-layout printing

robinwhittleton:
This PR sets up a new print stylesheet specific to core-layout (and header-footer-layout) that only contains the necessary styles.


robinwhittleton:
@dsingleton opinion on this?

dsingleton:
This LGTM.

Some of the (older) naming makes this a little harder to follow than it should be (`static-print` pulling in `core-print` is a little confusing. Not caused by this PR, but until we can get rid of the legacy(ish) static layout it's going to be confusing :/

dsingleton:
One quick question, `static.css` [pulls in](https://github.com/alphagov/static/blob/bbbf7d8b34bb6d8b881ecb3224af6ef204562066/app/assets/stylesheets/static.scss#L2) `header-footer-only`, but the print equivalent doesn't, i'm not sure if this will causes problems with real pages, but it is unintuitive.

robinwhittleton:
Hmm, good point. I’m not entirely sure what the relationship here is: was under the impression that header-footer-only was something completely new. @edds ?

edds:
Good point @dsingleton. 

`header-footer-only.css` is supposed to be a subset of `static.css`. It would probably make sense to remove anything from `_core-print.scss` that also exists in `_print-base.css`. Then change `static-print.scss` to first include `header-footer-only-print.scss` then also include what is left in `_core-print.scss`. If that makes sense @robinwhittleton?

robinwhittleton:
@edds: Pushed a commit that sorts that out. Look OK?

PR #640, opened 2015-08-24
Add analytics meta tag component

edds:
Create a component which takes a links hash object from a content-store
object and turns that into a collection of meta-tags which can then be
read by our analytics JavaScript. This information in then sent to
Google as a custom dimension.

The component doesn't actually render anything visual so doesn't have a
corresponding stylesheet.

Would appreciate @dsingleton's thoughts on this.

### DIFF
--- a/app/views/govuk_component/analytics_meta_tags.raw.html.erb
+++ b/app/views/govuk_component/analytics_meta_tags.raw.html.erb
@@ -1,0 +1,17 @@
+<%
+  links_hash = content_item["links"] || {}
+
+  organisations = []
+  organisations += links_hash["organisations"] || []
+  organisations += links_hash["lead_organisations"] || []
+  organisations += links_hash["supporting_organisations"] || []
+  organisations += links_hash["worldwide_organisations"] || []
+
+  world_locations = links_hash["world_locations"] || []
+%>
+<% if organisations.any? %>
+  <meta name="govuk:analytics:organisations" content="&lt;<%= organisations.map { |link| link["analytics_identifier"] }.join('><') %>&gt;">
+<% end %>
+<% if world_locations.any? %>
+  <meta name="govuk:analytics:world-locations" content="&lt;<%= world_locations.map { |link| link["analytics_identifier"] }.join('><') %>&gt;">
+<% end %>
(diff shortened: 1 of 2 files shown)


app/views/govuk_component/docs.yml
```diff
@@ -427,3 +427,26 @@
...
+  name: "Analytics Meta Tags"
+  description: "Meta tags to provide analytics information about the current page"
+  body: |
+    Currently this just takes a content-store links hash like object which it
+    can then turn into the correct analytics identifier metadata tags.
+  fixtures:
+    default: {}
+    with_orgnisations:
+      links_hash:
+        orgnisations:
```
fofr: There's some typos here. `with_orgnisations`, `orgnisations` and `lead_orgnisations`

app/views/govuk_component/docs.yml
```diff
@@ -427,3 +427,26 @@
       next_page:
         url: "next-page"
         title: "Next page"
+- id: "analytics_meta_tags"
+  name: "Analytics Meta Tags"
+  description: "Meta tags to provide analytics information about the current page"
+  body: |
+    Currently this just takes a content-store links hash like object which it
+    can then turn into the correct analytics identifier metadata tags.
```
jamiecobbett: Is it worth providing a pointer to the code which reads the meta tags?

app/views/govuk_component/analytics_meta_tags.raw.html.erb
```diff
@@ -0,0 +1,15 @@
+<%
+  organisations = []
+  organisations += links_hash["organisations"] || []
```
jamiecobbett: Given that we know we want to render meta tags from attributes that aren't in the links hash, should this instead take a content item hash?

edds:
I have updated this in relation to the feedback.

dsingleton:
:+1: to the idea of encapsulating the analytics meta into a component. 

It feels a bit weird to be tying the component so closely to the specific `content_item` representation but presumably can be expanded to include `format`, `rendering_app`etc, without changing the calling code in the application (or slimmer).

So, generally, :+1:

PR #641, opened 2015-08-25
Add static-print.css to precompile list

boffbowsh:
/cc @edds @robinwhittleton

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -47,12 +47,12 @@ module Static
       *.js
       static.css
       static-ie*.css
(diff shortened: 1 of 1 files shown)


edds:
So this actually wants to add `core-layout-print.css`, `header-footer-only-print.css`, `static-print.css` and remove `print.css`. As per the changes in https://github.com/alphagov/static/pull/638

boffbowsh:
Added the removal of print.css. The others are covered by wildcards.

PR #646, opened 2015-09-09
Upgrade ruby to 2.2.3

binaryberry:
Added test-unit and minitest gems as they are no longer part of the ruby standard library in 2.2.x. Upgraded sass, sass-rails and yui-compressor to avoid breakages.

cc @boffbowsh 

[Trello](https://trello.com/c/e4NNxX15/99-upgrade-ruby-and-rails-for-the-static-app-medium)


Gemfile
```diff
@@ -24,6 +24,8 @@ group :test do
   gem 'mocha', '0.13.3', :require => false
   gem 'shoulda', '2.11.3'
   gem 'webmock'
+  gem 'test-unit'
```
jamiecobbett: Do we really need both test-unit and minitest?

binaryberry: We do! once we upgrade rails we could consider removing test-unit though.

Gemfile
```diff
@@ -21,9 +21,12 @@ end
 
 group :test do
   gem 'capybara', '2.1.0'
-  gem 'mocha', '0.13.3', :require => false
+  gem 'mocha', '~> 1.1.0', :require => false
   gem 'shoulda', '2.11.3'
   gem 'webmock'
+  gem 'test-unit'
+  gem 'minitest', '~> 5.8.0'
+  gem 'yui-compressor', '~> 0.12.0'
```
jamiecobbett: Doesn't this belong in the `assets` group?

binaryberry: Good point. Corrected!

PR #647, opened 2015-09-11
Rails 4.0 upgrade

benilovj:
This PR upgrades Rails from 3.2 to 4.0.13. An upgrade to higher versions of Rails will follow shortly.

The most dramatic change is moving the running of the jasmine specs to the CI build (see 7c45659).

/cc @binaryberry @boffbowsh


config/application.rb
```diff
@@ -54,7 +48,7 @@ class Application < Rails::Application
       static-print.css
     }
 
-    config.assets.css_compressor = :yui
+    # config.assets.css_compressor = :yui
```
boffbowsh: Why's this commented out?

config/initializers/filter_parameter_logging.rb
```diff
@@ -0,0 +1,4 @@
+# Be sure to restart your server when you modify this file.
```
boffbowsh: I don't think this file is needed at all, there aren't any passwords used in this repo.

config/application.rb
boffbowsh: Should probably remove this instead of commenting it out

config/initializers/filter_parameter_logging.rb
benilovj: fixed with 5ef6c2d

config/application.rb
benilovj: Removed

I also:

- got rid of the `assets` Gem group
- pinned the asset gems as per the style guide
- removed the (now unused) yui gem

benilovj:
This should be ready for another review.

boffbowsh:
Looks good - would be good to deploy it to preview.

benilovj:
[Pushed to preview](https://deploy.preview.alphagov.co.uk/job/Preview_Deploy/3422/console), seems to have been deployed and running OK.

PR #648, opened 2015-09-11
Rails 4.1 upgrade

benilovj:
This is to keep the app on a supported version of Rails.

The companion deployment [pull request is here](https://github.gds/gds/alphagov-deployment/pull/1051).


boffbowsh:
:+1: for this change, but I'll leave merging this to someone with access to alphagov-deployment

jamiecobbett:
@benilovj merged - the companion PR does need merging, but I'll let you resolve that when you're back.

PR #649, opened 2015-09-15
Take secret key base from environment

benilovj:
This is in accordance with the [12 factor app principles](http://12factor.net/config)
of storing config in the environment.

/cc @alext @jamiecobbett

### DIFF
--- a/config/secrets.yml
+++ b/config/secrets.yml
@@ -2,3 +2,7 @@ development:
   secret_key_base: d9f7b26908faa9dcb8d2c9485deb4c57e78bbeac83b9e5d9811a17573aaa6f68e31022e099bb492dc214604687a03a7a8ff3246e183a4e390de9355a88c61ea9
 test:
   secret_key_base: 736e7edacf2548509cb6433907d98d492fcb308434dc38999e335c8960084dc4d4baab13e9c0e65ed30fb2c2ad0204a691532746bf0a7329d1bc5973af8b5e38
+# Do not keep production secrets in the repository,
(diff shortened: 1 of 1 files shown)


config/secrets.yml
```diff
@@ -2,3 +2,7 @@ development:
   secret_key_base: d9f7b26908faa9dcb8d2c9485deb4c57e78bbeac83b9e5d9811a17573aaa6f68e31022e099bb492dc214604687a03a7a8ff3246e183a4e390de9355a88c61ea9
 test:
   secret_key_base: 736e7edacf2548509cb6433907d98d492fcb308434dc38999e335c8960084dc4d4baab13e9c0e65ed30fb2c2ad0204a691532746bf0a7329d1bc5973af8b5e38
+# Do not keep production secrets in the repository,
+# instead read values from the environment.
+production:
+  secret_key_base: <%= ENV["SECRET_KEY_BASE"] %>
```
jamiecobbett: `ENV.fetch` might be preferable so that an error is raised if it is missing?

alext: As it stands, it matches what `rails new` will generate, which is worth sticking with if possible. I think Rails will already raise an error if the secret key isn't set.

tijmenb: I looked up why rails doesn't do this by default: https://github.com/rails/rails/pull/17175#issuecomment-85163310

TLDR:

> Turns out that you can't use fetch in secrets.yml as it would blow up in development when secrets.yml is loaded.

PR #657, opened 2015-09-24
Rich Govspeak with bold

tommyp:
As we need to have bold on certain pages of GOV.UK (the Highway Code)
this commit updates the Govspeak component to have extra CSS set by a
flag of `rich_govspeak`. This can be set by the rendering App.
Additional CSS could be added to this rich mode if we wish.

This has been discussed with both Product Managers, Content and
Designers and the approach I've followed was agreed upon.

[Ticket](https://trello.com/c/eYyOkNRR/131-highway-code-manual-has-styled-strong-tags).

### DIFF
--- a/app/assets/stylesheets/govuk-component/_govspeak.scss
+++ b/app/assets/stylesheets/govuk-component/_govspeak.scss
@@ -8,6 +8,12 @@
     text-align: start;
   }
 
+  &.rich-govspeak {
+    strong {
+      font-weight: bold;
+    }
+  }
+
   h2:first-child,
   h3:first-child,
   h4:first-child,
--- a/app/views/govuk_component/docs.yml
+++ b/app/views/govuk_component/docs.yml
@@ -201,6 +201,10 @@
           <li>تابعنا باللغة العربية عبر</li>
           <li>تابعنا باللغة العربية عبر تويتر </li>
         </ul>
+    with_rich_govspeak:
+      rich_govspeak: true
+      content: |
+        <p>This content has some <strong>rich govspeak</strong></p>
 - id: "title"
   name: "Page title"
   description: "A page title with optional context label"
--- a/app/views/govuk_component/govspeak.raw.html.erb
+++ b/app/views/govuk_component/govspeak.raw.html.erb
@@ -2,6 +2,10 @@
   direction_class = ""
   direction_class = " direction-#{direction}" if local_assigns.include?(:direction)
 %>
-<div class="govuk-govspeak<%= direction_class %>">
+<%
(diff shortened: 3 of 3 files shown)


dsingleton:
Approach LGTM. `rich_govspeak` is a good descritive name, but a little bit redundant as it's already scoped to the govspeak component, though i'm not sure just `rich` wouldn't be a little confusing. Any ideas for something that would be meaningful, but not need to duplicate `govspeak`?

If we can't come up with anything better then i'd be happy to merge.

(I wouldn't normally be so pedantic about naming, but with components it gets harder to change later, so worth debating now)

jamiecobbett:
What other differences might **rich** govspeak have?

tommyp:
@jamiecobbett I spoke to @markhurrell and he said that we may want to add more things to a `rich` version of Govspeak later, so I didn't want to start chaining classes.

app/assets/stylesheets/govuk-component/_govspeak.scss
```diff
@@ -8,6 +8,12 @@
     text-align: start;
   }
 
+  &.rich-govspeak {
+    strong {
+      font-weight: bold;
```
fofr: Presumably somewhere there's a reset that removes the default `strong` browser rendering?

tommyp: Probably [this line](https://github.com/alphagov/static/blob/master/app/assets/stylesheets/helpers/_reset.scss#L30).

alextea: Shouldn't this be enabled by this commit https://github.com/alphagov/static/commit/2184c82e45c186c71f12b026209eca610f816228?

fofr: Whitehall seems to use a different reset file without that __strong__ rule: https://github.com/alphagov/whitehall/blob/master/app/assets/stylesheets/frontend/reset/_reset.scss#L30

fofr: Seems to be true with Manuals too:
https://github.com/alphagov/manuals-frontend/blob/master/app/assets/stylesheets/_reset.scss

So in those cases these rules would work, but perhaps not everywhere.

tommyp: So what should I do then?

fofr:
Spoken with @tommyp in person and happy to merge this, with the caveat that future apps that enable rich govspeak might need to alter reset styles for strong markdown to display correctly.

PR #659, opened 2015-10-07
Fix emergency publishing styling

benilovj:
The styling for the emergency banners need to be in every layout,
otherwise they will appear unstyled in apps using `header-footer-only`
or `core-layout`.

/cc @dsingleton @boffbowsh


boffbowsh:
LGTM, but I'll let @dsingleton give the final OK

dsingleton:
:+1:

PR #660, opened 2015-10-08
Fix alignment of report-a-problem at certain widths

dsingleton:
At a breakpoint below 1020 the report a problem link, and form, weren't aligned to the grid of the rest of the page, and were indented by an additional 30px on each side.

This was caused by `%site-width-container` being used on the report a problem containers, within `#wrapper`, which already extends `%site-width-container`, causing double margins.

As the report a problem elements already inside a page width container it's same to remove the use of `%site-width-container` for them.

| 🌟  | Before | After |
| --- | --- | --- |
| Closed | report-before | report-after |
| Open | report2-before | report2-after |
| Mobile | report3-before | report3-after  |

### DIFF
--- a/app/assets/stylesheets/helpers/_report-a-problem.scss
+++ b/app/assets/stylesheets/helpers/_report-a-problem.scss
@@ -3,7 +3,6 @@
 }
 
 .report-a-problem-container{
-  @extend %site-width-container;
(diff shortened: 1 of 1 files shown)


fofr:
:+1:

PR #664, opened 2015-10-15
Split Component docs into one file per component

dsingleton:
This is an internal only change, and won't affect the response the `govuk_component_docs` route returns - though it will enforce an alphabetical order, which is a good default anyway

As we added more components, and fixtures, this file was growing pretty big and hard to read/edit.

It's rare for someone to be looking at more than one component at a time, the common case being working on fixtures for a single component, so it makes sense to group the YML that way and keep the files smaller.

If the number of fixtures continue to grow then it might be worth splitting the fixtures out to one per file, grouped in a directory by component.

Also updates the rails generator to use the file-per component structure, which simplifies that code quite a bit too. 👍

**Note**: I've updated the docs in a couple of places, where they'd have been outright wrong, but otherwise tried to leave them along. They need a bit of an overhaul, and doing that as part of this PR would be a rabbit hole.

:eyes: CC @gemmaleigh, as we were discussing this earlier, and @fofr to get another GOV.UK FE


app/views/govuk_component/docs/alpha_label.yml
```diff
@@ -0,0 +1,9 @@
+---
```
tommyp: Discussed with @dsingleton via other channels but summarising here. `---` is being added by the generator and seems to be a YAML thing called a [structure](http://www.yaml.org/spec/1.2/spec.html). 

> YAML uses three dashes (“---”) to separate directives from document content. This also serves to signal the start of a document if no directives are present.

Probably better leaving them in each file if it's output from the generator so people later on won't think the generator is being weird.

fofr:
:heart: :+1:

gemmaleigh:
This makes it so much clearer! Thanks @dsingleton @tommyp

PR #669, opened 2015-10-22
Fix AnalyticsMetadata component in a better way

dsingleton:
Follow up to https://github.com/alphagov/static/pull/668/

With real Content Items, rather than fixtures, the object will be
a `GdsApi::Response` object, which doesn't support `with_indifferent_access`
and would break for an app passing that in.

Instead of trying to make the fixtures/content item both response to string
keys, which is an exception to the component style guide, ensure the
content item keys are all symbols, by casting to a hash.

For fixtures, this will be a no-op, for the API response it will switch to
a symbol-key'd hash, the rest of the template can then deal only with
symbol key'd hashes.

IMO this is a cleaner fix, as it also further removes the main bit of
component logic from the specific structure/object of a content item

### DIFF
--- a/app/views/govuk_component/analytics_meta_tags.raw.html.erb
+++ b/app/views/govuk_component/analytics_meta_tags.raw.html.erb
@@ -1,17 +1,18 @@
 <%
-  links_hash = content_item.with_indifferent_access["links"] || {}
+
+  links_hash = content_item.to_h[:links] || {}
 
   organisations = []
-  organisations += links_hash["organisations"] || []
-  organisations += links_hash["lead_organisations"] || []
-  organisations += links_hash["supporting_organisations"] || []
-  organisations += links_hash["worldwide_organisations"] || []
+  organisations += links_hash[:organisations] || []
+  organisations += links_hash[:lead_organisations] || []
+  organisations += links_hash[:supporting_organisations] || []
+  organisations += links_hash[:worldwide_organisations] || []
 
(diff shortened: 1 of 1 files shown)


app/views/govuk_component/analytics_meta_tags.raw.html.erb
```diff
@@ -1,17 +1,18 @@
...
-  links_hash = content_item.with_indifferent_access["links"] || {}
+
+  links_hash = content_item.to_h[:links] || {}
 
   organisations = []
-  organisations += links_hash["organisations"] || []
-  organisations += links_hash["lead_organisations"] || []
-  organisations += links_hash["supporting_organisations"] || []
-  organisations += links_hash["worldwide_organisations"] || []
+  organisations += links_hash[:organisations] || []
```
benlovell: `links_hash.fetch(:organisations, [])` would be more idiomatic. If you agree, you can apply it to all the `links_hash` accesses here.

dsingleton: It is, but component templates _try_ to avoid using too many ruby/ERB specific features, with the idea that they could be translated into something like mustache and used elsewhere easily - though this template is very much the exception to that rule :(

I think the key-value-or-empty-array behaviour could be better expressed using `slice` anyway, so I'm inclined to leave that, and this, kind of refactoring to another PR, and this one can fix the live issue.

PR #670, opened 2015-10-30
HMRC webchat

robinwhittleton:
Add a webchat shim for HMRC’s eGain implementation. This is currently limited to one page: the HMRC self-assessment contact page.

### DIFF
--- a/app/assets/stylesheets/helpers/_webchat.scss
+++ b/app/assets/stylesheets/helpers/_webchat.scss
@@ -29,3 +29,7 @@
   color: $white;
   @include bold-19;
 }
+
+.webchat-banner .reject {
+  margin-left: 1em;
+}
--- a/app/assets/stylesheets/helpers/_webchat.scss
+++ b/app/assets/stylesheets/helpers/_webchat.scss
@@ -1,30 +1,31 @@
 .webchat-banner {
-  background-color: #005ea5;
-  color: white;
-  font-size: 19px;
-  height: 0;
-  padding: 0;
+  background-color: $govuk-blue;
+  color: $white;
+  @include core-19;
+  max-height: 0;
+  padding: 0 $gutter-half;
   opacity: 0;
   margin-bottom: 0;
-  transition: height .7s ease-out, opacity .7s ease-out, margin-bottom 0.7s ease-out;
+  -webkit-transition: .7s ease-out;
+  -webkit-transition-property: max-height, opacity, margin-bottom, padding-top, padding-bottom;
+  transition: .7s ease-out;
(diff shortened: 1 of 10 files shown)


app/assets/stylesheets/helpers/_webchat.scss
```diff
@@ -0,0 +1,35 @@
...
+  color: $white;
+  @include core-19;
+  max-height: 0;
+  padding: 0 $gutter-half;
+  opacity: 0;
+  margin-bottom: 0;
+  -webkit-transition: .7s ease-out;
+  -webkit-transition-property: max-height, opacity, margin-bottom, padding-top, padding-bottom;
+  transition: .7s ease-out;
+  transition-property: max-height, opacity, margin-bottom, padding-top, padding-bottom;
```
alextea: We have a [transition helper mixin](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/stylesheets/_css3.scss#L57) in the FET

fofr:
Are there some screens of this in action?

fofr:
I'm not seeing any JS specs for this. Should we have some test coverage?

robinwhittleton:
@alextea I hadn’t actually realised that, but either way we don’t have a `transition-property` mixin and I want that as we’ve got so many things we’re transitioning. I guess I could use `all` as a property but that seems a little heavy handed.

robinwhittleton:
@fofr the short answer: yes. The long answer: because the process relies on there being a human at the other end, end-to-end testing is not possible. We definitely should have some unit tests though.

tombooth:
We have added some tests for what we can independent of the 3rd party service. Attached is an image of the banner, as there is currently no one in the queue to answer the web chat I cannot demonstrate what happens when you click accept (that is in HMRC's control anyway).

Picture of the banner displayed to the user

tombooth:
I am going to merge this as it needs to be deployed today. I worked with Robin on this and we have reviewed each-others code. I take full responsibility if this breaks anything today in production.

fofr:
It looks like a hidden iframe is causing problems on all HMRC contact pages.

Try mousing over "deaf" on this page:
https://www.gov.uk/government/organisations/hm-revenue-customs/contact/accessibility-and-specialist-services

Or copying the top email address on this page:
https://www.gov.uk/government/organisations/hm-revenue-customs/contact/air-passenger-duty

bradwright:
I think this is meant to fix this: https://github.com/alphagov/static/pull/671

robinwhittleton:
@fofr @bradwright fix for that in https://github.com/alphagov/static/pull/672

PR #671, opened 2015-11-05
Make the webchat unable to be interacted with, when closed

robinwhittleton:
We’d missed a hidden overflow in the original version, which was leaving the white text unnoticed on top of the main content. The links could then be clicked, which would result in a JS error.


fofr:
Presumably these bad links are also available to assistive devices?

robinwhittleton:
No, we set `aria-hidden:true` when we [remove the class](https://github.com/alphagov/static/blob/master/app/assets/javascripts/webchat.js.erb#L33).

PR #672, opened 2015-11-06
Only apply the web chat banner to pages it’s required on

robinwhittleton:
The banner was being appended before the shouldOpen() check. This moves the insertion point calculation and banner appending to after that.


fofr:
:+1:

PR #673, opened 2015-11-06
Fix analytics component and add tests for some components

jamiecobbett:


dsingleton:
Broadly looks good to me, it's good to get an example of some testing.

There's some stuff we should add in the future, but it's worth mentioning here for anyone interested/

- Making the component generator create a test file - to gently encourage testing
- Backfilling tests for all components - at least one unit test each, so someone making changes can add to it, rather than having to start from scratch (bearing in mind those making changes are less likely to be familiar with ruby testing
- Finding a nicer/terser way to express `assert`s, it's [a bit clunky](https://github.com/alphagov/static/commit/6d2aeef1f56daa0c95c138080559cd35fbecf915#diff-082b11654fe5f535469a714847a40fa5R22) currently, and again, considering who'll be writing tests, the less painful it is the better

PR #674, opened 2015-11-09
Add format and need-ids to analytics component

fofr:
https://trello.com/c/2LJAtceC/156-new-world-analytics-component-medium
Allow format and Need IDs to be tracked to Google Analytics through the existing component.

Matches the format that "old world" frontends are using via Slimmer:
https://github.com/alphagov/slimmer/blob/master/lib/slimmer/processors/metadata_inserter.rb

cc @jamiecobbett @dsingleton

### DIFF
--- a/app/views/govuk_component/analytics_meta_tags.raw.html.erb
+++ b/app/views/govuk_component/analytics_meta_tags.raw.html.erb
@@ -2,8 +2,14 @@
   # We have to call deep_symbolize_keys because we're often dealing with a
   # parsed JSON document which will have string keys by default, but our
   # components use symbol keys and we want consistency.
-  links_hash = content_item.to_h.deep_symbolize_keys[:links] || {}
+  content_item_hash = content_item.to_h.deep_symbolize_keys
+  links_hash = content_item_hash[:links] || {}
   meta_tags = {}
+
+  meta_tags["govuk:format"] = content_item_hash[:format] if content_item_hash[:format]
+
+  need_ids = content_item_hash[:need_ids] || []
+  meta_tags["govuk:need-ids"] = need_ids.join(',') if need_ids.any?
 
   organisations = []
   organisations += links_hash[:organisations] || []
--- a/app/views/govuk_component/analytics_meta_tags.raw.html.erb
+++ b/app/views/govuk_component/analytics_meta_tags.raw.html.erb
@@ -16,12 +16,12 @@
   organisations += links_hash[:lead_organisations] || []
   organisations += links_hash[:supporting_organisations] || []
   organisations += links_hash[:worldwide_organisations] || []
-  organisations_content = organisations.map{ |link| "<#{link[:analytics_identifier]}>" }.join
-  meta_tags["govuk:analytics:organisations"] = organisations_content if organisations_content
+  organisations_content = organisations.map { |link| "<#{link[:analytics_identifier]}>" }.join
+  meta_tags["govuk:analytics:organisations"] = organisations_content if organisations.any?
 
   world_locations = links_hash[:world_locations] || []
-  world_locations_content = world_locations.map{ |link| "<#{link[:analytics_identifier]}>" }.join
-  meta_tags["govuk:analytics:world-locations"] = world_locations_content if world_locations_content
+  world_locations_content = world_locations.map { |link| "<#{link[:analytics_identifier]}>" }.join
+  meta_tags["govuk:analytics:world-locations"] = world_locations_content if world_locations.any?
 %>
 <% meta_tags.each do |name, content| %>
   <meta name="<%= name %>" content="<%= content %>">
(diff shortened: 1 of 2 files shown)


app/views/govuk_component/analytics_meta_tags.raw.html.erb
```diff
@@ -2,19 +2,27 @@
   # We have to call deep_symbolize_keys because we're often dealing with a
   # parsed JSON document which will have string keys by default, but our
   # components use symbol keys and we want consistency.
-  links_hash = content_item.to_h.deep_symbolize_keys[:links] || {}
+  content_item_hash = content_item.to_h.deep_symbolize_keys
+  links_hash = content_item_hash[:links] || {}
```
benlovell: `Hash#fetch(:key, {})` may be preferable.

app/views/govuk_component/analytics_meta_tags.raw.html.erb
```diff
@@ -2,19 +2,27 @@
...
+  meta_tags = {}
+
+  meta_tags["govuk:format"] = content_item_hash[:format] if content_item_hash[:format]
+
+  need_ids = content_item_hash[:need_ids] || []
```
benlovell: Same here.

app/views/govuk_component/analytics_meta_tags.raw.html.erb
jamiecobbett: We're keeping the language constructs used in these templates as generic as possible so that they could be ported to another templating language - @dsingleton can explain the rationale for that.

jamiecobbett: Having said that, we're using `map` and `join`...

app/views/govuk_component/analytics_meta_tags.raw.html.erb
```diff
@@ -2,19 +2,27 @@
...
+  meta_tags["govuk:need-ids"] = need_ids.join(',') if need_ids.any?
 
   organisations = []
   organisations += links_hash[:organisations] || []
   organisations += links_hash[:lead_organisations] || []
   organisations += links_hash[:supporting_organisations] || []
   organisations += links_hash[:worldwide_organisations] || []
+  organisations_content = organisations.map{ |link| "<#{link[:analytics_identifier]}>" }.join
```
jamiecobbett: We should have a space before the `{` as per the [styleguide](https://github.com/alphagov/styleguides/blob/master/ruby.md#general), same below.

dsingleton:
LGTM, I would be happy to merge - though I think making `meta_tags` (by whatever name) part of the official API, with fixtures is worthwhile.

app/views/govuk_component/analytics_meta_tags.raw.html.erb
dsingleton: Does our rubocop gem catch bits like that? Might be worth adding to static?

app/views/govuk_component/analytics_meta_tags.raw.html.erb
dsingleton: The distinction is a bit fuzzy at the moment, but inside the "presenter" bit (eg, the complex processing of `content_item`) we can use any ERB features, as that is GOV.UK specific. We should avoid them in the "view" bit, which should be translatable in to other (simpler) languages, like mustache. The more ERB features we use in "views" the harder it is, as [mentioned before](https://github.com/alphagov/static/pull/669#discussion_r42742005).

As both the "presenter" and view" are in the same file, and we don't transpile the "views" yet, this is a bit obtuse. Sorry. Hopefully this will be covered better in some [component principles documentation](https://trello.com/c/ZyhODRnG/19-document-component-principles), when I have time to write it :(

In the mean time, `fetch` isn't a big deal in either place, as it's semantically equivalent to `hash[:key] || "default"`, right? So both are translatable. We've avoided `fetch` in (most) other view templates to avoid encouraging use of more complex usage, which should hopefully be bettered covered by the principles, when we have them.

So i'd be fine with `fetch` being used like this, but we should avoid using it in the view, or other "view only" components (which is most of them). 

Does that make sense?

benlovell: :+1: seems legit.

fofr:
@dsingleton Aside: We discussed creating a simple meta tags component, but weren't sure whether there was a need. It's fairly standard HTML that isn't going to change between apps.

dsingleton:
> @dsingleton Aside: We discussed creating a simple meta tags component, but weren't sure whether there was a need. It's fairly standard HTML that isn't going to change between apps.

That makes sense, it's a pretty GOV.UK specific usage anyway, so :+1: to not generalising it.

dsingleton:
:+1: to merging, once the syntax issue is cleaned up.

fofr:
I've updated to fix whitespace formatting.

dsingleton:
tumblr_inline_n8turylmea1qew7bs

PR #675, opened 2015-11-18
BAU: refactor organisation logo CSS

robinwhittleton:
Desktop mode uses a function to generate the background image, mobile code is hardcoded. This moves the mobile code into the same function.

### DIFF
--- a/app/assets/stylesheets/helpers/_organisations.scss
+++ b/app/assets/stylesheets/helpers/_organisations.scss
@@ -1,13 +1,13 @@
 @mixin organisation-logo-19 {
   font-family: $Helvetica-Regular;
   font-size: 13px;
-  line-height: (15/13);
+  line-height: (15 / 13);
   font-weight: 400;
   text-transform: none;
   text-decoration: none;
   color: $text-colour;
   display: block;
-  border-left: 2px solid #bbb;
+  border-left: 2px solid $grey-2;
   background: image-url("coa-default-18.gif") no-repeat 8px 0;
   height: auto;
   width: auto;
@@ -16,7 +16,7 @@
 
   @include media(tablet) {
     font-size: 19px;
-    line-height: (20/19); /* 20px */
+    line-height: (20 / 19);
     padding-top: 30px;
     background-image: image-url("coa-default-24.gif");
   }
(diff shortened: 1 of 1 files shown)


dsingleton:
SASS change looks good, have you tested on a page using those styles? (it looks like it's only [campaign pages in frontend](https://github.com/search?utf8=%E2%9C%93&type=Code&q=organisation-logo-+repo:alphagov/business-support-finder+repo:alphagov/calculators+repo:alphagov/calendars+repo:alphagov/collections+repo:alphagov/contacts-admin+repo:alphagov/contacts-frontend+repo:alphagov/courts-frontend+repo:alphagov/design-principles+repo:alphagov/email-alert-frontend+repo:alphagov/feedback+repo:alphagov/finder-frontend+repo:alphagov/frontend+repo:alphagov/government-frontend+repo:alphagov/homepage-frontend+repo:alphagov/info-frontend+repo:alphagov/licence-finder+repo:alphagov/manuals-frontend+repo:alphagov/smart-answers+repo:alphagov/specialist-frontend+repo:alphagov/whitehall)

alextea:
Could these changes also be applied to [whitehall](https://github.com/alphagov/whitehall/blob/master/app/assets/stylesheets/frontend/helpers/_organisations.scss)?

fofr:
It's hard to judge/test the affect this has. I would :heart: a style guide page that demonstrates these rules, this would allow wraith regression tests too.

dsingleton:
A styleguide would be useful. It would need to cover all the things on static not currently in components, which is pretty much [things from the wrapper layout](https://github.com/alphagov/static/blob/master/app/assets/stylesheets/static.scss), that we're trying to get rid of - by converting to components or moving directly into apps.

There's a question about how much effort a style guide would be vs the 'getting rid' work though.

fofr:
For the migration of HTML publications we're looking at porting organisation logos from Whitehall into government-frontend, this will probably involve creating a component.

With that in mind, I'm happy for this change to be merged.

app/assets/stylesheets/helpers/_organisations.scss
```diff
@@ -156,53 +156,33 @@
 /* Crests */
 
 @mixin organisation-logo-image($crest) {
-  background: image-url($crest) no-repeat 8px 0;
+  background: image-url('crests/#{$crest}_18px.png') no-repeat 8px 0;
+  @include media-down(mobile) {
```
gemmaleigh: Is it possible to swap these two, so the 13px crest is defined first and `@include media(tablet)` is used to increase the crest size for larger screens?

I'd like to remove `media-down` where we can.

robinwhittleton:
@gemmaleigh happy with this change?

gemmaleigh:
:+1: thanks @robinwhittleton.

PR #677, opened 2015-11-20
Add Spending review and autumn statement 2015 page to GA scroll depth…

zilnhoj:
… tracking


bishboria:
Merging for @erik-eide

PR #684, opened 2015-11-27
Add unit tests for all existing components

fofr:
https://trello.com/c/NLkm3EDy/17-unit-test-components

Includes:
* Beta label
* Govspeak
* Pagination
* Government navigation
* Document footer
* Option select
* Metadata

Also adds some test helpers for making future component unit tests easier, eg
```ruby
def assert_link_with_text(link, text)
   assert_select "a[href=\"#{link}\"]", text: text
end
```

cc @dsingleton @jamiecobbett


dsingleton:
Had a conversation with @fofr offline about the behaviour of components when not passed any arguments. The options were raising an exception, or not rendering anything.

We went with the later for now, to get a standard behaviour defined and revisit later.

PR #687, opened 2015-12-01
Add YouTube embed example to Govspeak Component

dsingleton:
Youtube links in govspeak will be converted into an embedded video
player.

Add an example of this to the fixtures for the govspeak component.

**Screenshot from local copy of `govuk-component-guide`:**

govspeak content with youtube embed gov uk component guide

### DIFF
--- a/app/views/govuk_component/docs/govspeak.yml
+++ b/app/views/govuk_component/docs/govspeak.yml
@@ -155,3 +155,7 @@ fixtures:
     rich_govspeak: true
     content: |
       <p>This content has some <strong>rich govspeak</strong></p>
+  with_youtube_embed:
+    content: |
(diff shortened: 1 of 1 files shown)


jamiecobbett:
At first I thought this was a feature of the Govspeak gem, but its actually a feature of the Javascript in static. Is it worth making that clear somehow?

boffbowsh:
:+1: but what Jamie said

fofr:
Added a commit to clarify that the embedded player comes via a script in static.

dsingleton:
Updated the documentation describing the Govspeak component itself, to clarify your comments @jamiecobbett and @boffbowsh. Does that help?

(@fofr I rolled your commit into mine, sorry)

PR #690, opened 2015-12-04
Update offer selector to match new eGain version

tombooth:
When the version of eGain changed it altered the selector that we need
to look for to see if the offer has been opened. This brings it in line
to what is in production.


robinwhittleton:
LGTM

PR #691, opened 2015-12-04
Reflect eGain changes in the other selectors used

tombooth:
The rest of the selectors also needed to be changed with the template
change at eGain's end.


issyl0:
Spoken to @tombooth in person. :-)

PR #692, opened 2015-12-11
Block deepcrawl bot

alexmuller:
I'd prefer to slow this bot down rather than blocking it completely but documentation for the bot doesn't indicate that it supports the crawl-delay property.

They describe themselves as a:

> site architecture analysis tool used by SEOs and web developers

Users won't be using this to find GOV.UK, so we can safely block it at the moment.


joshmyers:
Seems sane to me! :+1:

PR #694, opened 2015-12-22
Allow option-selects to send their 'closed state' with their form data

tombye:
This pull request is aimed at solving an issue found when using option-selects without AJAX.

### The problem

1. the option-select is in a form which posts to its parent page without the use of JavaScript (ie a search results page with option-selects used for filtering)
2. every time the form is submitted the option-selects do not remember whether they were closed or open when a new version of the page is returned

### Why it's only come up now

Option-selects are used on www.gov.uk in pages where, if the JavaScript sent runs, all form submissions use AJAX meaning the option-select HTML is not reloaded so this is not a problem.

### How this pull request attempts to solve it

These changes allow option-selects that do not use JavaScript for form submissions to preserve their 'closed state' by sending it in the form data.

A `closed_state_name` parameter can be added to the component data which is used by the option-select JavaScript to create a hidden `<input>` tag with that name. This is used to send the 'closed state' of the option-select in the form data.

image

image

Any help getting this to an acceptable state much appreciated @fofr, @dsingleton, @quis, @gemmaleigh.

### DIFF
--- a/app/assets/javascripts/govuk-component/option-select.js
+++ b/app/assets/javascripts/govuk-component/option-select.js
@@ -43,6 +43,11 @@
       if (this.$optionSelect.data('closed-on-load') == true) {
         this.close();
       }
+
+      var closedStateName = this.$optionSelect.data('closed-state-name');
+      if (closedStateName !== undefined) {
+        this.$closedStateField = this.attachStateField(closedStateName);
+      }
     }
   }
 
@@ -69,8 +74,20 @@
     this.$optionSelect.find('.js-container-head').append('<div class="js-selected-counter">'+this.checkedString()+'</div>');
   };
 
+  OptionSelect.prototype.attachStateField = function attachStateField(name){
+    var $closedStateField = $('<input type="hidden" name="'+name+'" value="'+this.isClosed()+'" />');
+    this.$optionList.append($closedStateField);
+    return $closedStateField;
+  }
+
   OptionSelect.prototype.updateCheckedCount = function updateCheckedCount(){
     this.$optionSelect.find('.js-selected-counter').text(this.checkedString());
+  };
+
+  OptionSelect.prototype.updateClosedStateField = function updateClosedStateField(){
+    if (typeof this.$closedStateField !== 'undefined'){
+      this.$closedStateField.val(this.isClosed().toString());
+    }
   };
 
   OptionSelect.prototype.checkedString = function checkedString(){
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/govuk-component/option-select.js
```diff
@@ -43,6 +43,11 @@
       if (this.$optionSelect.data('closed-on-load') == true) {
         this.close();
       }
+
+      var closedStateName = this.$optionSelect.data('closed-state-name');
+      if (closedStateName !== undefined) {
```
AndrewVos: https://api.jquery.com/jQuery.hasData/

AndrewVos:
I'm not sure I like that this just wouldn't work if you don't specify that closed state name parameter. Maybe it should be a mandatory or generated value?

app/assets/javascripts/govuk-component/option-select.js
fofr: `hasData` checks for any data set on an element, rather than data for a particular key. In this case, it would return a false positive when there is `closed-on-load` data but no `closed-state-name`.

tombye:
I might be mis-understanding but wouldn't it work without that parameter?

I did think of just adding the hidden `<input>` if you set `closed_state_name` to the partial data but that seemed a bit odd. Control over the closed state is added by JS so it seemed to make sense to have the communication of that state also added by JS.

Also, the reason I made it optional was because the main use of option-selects at the moment won't use it (because they use AJAX).

AndrewVos:
If control over the closed state is added by JS then is this code really needed?

What does closed actually mean? Do you have an example link?

AndrewVos:
By not working I mean the functionality you added only works if that data key gets set.

tombye:
This could sit outside of this component in a slightly different form (in the project JS of whatever applications use option-selects without AJAX).

My intention in adding it here was to grow the potential uses of this component by allowing a non-AJAX approach to be supported where the code that renders the option-selects can know if the user expects them closed or open.

If the value this adds to all users is worth it I think there's an argument for including it in the existing JS. That's part of the point of this pull request actually, to see if the maintainers feel it's worth adding here.

tombye:
Maybe I'm mis-reading things but I'm not sure I see what problems could occur from not setting the parameter in the partial data. My intention in having it as optional was to allow users who don't need this functionality to ignore it.

tombye:
I made a test-page in a copy of the components guide that might be useful:

https://github.com/tombye/option-select-test

The page is at the '/option-select-test' route.

(It assumes you're running a local version of static and running the apps in a VM.)

fofr:
@tombye Any progress on this, or can I close it?

tombye:
@fofr judging by this thread it doesn't look like there's a need for this functionality on GOVUK apps so I'm happy for you to close it.

Worth saying if this functionality is of interest to anyone in future, it exists on the Digital Marketplace version (viewable here: http://alphagov.github.io/digitalmarketplace-frontend-toolkit/forms/option-select.html).

PR #696, opened 2015-12-23
Add component unit test to component generator

fofr:
When generating a component include a unit test.

The basic unit test asserts that the component template renders with the expected `govuk-` class and that no errors are thrown when no parameters are passed in. It’s modelled on the alpha_label_test.

Includes documentation about component unit tests

https://trello.com/c/NLkm3EDy/17-unit-test-components

cc @dsingleton


dsingleton:
Huge improvement :+1:, especially to the docs. 

Any chance you could squash 66803de and 47321ce? I'd be happy merging without that, but it would make the history clean :)

PR #697, opened 2016-01-06
Fix HMRC webchat offer

robinwhittleton:
The iframe was relying on finding an element with a specific id in the offer markup loaded from KCOM. At some point recently they have updated their markup with a new id, causing the offer to fail to load.

This changes the code such that it now just looks for the first div inside the body rather than relying on a specific id which appears likely to change in future.

### DIFF
--- a/app/assets/javascripts/webchat-iframe.html.erb
+++ b/app/assets/javascripts/webchat-iframe.html.erb
@@ -11,7 +11,7 @@
     'use strict';
 
     function isOfferOpen() {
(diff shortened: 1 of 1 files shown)


fofr:
LGTM :+1:

Are there any smoke tests we could perform so that we get alerted to changes?
`body>div` could become `body>main` for instance, and this would break again.

robinwhittleton:
I imagine smoke tests would be hard to write, as the HTML is only sent if the offer is open.

daibach:
LGTM. Happy to merge.

PR #699, opened 2016-01-08
Warn IE10 users that they’re out of date

robinwhittleton:
Add a test for IE10’s user agent (https://blogs.msdn.microsoft.com/ie/2012/07/12/ie10-user-agent-string-update/) as it goes out of support on 12/1/2016.

cc @tombye

### DIFF
--- a/app/assets/javascripts/libs/suchi/isOld.js
+++ b/app/assets/javascripts/libs/suchi/isOld.js
@@ -41,20 +41,21 @@
    * of userbase at the time of writing.
    */
   suchi.laggards = {
-    // IE 6:      ?%
+    // IE 6:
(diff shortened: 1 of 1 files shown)


alextea:
Maybe you could duplicate this PR on the original repro as well https://github.com/slightlyoff/suchi.js

fofr:
For reference: In the last week IE10 has represented 2.2% of all visits to GOV.UK. That's less than IE9 (3.4%) and more than IE8 (2.0%).

robinwhittleton:
@alextea didn’t realise there was an original repo :) I’ll do a PR there now.

robinwhittleton:
@tombye want to give a final 👍/👎 on this?

tombye:
IE10 users are now officially unsupported by Microsoft (re. https://www.microsoft.com/en-us/WindowsForBusiness/End-of-IE-support) so this gets the :+1: from me. Nice work!

PR #702, opened 2016-01-08
Add two new paths where webchat is available

christophwong:
As requested by HMRC, these two outcome pages from smart answers need
webchat assistance. The only change here is making the script load the
iframe on those two paths in addition to the self-assessment page.

### DIFF
--- a/app/assets/javascripts/webchat.js.erb
+++ b/app/assets/javascripts/webchat.js.erb
@@ -9,7 +9,13 @@
   var webchat = {
 
     shouldOpen: function(){
-      return (window.location.pathname === '/government/organisations/hm-revenue-customs/contact/self-assessment');
+      var webchatEnabledPaths = [
+        '/government/organisations/hm-revenue-customs/contact/self-assessment',
+        '/check-if-you-need-a-tax-return/y/employed-even-if-just-for-part-of-the-year/no/no/no-neither-of-us-claimed-child-benefit/no/no/no/none-of-these',
+        '/check-if-you-need-a-tax-return/y/retired/no/no-neither-of-us-claimed-child-benefit/no/no/no/none-of-these'
+      ];
+
+      return (webchatEnabledPaths.indexOf(window.location.pathname) >= 0);
     },
 
     validOrigin: function(origin){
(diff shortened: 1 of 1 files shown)


christophwong:
This is tested locally by running frontend with `PLEK_SERVICE_STATIC_URI=http://127.0.0.1:3013 PLEK_SERVICE_CONTENTAPI_URI=https://www.gov.uk/api ./startup.sh' and `static`

While `this.GOVUK.webchat.shouldOpen()` returns`false` for most pages, it returns `true` for the two specific paths that are added in this PR.

fofr:
As far as I remember, some Smart Answers update the URL dynamically – so a user can click through their answers without the page reloading each time.

As `shouldOpen` runs on `init`, and not each time the URL updates I think users could reach the new URLs without the `shouldOpen` check running at the right point.

dsingleton:
> As far as I remember, some Smart Answers update the URL dynamically – so a user can click through their answers without the page reloading each time.

This is correct, [this is the SA JS](https://github.com/alphagov/smart-answers/blob/master/app/assets/javascripts/smart-answers.js).

The usefulness of the JS behaviour on SA has been debated, and we've considered removing it entirely to reduce complexity, as we don't believe it offers significant user benefit. But we've not got as far as concrete plans to remove it (CC @jennyd from @alphagov/team-custom-formats, who was involved in that conversation and own that part of the SA app).

daibach:
> The usefulness of the JS behaviour on SA has been debated

It'd be good to debate that again at some point, but we can do that later. 

For now, we need to make sure the webchat service loads on pages where this JS is used.

christophwong:
Is it still an issue at those two endpoints if the rendering app is 'frontend' instead of 'smartanswers'?

dsingleton:
You're totally correct, sorry about that. It's not an issue for Simple Smart Answers served by Frontend :+1:

dsingleton:
(that counts as a LGTM, I don't know if @robinwhittleton wants to cast an eye as he's more familiar with the webchat JS, otherwise i'd be happy merging)

robinwhittleton:
I was going to suggest `jQuery.inArray()` but it doesn’t look like [it doesn’t do anything extra apart from a null check](https://github.com/jquery/jquery/blob/master/src/core.js#L324) which we don’t need to worry about. So yes, 👍

PR #704, opened 2016-01-11
Related items component

binaryberry:
> __Headed sections of related items.__
> Accepts an array of sections. Each section can have a title, url, id and a list of related items.
> Each item is a hash with a title and url. If the url is external, a rel value can also be provided.

Create a related-items component based on188e526 and on [core stylesheets](https://github.com/alphagov/static/blob/master/app/assets/stylesheets/helpers/_core.scss#L56-L147)

Care must be taken when using the component so that navigation is always
correctly labelled (pattern based on http://tink.uk/screen-readers-aria-roles-html5-support/). When passing in a section id, use this to add an id attribute and a labelledby attribute. This is in favour of generating an id (eg. by parameterising the title) so that the component isn't using rails specific
features.

screen shot 2016-01-11 at 16 37 44

[Trello](https://trello.com/c/gRHBCfVy/236-component-creation-related-links-medium)

paired with @fofr 
cc @dsingleton

### DIFF
--- a/app/assets/stylesheets/govuk-component/_related-items.scss
+++ b/app/assets/stylesheets/govuk-component/_related-items.scss
@@ -20,7 +20,7 @@
     li {
       margin-bottom: 0.75em;
 
-      &.section {
+      &.related-items-more {
         font-weight: bold;
       }
 
--- a/app/assets/stylesheets/govuk-component/_related-items.scss
+++ b/app/assets/stylesheets/govuk-component/_related-items.scss
@@ -3,9 +3,7 @@
   padding-top: 5px;
 
   h2 {
-    @include core-24;
-    font-weight: 600;
-    color: $text-colour;
+    @include bold-24;
     margin-top: 0.3em;
     margin-bottom: 0.5em;
   }
@@ -13,9 +11,7 @@
   ul {
     @include core-16;
     list-style: none;
-    margin: 0 0 1.25em 0;
-    padding: 0;
-    overflow: hidden;
+    margin-bottom: 1.25em;
 
     li {
       margin-bottom: 0.75em;
--- a/app/assets/stylesheets/govuk-component/_related-items.scss
+++ b/app/assets/stylesheets/govuk-component/_related-items.scss
@@ -19,10 +19,6 @@
       &.related-items-more {
         font-weight: bold;
       }
-
-      a {
-        orphans: 2;
-      }
     }
   }
 }
--- a/app/assets/stylesheets/govuk-component/_related-items.scss
+++ b/app/assets/stylesheets/govuk-component/_related-items.scss
@@ -1,3 +1,32 @@
 .govuk-related-items {
+  border-top: 10px solid $mainstream-brand;
+  padding-top: 5px;
 
+  h2 {
+    @include core-24;
+    font-weight: 600;
+    color: $text-colour;
+    margin-top: 0.3em;
+    margin-bottom: 0.5em;
+  }
+
+  ul {
+    @include core-16;
+    list-style: none;
+    margin: 0 0 1.25em 0;
+    padding: 0;
+    overflow: hidden;
+
+    li {
+      margin-bottom: 0.75em;
+
+      &.section {
+        font-weight: bold;
+      }
+
+      a {
+        orphans: 2;
+      }
+    }
+  }
 }
--- a/app/assets/stylesheets/govuk-component/_related-items.scss
+++ b/app/assets/stylesheets/govuk-component/_related-items.scss
@@ -1,0 +1,3 @@
+.govuk-related-items {
+
+}
--- a/app/views/govuk_component/docs/related_items.yml
+++ b/app/views/govuk_component/docs/related_items.yml
@@ -1,0 +1,10 @@
+name: "Related items"
+description: "A short description of the Related items component"
+body: |
+  A long form description of the Related items component, which may
+  include markdown formatting.
+
+  It should try and describe the intent of the component, and document
+  any parameters passed to the component.
+fixtures:
+  default: {}
--- a/app/views/govuk_component/docs/related_items.yml
+++ b/app/views/govuk_component/docs/related_items.yml
@@ -1,10 +1,33 @@
 name: "Related items"
-description: "A short description of the Related items component"
+description: "Headed sections of related items."
 body: |
-  A long form description of the Related items component, which may
-  include markdown formatting.
+  Accepts an array of sections. Each section can have a title, url, id and a list of related items.
 
-  It should try and describe the intent of the component, and document
-  any parameters passed to the component.
+  Each item is a hash with a title and url. If the url is external, a rel value can also be provided.
 fixtures:
-  default: {}
+  default:
+    sections:
+    - title: 'Travel Abroad'
+      url: '/'
+      items:
+      - title: 'Link A'
+        url: /
+      - title: 'Link B'
+        url: /
+    - title: 'Travel'
+      items:
+      - title: 'Link 1'
+        url: /
+      - title: 'Link 2'
+        url: /
+  real_example:
+    # from https://www.gov.uk/register-birth
+    sections:
+    - title: Pregnancy and birth
+      url: /browse/childcare-parenting/pregnancy-birth
+      items:
+      - title: 'Register a birth'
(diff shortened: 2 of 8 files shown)


app/assets/stylesheets/govuk-component/_related-items.scss
```diff
@@ -0,0 +1,32 @@
+.govuk-related-items {
+  border-top: 10px solid $mainstream-brand;
+  padding-top: 5px;
+
+  h2 {
+    @include core-24;
+    font-weight: 600;
```
gemmaleigh: Is there a reason for using 600 here? 700 maps to bold and there's a bold mixin for 24px text which would allow these two lines to be combined.

`@include bold-24`  from the [front end toolkit's _typography.scss](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/stylesheets/_typography.scss#L114).

app/assets/stylesheets/govuk-component/_related-items.scss
```diff
@@ -0,0 +1,32 @@
...
+
+    li {
+      margin-bottom: 0.75em;
+
+      &.related-items-more {
+        font-weight: bold;
+      }
+
+      a {
+        orphans: 2;
```
tombye: Should this be in the main styles if it just applies to print CSS?

There's a print mixin you can use to just deliver print styles to the print stylesheet in [this Sass file](https://github.com/alphagov/static/blob/31f79a61ca5535ec805c02ff690616823cc32ad9/app/assets/stylesheets/styleguide/_print.scss#L20).

app/assets/stylesheets/govuk-component/_related-items.scss
fofr: Good spot – we should update this. It has been ported from: https://github.com/alphagov/static/blob/master/app/assets/stylesheets/helpers/_core.scss#L64

app/assets/stylesheets/govuk-component/_related-items.scss
```diff
@@ -0,0 +1,32 @@
...
+    margin-top: 0.3em;
+    margin-bottom: 0.5em;
+  }
+
+  ul {
+    @include core-16;
+    list-style: none;
+    margin: 0 0 1.25em 0;
+    padding: 0;
+    overflow: hidden;
```
gemmaleigh: I'm not sure `overflow: hidden` is needed here (as the child elements aren't floated), but if it is being used a clearfix, consider using the `contain-floats` [placeholder from the front end toolkit](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/stylesheets/_shims.scss#L45).

benlovell:
The ruby bits look good to me. :+1:

fofr:
@gemmaleigh @tombye I've simplified the styles in 7d9fa7e

I haven't added the print mixin as it appears to not yet be imported for components – weirdly the print mixin doesn't come from the toolkit as I expected.

dsingleton:
> I haven't added the print mixin as it appears to not yet be imported for components – weirdly the print mixin doesn't come from the toolkit as I expected.

Component print styles work differently, see https://github.com/alphagov/static/pull/638 (this is something we may want to re-examine, I can't remember the reason for going with this vs print mixin right now)

fofr:
@dsingleton @tombye Updated print styles in edd2e45

tombye:
Looks good to me then :+1:

app/views/govuk_component/docs/related_items.yml
```diff
@@ -0,0 +1,45 @@
...
+      - title: "Link 2"
+        url: /
+  external_links:
+    sections:
+      - title: "Elsewhere on the web"
+        id: "related-elsewhere-on-the-web"
+        items:
+        - title: "Wikivorce"
+          url: "http://www.wikivorce.com"
+          rel: external
```
dsingleton: Did you have other `rel` values in mind?

If not, it might be simpler to have a boolean `is_external` field instead.

app/views/govuk_component/related_items.raw.html.erb
```diff
@@ -0,0 +1,34 @@
+<%
+  sections ||= []
+%>
+<aside class="govuk-related-items" role="complementary">
+  <% sections.each do |section| %>
+    <% section_title = section[:title] || "Elsewhere on GOV.UK" %>
```
dsingleton: Should this be a default in the component, or up to the code calling the component to specify?

app/assets/stylesheets/govuk-component/_related-items.scss
```diff
@@ -0,0 +1,24 @@
+.govuk-related-items {
```
dsingleton: Does this give a standard whispace/margin to the bottom of the component? The `em` values don't seem to map to our usual `$gutter` variables, if we're porting them across, should we standardise?

dsingleton:
The component API looks good, markup/styling looks good.

Couple of small comments. The only one i'd want to resolve before merging is the use of `rel`. 

As discussed offline with @fofr having a list of objects, rather than an object of key/value (as we do for the metadata component) is a lot more robust. Arrays preserve order, unlike objects, and support additional values (eg, `rel` in this case). 

We've already run into problems with the later on the metadata component, so I think thats a pattern we should avoid for future components (and consider moving away from with metadata).

app/views/govuk_component/docs/related_items.yml
fofr: I preferred the flexibility of setting rel.

app/views/govuk_component/related_items.raw.html.erb
fofr: We discussed this and weren't sure what would be best so opted to maintain the status quo. The component would certainly be cleaner without it. What do you think?

app/assets/stylesheets/govuk-component/_related-items.scss
fofr: Yes, we should use standard gutters. I will update.

app/views/govuk_component/docs/related_items.yml
fofr: To expand, rel values that _might_ be relevant: `nofollow`, `prefetch`.

dsingleton: :+1: those make sense, possibly `rel="up"` too.

I was thinking that in a mustache environment, if we were to do something conditional on a specific value of `rel` then we'd have issues, but thats theoretical, and it wouldn't be a hard API chance to accommodate that.

app/views/govuk_component/related_items.raw.html.erb
dsingleton: Lets remove it. If we need to add it back when we use this on real GOV.UK pages we can.

app/assets/stylesheets/govuk-component/_related-items.scss
fofr: I've noticed our current margins are responsive, but the gutters are fixed pixels – updating to `$gutter` _feels_ like a regression.

app/views/govuk_component/related_items.raw.html.erb
fofr: If `Elsewhere on GOV.UK` gets passed in as a section title, then this part of the component (which was ported) reads poorly:
```
<a href="<%= section[:url] %>">
  More<% if section[:title] %> <span class="visuallyhidden">in <%= section_title %></span><% end %>
</a>
```

app/assets/stylesheets/govuk-component/_related-items.scss
dsingleton: Fair point, happy to merge with the pixel values.

I think theres a discussion to have about consistency across components, but it's not a blocker for this PR :+1:

app/views/govuk_component/related_items.raw.html.erb
dsingleton: Is that a [regression vs the old behaviour](https://github.com/alphagov/static/blob/master/app/views/root/related.raw.html.erb#L7-L34)? If not, then I think it's not a blocker, unless fixing it required a significant API change - like passing "Elsewhere" (internal, and [external](https://github.com/alphagov/static/blob/master/app/views/root/related.raw.html.erb#L37-L51) links as a separate arg?

PR #705, opened 2016-01-12
Change HMRC webchat endpoint

christophwong:
As of Jan 12 3pm the endpoint for webchat will be updated and we are
changing it on this end.


christophwong:
@daibach has looked at this and is ok with it. :+1:

PR #707, opened 2016-01-13
Change insertion point selector so webchat banner appears on new pages

christophwong:
Webchat has been working on 1 out of the 3 pages because only one has a
.beta-wrapper element. The headers also have different class names on
different pages so I decided to point the selector to a fairly
consistent #content element.

### DIFF
--- a/app/assets/javascripts/webchat.js.erb
+++ b/app/assets/javascripts/webchat.js.erb
@@ -50,7 +50,7 @@
       if (window.sessionStorage && window.postMessage) {
         if (webchat.shouldOpen()) {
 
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/webchat.js.erb
```diff
@@ -50,7 +50,7 @@
       if (window.sessionStorage && window.postMessage) {
         if (webchat.shouldOpen()) {
 
-          insertionPoint = $('.beta-wrapper').length ? $('.beta-wrapper') : $('.heading-block');
+          insertionPoint = $('.beta-wradpper').length ? $('.beta-wrapper') : $('#content > header');
```
fofr: There's a typo here: `wradpper`

christophwong:
@fofr thanks for reviewing

PR #708, opened 2016-01-13
Breadcrumb component should use toolkit styles

fofr:
The updated breadcrumb styles were added to the front-end toolkit in
https://github.com/alphagov/govuk_frontend_toolkit/pull/236

cc @dsingleton @gemmaleigh


gemmaleigh:
This looks great and works in the same way as the phase banner (alpha and beta label) components. :+1: 

I've updated [govuk elements to use the same breadcrumb styles from the front-end toolkit](https://github.com/alphagov/govuk_elements/pull/157) too.

PR #712, opened 2016-01-15
Validate analytics against schemas

benlovell:
Verifies the analytics component against the example case study document in the content schemas project. So far, we only integrate with a single case study example. Ideally we'd integrate with as many different examples as possible, but before this can happen we'll need a way of expressing just which capabilities or attributes each component requires from the example documents.

I'll complete the CI configuration to ensure that content schema commits are triggering builds over here, once this is merged.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -21,6 +21,7 @@ end
 
 group :test do
   gem 'capybara', '~> 2.5.0'
+  gem 'govuk-content-schema-test-helpers', '~> 1.4'
   gem 'mocha', '~> 1.1.0', :require => false
   gem 'shoulda'
   gem 'webmock'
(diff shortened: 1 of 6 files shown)


dsingleton:
Not looked at the implementation in detail, but :+1: to the overall idea. 

Currently `analytics_meta_tags` is the only component that takes a `ContentItem` argument, but I expect we'll be adding more (like `breadcrumb`, or `related_links`), so having a way to assert the behaviour of those arguments against real examples (vs [faked fixtures](https://github.com/alphagov/static/blob/d0b1556b2cb1613bdddc886fcf29ece496755a2a/app/views/govuk_component/docs/analytics_meta_tags.yml#L10-L20)) will give us more confidence on changes.

PR #714, opened 2016-01-20
Add a link to GOV.UK's Terms and Conditions to the footer

issyl0:
- It was advised that we include this link prominently in the (rare)
  event of legal action against us for something on the site.

Trello:
https://trello.com/c/cZrYVw6E/258-gov-uk-terms-and-conditions-need-to-appear-in-the-footer


brenetic:
:+1:

PR #716, opened 2016-01-21
Make 'Home' breadcrumb optional

tadast:
Allow passing in `include_home: false` to omit the first 'Home' breadcrumb.

Some formats e.g. Service Manual require the root breadcrumb to link to `/service-manual` rather than GOV.UK root.

### DIFF
--- a/app/views/govuk_component/breadcrumbs.raw.html.erb
+++ b/app/views/govuk_component/breadcrumbs.raw.html.erb
@@ -3,9 +3,6 @@
 %>
 <div class="govuk-breadcrumbs">
   <ol role="breadcrumbs">
-    <li>
-      <a href="/">Home</a>
-    </li>
   <% breadcrumbs.each do |crumb| %>
     <li>
       <% if crumb[:url] %>
--- a/app/views/govuk_component/docs/breadcrumbs.yml
+++ b/app/views/govuk_component/docs/breadcrumbs.yml
@@ -17,20 +17,32 @@ fixtures:
       url: '/section'
   many_breadcrumbs:
     breadcrumbs:
+    - title: 'Home'
+      url: '/'
     - title: 'Section'
       url: '/section'
     - title: 'Sub-section'
       url: '/section/sub-section'
     - title: 'Sub Sub-section'
       url: '/section/sub-section/sub-sub-section'
+  no_home:
+    breadcrumbs:
+    - title: 'Service Manual'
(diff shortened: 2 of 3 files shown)


dsingleton:
Could you include an example of that usage in the [breadcrumb fixtures](https://github.com/alphagov/static/blob/master/app/views/govuk_component/docs/breadcrumbs.yml) as well please? (they power the examples on the [Component Guide](http://govuk-component-guide.herokuapp.com/components/breadcrumbs)

tadast:
Thanks @dsingleton. Added and rebased the commit, have a look please.

app/views/govuk_component/breadcrumbs.raw.html.erb
```diff
@@ -1,11 +1,14 @@
 <%
   breadcrumbs ||= []
+  include_home = true if include_home.nil?
```
dsingleton: One of the longer term goals of components is to make them more generalised, usable on other GDS projects, in multiple templating languages (like govuk_template) *

I think the default-true arguments would be hard to convert into some of the simpler templating languages (like `mustache`). Could we switch it to something more explicitly negative, like `no_home=true`? 

( * think isn't documented on the repo/contributing guide yet, so it's not clear, we're going to be improving that soon )

dsingleton:
I'm happy with the feature change itself, and the supporting tests/examples are :+1:.

I've got one comment on the implementation, that i'll make on the line itself.

tadast:
@dsingleton I agree to your comment and have addressed it. Let me know if it looks good now.

dsingleton:
LGTM, CC @fofr for a second set of eyes (argument names are harder to change later, so worth a bit of extra review now)

fofr:
Would it make more sense to remove the `Home` first crumb default?

tadast:
Would this not break backwards compatibility in other applications using this component?

fofr:
Nothing is using this component yet.

tadast:
Alright, I'm fine with either, your call :)

tadast:
I've added another commit to this PR to support URLs like this:


See the commit note for explanation.

fofr:
I'd prefer the `Home` default to be removed. I've merged the other breadcrumb component so this needs a rebase.

tadast:
@fofr I've updated the commit. Can you please check and let me know if that's what you meant?

fofr:
Perfect :+1:

PR #717, opened 2016-01-22
Allow text-only breadcumbs

tadast:
User testing conducted by the Service Manual team has shown that users understand breadcrumbs better when the last breadcrumbs element references the current page, thus giving the user a mental anchor.

ksyzxtir-2016 01 21-17-03-01

Allow passing in a breadcrumb hash with no URL that renders it as a text element. Only the last breadcrumb element should be text-only. There are not checks in code to enforce it, so it's left to a developer of each application.

However we should discuss if we want to allow different application developers to deviate from what is currently on GOV.UK.

### DIFF
--- a/app/views/govuk_component/breadcrumbs.raw.html.erb
+++ b/app/views/govuk_component/breadcrumbs.raw.html.erb
@@ -8,7 +8,11 @@
     </li>
   <% breadcrumbs.each do |crumb| %>
     <li>
-      <a href="<%= crumb[:url] %>"><%= crumb[:title] %></a>
+      <% if crumb[:url] %>
+        <a href="<%= crumb[:url] %>"><%= crumb[:title] %></a>
+      <% else %>
+        <%= crumb[:title] %>
+      <% end %>
     </li>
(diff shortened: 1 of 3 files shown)


dsingleton:
Good commit, and tests :+1: 

I'm less sure on the feature side personally, but the commit documents the why really well. It'd be good to get some more input on that.

tadast:
Brad on slack:
> you should trust your own research, but support both use cases

tadast:
Not sure what's going on with the CI, Jenkins seems to be having issues

dsingleton:
:+1: OK.

This is a bit new (but good) ground for Components, which have been GOV.UK exclusive so far. With the long term plan for Components to be generalised GDS wide, ut's really useful to see how non-GOV.UK use can be.

dsingleton:
CI is having some general issues. I'm sure build is good, but we should hold off merging, as the release tags are generated by CI, and may not work right now :(

fofr:
Tests passing now :+1:

PR #718, opened 2016-01-22
Add an alias for 1/3 width column class

dsingleton:
The class for 2/3 columns is `column-two-thirds`, but the 1/3
column class is `column-third`, which has led to me using
`column-one-third` in templates a few times and being confused
why it doesn't work.

Add `column-one-third` as an alias, for ease of use.

Related: https://github.com/alphagov/govuk_elements/pull/127


fofr:
:+1:

fofr:
:+1:

PR #719, opened 2016-01-22
Temporary survey

benlovell:

Displays the temporary survey link between 26/01 and 27/01. The approach isn't particularly nice, but this is entirely temporary and will be reverted once the survey window closes.

This was previously done (#689) in the ERB template, but during that particular survey window we had some problems with caching. The trello card alludes to this, for reference.

### DIFF
--- a/app/assets/javascripts/user-satisfaction-survey.js
+++ b/app/assets/javascripts/user-satisfaction-survey.js
@@ -67,12 +67,21 @@
     setSurveyUrl: function(href) {
       var $surveyLink = $('#take-survey');
       var surveyUrl = $('#user-satisfaction-survey-container').data('survey-url');
+      var surveyStarts = new Date(2016, 1, 26).getTime();
+      var surveyEnds = new Date(2016, 1, 27, 23, 59, 59).getTime();
+
+      if (userSatisfaction.currentDate() >= surveyStarts &&
+          userSatisfaction.currentDate() <= surveyEnds) {
+        surveyUrl = 'https://www.surveymonkey.co.uk/r/2MRDLTW';
+      }
+
       if (surveyUrl.indexOf('?c=') === -1) {
         surveyUrl += "?c=" + root.location.pathname;
       }
 
       $surveyLink.attr('href', surveyUrl);
-    }
+    },
+    currentDate: function() { Date.now(); }
   };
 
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/user-satisfaction-survey.js
```diff
@@ -67,12 +67,21 @@
     setSurveyUrl: function(href) {
       var $surveyLink = $('#take-survey');
       var surveyUrl = $('#user-satisfaction-survey-container').data('survey-url');
+      var surveyStarts = new Date(2016, 1, 26).getTime();
+      var surveyEnds = new Date(2016, 1, 27, 23, 59, 59).getTime();
+
+      if (userSatisfaction.currentDate() >= surveyStarts &&
+          userSatisfaction.currentDate() <= surveyEnds) {
+        surveyUrl = 'https://www.surveymonkey.co.uk/r/2MRDLTW';
+      }
```
jamiecobbett: I wondered if we should move all the URLs into the JavaScript to keep them consistent. After all, I believe the banner is only ever displayed if JavaScript is available. Maybe @dsingleton or @fofr have an opinion?

benlovell: That would mean under normal circumstances it isn't configurable by consumers. We did discuss this briefly offline earlier and the recommendation was to take this route.

I guess given the transient nature of this particular survey these things haven't been given too much consideration. Shrug.

dsingleton: @benlovell chatted to @fofr and I about this. We were both happy with one URL being in the JS, given this is a very short term thing.

We're generally moving towards [the pattern of JS being logic-only, and pulling configuration from markup](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/docs/javascript.md#use-data-attributes-for-configuration), so if this were a longer-lived implementation, I'd suggest putting both URLs in markup.

dsingleton:
LGTM.

fofr:
This has been merged with the wrong date.
`new Date(2016, 1, 26)` is Feb 26. (JS dates are 0-indexed)

app/assets/javascripts/user-satisfaction-survey.js
jamiecobbett: :+1:

PR #720, opened 2016-01-25
Alter dates for temporary survey period

benlovell:

There was an issue with the first implementation in that the date portions were ordinal and didn't take into account JavaScript's months being zero-based. This resolves that issue and another issue regarding `Date.now()` IE8 incompatibility.


fofr:
:clock10: :+1:

PR #723, opened 2016-01-28
Run survey between 2016/02/01 and 2016/05/01

benlovell:

The temporary survey will now run between these two dates.


jamiecobbett:
Did one or both of you test this by eg changing your computer's clock and visiting the page?

PR #724, opened 2016-01-29
Add button attribute to button element

carolinegreen:
When this element is used with forms, pressing enter on a textbox within the same form triggers a synthetic mouselick event on the nearest button in the form.

Adding type button to the button element overrides the button default type of submit within the form.


fofr:
:+1:

I didn't know about the `type="button"` attribute for `<button>`. This is a nice fix.

fofr:
Could we add a comment by the line too? Without knowledge of the `type="button"` overriding the default submit action the step of adding a button type to a button might be confusing.

carolinegreen:
Yes, thanks.  I've added a comment now.

PR #725, opened 2016-02-01
Fix date-dependent tests

benlovell:
Now we're in the survey window, these tests were failing as the
alternative URL is mandated. These modifications ensure the tests run
outside the temporary date window.


fofr:
:+1:

PR #729, opened 2016-02-04
Update govuk_frontend_toolkit gem to v4.7.0

erikse:
Bump the gem version to enable click tracking in Google analytics of across gov.uk pages.

Changelog - [govuk_frontend_toolkit](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/CHANGELOG.md#470)
```
4.7.0

 * Add 'mailto' tracking to GOV.UK Analytics (PR #244)
```


fofr:
This alone doesn't enable the mailto link tracking. A line needs to be added here: https://github.com/alphagov/static/blob/master/app/assets/javascripts/analytics/static-analytics.js#L24

erikse:
Do we restore the branch and revert the merge, or do a new PR to add the line?

fofr:
I would leave this as-is, it's a gem bump that has an effect because it also includes `4.6.1`. The commit didn't include context, so it's not incorrect.

I'd create a new branch, add the feature and test that the mailto link tracking is working and that the data you want to track is appearing correctly in Google Analytics before raising the PR.

PR #731, opened 2016-02-10
Update scroll-tracker.js

vp2015:
add scroll tracking to select headers on /government/publications/see-potential-case-studies-and-guidance-for-employers/see-potential-case-studies-and-guidance-for-employers


erikse:
Just verified the URL has all those headings, looks ok

PR #733, opened 2016-02-15
Increase margins between text in govspeak component

fofr:
* Use 20px margins (15px at thinner viewports) to match GOV.UK elements. This improves the legibility of content that contains many medium to long paragraphs, it comes at the cost of legibility for content with groups of one line paragraphs, mainly specialist content.
* Add spacing between list items, 5px – matching GOV.UK elements
* Use bold mixins
* Clean up call to action styles

cc @dsingleton @gemmaleigh @alextea 

# Before and after examples
(The more spaced out examples in the GIFs are the new margins)

## Travel advice
* Better legibility on content with long paragraphs

spacing-travel-advice 1

## Mobile
* Clearer spacing at thinner viewports

spacing-mobile

## Specialist content
* Better positioning of call to action when it's the first piece of content
* Better padding on call to action

spacing-specialist-content

### DIFF
--- a/app/assets/stylesheets/govuk-component/_govspeak.scss
+++ b/app/assets/stylesheets/govuk-component/_govspeak.scss
@@ -23,7 +23,7 @@
 
   h2 {
     @include bold-27;
-    margin-top: $gutter-half;
+    margin-top: $gutter;
 
     @include media(desktop) {
       margin-top: $gutter * 1.5;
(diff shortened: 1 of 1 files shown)


dsingleton:
Huge improvement :+1: 

It makes body copy much more readable, and moves us closer to consistency between Whitehall/Mainstream styling.

alextea:
This looks great. Especially the list items.

fofr:
## Before
screen shot 2016-02-16 at 11 54 35

## After
screen shot 2016-02-16 at 11 53 40

dsingleton:
I'd be happy with this getting merged, but should we leave this open a bit longer to get more reckons?

gemmaleigh:
:+1: Nice work @fofr!

fofr:
@markhurrell pointed out that Whitehall doesn't add margins to list items. As frontend and elements both do this we're keeping the 5px margin for now, and when things are consistent we can re-asses and address all at once.

PR #737, opened 2016-02-19
Remove invalid ARIA role

fofr:
`role="breadcrumbs"` is not a thing.
https://www.w3.org/TR/wai-aria/roles#role_definitions

cc @alextea @robinwhittleton

### DIFF
--- a/app/views/govuk_component/breadcrumbs.raw.html.erb
+++ b/app/views/govuk_component/breadcrumbs.raw.html.erb
@@ -2,7 +2,7 @@
   breadcrumbs ||= []
 %>
 <div class="govuk-breadcrumbs">
-  <ol role="breadcrumbs">
+  <ol>
   <% breadcrumbs.each do |crumb| %>
     <li>
       <% if crumb[:url] %>


robinwhittleton:
Should we be using the pattern in the accessibility sandbox? https://github.com/alphagov/govuk_accessibility_sandbox/blob/master/app/views/includes/breadcrumb_examples.html

app/views/govuk_component/breadcrumbs.raw.html.erb
```diff
@@ -2,7 +2,7 @@
   breadcrumbs ||= []
 %>
 <div class="govuk-breadcrumbs">
-  <ol role="breadcrumbs">
```
jamiecobbett: :bread: :roller_coaster:

dsingleton:
Any reason not to use `navigation`, as per the Sandbox? We should use that, or just ship this, either one would be an improvement.

robinwhittleton:
@aduggin , do you have an opinion here?

fofr:
We haven't gone back to this to improve the breadcrumbs component _yet_, but I've found this PR from 2013 that specifically removed the `nav` elements citing accessibility tests:
https://github.com/alphagov/static/pull/238

PR #738, opened 2016-02-19
Add starting of govuk_toolkit modules

brenetic:
Static will now start any govuk_toolkit modules available on the page.

Further info about the modules can be found [here](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/docs/javascript.md#modules).


fofr:
Looks good. Have you tested that this starts the module in your app correctly?

brenetic:
@fofr We have indeed. Works as it should.

PR #739, opened 2016-02-19
Fallback to static host if GOVUK_ASSET_ROOT not set

dsingleton:
This allows be run more easily in more situations;

- On the VM, but not under `bowl`, eg, `bowl frontend --without static` and
  running static in another tab (which is common).
- Outside the VM, setting `PLEK_SERVICE_STATIC_URI=http://localhost:3013`,
  which is also a common use pattern outside the VM.

Ideally `static` would generate absolute, including host, URLs for
assets and avoid needing the extra additional Plek service ENV var, but
this fixes the two issues described above for now

CC @erikse who was having this issue too, and @alext who I discussed the fix with.

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -25,9 +25,7 @@ Static::Application.configure do
   # rev filenames for assets
   config.assets.digest = false
 
-  if ENV['GOVUK_ASSET_ROOT'].present?
-    config.action_controller.asset_host = ENV['GOVUK_ASSET_ROOT']
-  end
+  config.asset_host = ENV['GOVUK_ASSET_ROOT'] || Plek.current.find('static')
 
   # Expands the lines which load the assets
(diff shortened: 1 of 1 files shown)


boffbowsh:
This is functionally equivalent to https://github.com/alphagov/plek/blob/master/lib/plek.rb#L92, so should be replaced with `Plek.current.asset_root`

alext:
> This is functionally equivalent to https://github.com/alphagov/plek/blob/master/lib/plek.rb#L92, so should be replaced with Plek.current.asset_root

It is equivalent at the moment, but that Plek function is incorrect - it doesn't work for apps other than static. The dev fallback in Plek would more correctly be `assets-origin`.

erikse:
Just tested this branch towards a local instance of smart answers, looks great :+1:

PR #741, opened 2016-02-22
Add WH attachment styling to Govspeak component

dsingleton:
As part of migration we are moving formats from WH to `content-store`
and `government-frontend`. Some of the formats in WH have `body`
content that includes embedded attachments, eg, markup describing
an attachment in the rendered HTML of a document from content-store.

To render this in a new-world frontend app, like government-frontend
the `govspeak` component, that renders a documents `body` content,
needs to be able to add styling and behaviour for the attachment
markup in the `body`.

This PR adds examples of attachments within govspeak, and ports the
CSS/JS that renders them/adds toggling behaviour, to the govspeak
component on static.

- Includes changes to work with accessible JS toggle module [1]

[1] https://github.com/alphagov/whitehall/pull/2476

Note, the CSS to styling attachment govspeak is still duplicated in
WH and at some point will need to be reconsiled with the component
version, but that needs to be done for all govspeak styling, not
just attachments. But for now we're adding the ability for the
component version to render attachments at all, to unblock the
migration of formats.

### DIFF
--- a/app/assets/stylesheets/govuk-component/_govspeak.scss
+++ b/app/assets/stylesheets/govuk-component/_govspeak.scss
@@ -413,4 +413,132 @@
   .media-player {
     @include media-player;
   }
+
+  // This block is duplicated from Whitehall as a transitional step, see the
+  // commit adding this warning before making changes, original version:
+  // https://github.com/alphagov/whitehall/blob/master/app/assets/stylesheets/frontend/helpers/_attachment.scss
+  $thumbnail-width: 99px;
+
+  .attachment {
+    position: relative;
+    margin: $gutter 0;
+    padding: $gutter-half 0 0 ($thumbnail-width + $gutter);
+    @extend %contain-floats;
+
+    &.inline {
+      background: transparent;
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/govuk-component/_govspeak.scss
```diff
@@ -413,4 +413,132 @@
...
+  // https://github.com/alphagov/whitehall/blob/master/app/assets/stylesheets/frontend/helpers/_attachment.scss
+  $thumbnail-width: 99px;
+
+  .attachment {
+    position: relative;
+    margin: $gutter 0;
+    padding: $gutter-half 0 0 ($thumbnail-width + $gutter);
+    @extend %contain-floats;
+
+    &.inline {
```
fofr: Seems crazy for the inline version to inherit from `.attachment` when it's effectively just a link, and these styles are resetting everything just set.

dsingleton:
I'm missing an example of an attachment, so i'm going to close this and re-open once updated.

PR #740, opened 2016-02-22
Add a JS toggle module

dsingleton:
Most of the work done by @fofr. I cherry-picked it off his branch and made a slight tweak in the second commit (which @fofr was happy with).

This is the first use of [JS Modules](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/docs/javascript.md#modules) at the `static` level of GOV.UK, though the [`start`](https://github.com/alphagov/static/pull/738) code was added to support it's use at an application level.

This particular module is for toggling visibility of content and is being added to [support the accessibility toggle on attachments](https://trello.com/c/HLb48j22/280-govspeak-implementing-the-ui-of-attachments-medium), which i'll be adding the `govspeak` component in a follow up PR.

Example usage of the general pattern:
```
<div data-module="toggle">
    <a href="#" class="my-toggle" data-expanded="false" data-controls="target">Toggle</a>
    <div id="target">Target</div>
</div>
```

_Includes a [bump to `govuk_template`](https://github.com/alphagov/static/commit/942118322958115b0c3e7dbfe3d564a33f8531bf)_

**Mergable when:**

- [x] https://github.com/alphagov/govuk_template/pull/203 is merged
- [x] A new version of govuk_template release
- [x] The gem version of govuk_template is updated on this branch

PR #742, opened 2016-02-23
Add 'embedded attachments' to Govspeak component

dsingleton:
As part of migration we are moving formats from WH to `content-store` and `government-frontend`. Some of the formats in WH have `body` content that includes embedded attachments, eg, markup describing an attachment in the rendered HTML of a document from content-store.

To render this in a new-world frontend app, like government-frontend the `govspeak` component, that renders a documents `body` content, needs to be able to add styling and behaviour for the attachment markup in the `body`.

This PR adds examples of attachments within govspeak, and ports the CSS/JS that renders them/adds toggling behaviour, to the govspeak component on static.

- Includes changes to work with accessible JS toggle module

Note, the CSS to styling attachment govspeak is still duplicated in WH and at some point will need to be reconsiled with the component version, but that needs to be done for all govspeak styling, not just attachments. But for now we're adding the ability for the component version to render attachments at all, to unblock the migration of formats.

Also note, the static images used to represent attachments without their own cover image, eg doc, csv, html pub, have full URLs which point at the WH-frontend app's assets. This is the approach we're taking short term, long term the image assets should belong to something within the FE stack (most likely static). The approach of storing fully rendered attachment HTML in the content store is a interim approach to support migration, and will be replaced with a system where the data for an attachment is passed to a frontend, with the body content containing a placeholder, and something in the FE stack owning the template to render the attachment, and inserting the output into the body for rendering. Keeping the responsibility/ownership of attachment template and images within the FE stack, not partially owned by Whitehall.

**Mergable once these PRs are merged**

- [x] https://github.com/alphagov/static/pull/740 - Provides the `toggle` JS module needed for toggling accessibility content.
- [x] https://github.com/alphagov/whitehall/pull/2476 - Decides the markup we should expect/support in govspeak `body` content

govspeak content preview gov uk component guide

### DIFF
--- a/app/assets/stylesheets/govuk-component/_govspeak.scss
+++ b/app/assets/stylesheets/govuk-component/_govspeak.scss
@@ -33,19 +33,6 @@
   h3 {
     @include bold-19;
     margin-top: $gutter + 5px;
-
-    &.hosted-externally {
-      @include bold-27;
-      padding-top: 2px;
-
-      a {
-        text-decoration: none;
-
-        &:hover {
-          text-decoration: underline;
-        }
-      }
-    }
   }
 
   h4 {
@@ -425,15 +412,9 @@
     padding: $gutter-half 0 0 ($thumbnail-width + $gutter);
     @extend %contain-floats;
 
-    &.inline {
-      background: transparent;
-      position: static;
-      padding: 0;
-      margin: 0;
-
-      &:after {
-        display: none;
-      }
+    &:first-child {
+      margin-top: 0;
+      padding-top: 0;
     }
 
     .attachment-thumb {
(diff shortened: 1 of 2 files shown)


dsingleton:
This is blocked on a decision on `.js-hidden` in https://github.com/alphagov/whitehall/pull/2476, otherwise it's reviewable.

app/assets/stylesheets/govuk-component/_govspeak.scss
```diff
@@ -33,19 +33,6 @@
...
-      padding-top: 2px;
-
-      a {
-        text-decoration: none;
-
-        &:hover {
-          text-decoration: underline;
-        }
-      }
-    }
```
fofr: Really pleased to see this removed :+1:

app/assets/stylesheets/govuk-component/_govspeak.scss
```diff
@@ -413,4 +400,103 @@
   .media-player {
     @include media-player;
   }
+
+  // This block is duplicated from Whitehall as a transitional step, see the
+  // commit adding this warning before making changes, original version:
+  // https://github.com/alphagov/whitehall/blob/master/app/assets/stylesheets/frontend/helpers/_attachment.scss
+  $thumbnail-width: 99px;
+
+  .attachment {
```
fofr: I think the `.attachment` styles could be split out into a separate file and referenced here, much like the media player.

app/views/govuk_component/docs/govspeak.yml
```diff
@@ -241,3 +241,164 @@ fixtures:
...
+    content: |
+      <p>testing my attachment <span class="inline-attachment" id="attachment_1399340">
+        <a href="/government/uploads/system/uploads/attachment_data/file/498071/PHE_Payments_over__25k_Jun_15.csv">testing</a>
+        (<span class="type"><abbr title="Comma-separated Values">CSV</abbr></span>, <span class="file-size">65.4KB</span>)</span> works in the middle of copy.
+      </p>
+  block_attachments:
+    content: |
+      <section class="attachment embedded" id="attachment_1399345">
+        <div class="attachment-thumb">
+            <a aria-hidden="true" class="thumbnail" href="https://www.gov.uk/government/publications/minimum-wage-jobs-in-ashfield-local-authority-since-2009"><img alt="" src="https://assets.digital.cabinet-office.gov.uk/government/uploads/system/uploads/attachment_data/file/496127/thumbnail_LPC_FoI_09.12.15_NMW_jobs_Ashfield_LA.pdf.png"></a>
```
fofr: 4 space indentation here.

fofr:
Some minor comments but otherwise :+1:, and nothing that should block this PR from being merged.

Parts of the attachment styles are a bit gnarly, and it'd be nice to get them cleaned up, but while we want to maintain consistency with Whitehall that probably doesn't make sense right now.

app/views/govuk_component/docs/govspeak.yml
fofr: * Do we need an example with an attachment between two paragraphs?
* Can the examples appear above the long specialist example?

dsingleton:
I agree splitting out `.attachment` SCSS into a helper, and improving examples, but i'd like to do those in follow up PRs, to get this out now. OK to merge?

boffbowsh:
Description says merging of this relies on https://github.com/alphagov/whitehall/pull/2476

dsingleton:
> Description says merging of this relies on alphagov/whitehall#2476

Merged now!

dsingleton:
Going to close this as i've missed some [print specific SCSS for attachments](https://github.com/alphagov/whitehall/blob/5631a1722e186b194f4f7bb1f53cd2eb56e48034/app/assets/stylesheets/frontend/print/_document.scss). Will re-open once thats added.

dsingleton:
Superseded by https://github.com/alphagov/static/pull/745, which includes print support.

PR #743, opened 2016-02-24
Add scheduled maintenance view

alexmuller:
This file is currently [hardcoded in the Puppet repo][1] which is definitely not the correct place for it for lots of reasons. It should be a view in static similar to our error views.

[1]: https://github.com/alphagov/govuk-puppet/blob/d717279c28008e54c0cabab2bd058306301344a8/modules/loadbalancer/templates/usr/share/nginx/html/maintenance_page.erb

### DIFF
--- a/app/views/root/scheduled_maintenance.html.erb
+++ b/app/views/root/scheduled_maintenance.html.erb
@@ -1,0 +1,13 @@
+<%
+# This view is fetched by Puppet so that the loadbalancer can serve a nice
+# error when it's put into maintenance mode (to switch off the backend).
+%>
+<%= render partial: 'error_page', locals: {
+  status_code: 503,
+  title: 'Scheduled maintenance',
+  heading: 'GOV.UK publishing tools are currently unavailable due to scheduled maintenance',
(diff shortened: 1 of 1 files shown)


boffbowsh:
Can you provide context on how this is accessed and enabled?

dsingleton:
+1 to more context, possible a comment/note that it's fetched from static by something external and that moving the path it would cause problems?

alexmuller:
I've added a comment, is that enough context or should I put more in?

boffbowsh:
Where in puppet? I searched for `scheduled_maintenance` and didn't find anything.

alexmuller:
Sorry, it's not there yet. It needs to exist here first and then it will be used as part of the [`router::errorpage` defined type](https://github.com/alphagov/govuk-puppet/blob/04671de3f00e6c5a53557baffe1204830c209bc6/modules/router/manifests/errorpage.pp).

dsingleton:
Once it exists a little follow up PR to add the cross linking would be :star2:

PR #745, opened 2016-02-25
Govspeak attachments

dsingleton:
**Successor to https://github.com/alphagov/static/pull/742, which I couldn't re-open due to a force push :disappointed:**

As part of migration we are moving formats from WH to `content-store` and `government-frontend`. Some of the formats in WH have `body` content that includes embedded attachments, eg, markup describing an attachment in the rendered HTML of a document from content-store.

To render this in a new-world frontend app, like government-frontend the `govspeak` component, that renders a documents `body` content, needs to be able to add styling and behaviour for the attachment markup in the `body`.

This PR adds examples of attachments within govspeak, and ports the CSS/JS that renders them/adds toggling behaviour, to the govspeak component on static.

- Includes changes to work with accessible JS toggle module

Note, the CSS to styling attachment govspeak is still duplicated in WH and at some point will need to be reconsiled with the component version, but that needs to be done for all govspeak styling, not just attachments. But for now we're adding the ability for the component version to render attachments at all, to unblock the migration of formats.

Also note, the static images used to represent attachments without their own cover image, eg doc, csv, html pub, have full URLs which point at the WH-frontend app's assets. This is the approach we're taking short term, long term the image assets should belong to something within the FE stack (most likely static). The approach of storing fully rendered attachment HTML in the content store is a interim approach to support migration, and will be replaced with a system where the data for an attachment is passed to a frontend, with the body content containing a placeholder, and something in the FE stack owning the template to render the attachment, and inserting the output into the body for rendering. Keeping the responsibility/ownership of attachment template and images within the FE stack, not partially owned by Whitehall.

**Mergable once these PRs are merged**

- [x] https://github.com/alphagov/static/pull/740 - Provides the `toggle` JS module needed for toggling accessibility content.
- [x] https://github.com/alphagov/whitehall/pull/2476 - Decides the markup we should expect/support in govspeak `body` content

govspeak content preview gov uk component guide

### DIFF
--- a/app/assets/stylesheets/govuk-component/_govspeak-print.scss
+++ b/app/assets/stylesheets/govuk-component/_govspeak-print.scss
@@ -2,4 +2,33 @@
   .play-container {
     display: none;
   }
+
+  .attachment {
+    margin: $gutter 0;
+
+    &:first-child {
+      margin-top: 0;
+    }
+
+    .attachment-thumb {
+      display: none;
+    }
+
+    .attachment-details {
+      .download {
+        display: block;
+      }
+    }
+
+    .accessibility-warning {
+      h2 {
+        @include core-14;
+      }
+
+      .toggler {
+        display: none;
+      }
+    }
+  }
+
 }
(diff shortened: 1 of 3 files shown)


dsingleton:
And RTL example:
screenshot 2016-02-25 13 55 44

boffbowsh:
:+1: :shipit:

PR #746, opened 2016-02-25
Improve Govspeak component print styles

dsingleton:
**Media player**

| Before | After |
| --- | --- |
|  screenshot 2016-02-25 15 00 39 | screenshot 2016-02-25 15 01 03 |

This is a lot less, but there's not a lot of benefit to including any part of the player in print.

**Callouts**

| Before | After |
| --- | --- |
| screenshot 2016-02-25 14 59 52 | screenshot 2016-02-25 14 59 12 |


boffbowsh:
:+1:

robinwhittleton:
I like it.

PR #748, opened 2016-02-26
Implement 48 hour test User Satisfaction Survey

boffbowsh:
Trello: https://trello.com/c/L9Lk1l4H/308-survey-test-for-start-of-march-small-ish

This promotes the existing test to be the default, and sets up a new test for 48 hours beginning midnight UTC on 2nd March 2016.

This test includes a copy change which will also be reverted at the end of the test.

Best reviewed commit-by-commit due to some refactoring.

### DIFF
--- a/app/assets/javascripts/user-satisfaction-survey.js
+++ b/app/assets/javascripts/user-satisfaction-survey.js
@@ -67,11 +67,8 @@
     setSurveyUrl: function(href) {
       var $surveyLink = $('#take-survey');
       var surveyUrl = $('#user-satisfaction-survey-container').data('survey-url');
-      var surveyStarts = new Date("March 2, 2016").getTime();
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/user-satisfaction-survey.js
```diff
@@ -81,6 +87,13 @@
 
       $surveyLink.attr('href', surveyUrl);
     },
+    inTestPeriod: function() {
```
jamiecobbett: This function doesn't appear to be tested any longer - it is stubbed out in the tests. If that's the case, are we comfortable with that?

boffbowsh: Well spotted. I meant to add a test for it and got distracted. Will add now.

boffbowsh: Added.

PR #752, opened 2016-02-29
Correcting header typo on scroll-tracking.js

vp2015:
Correcting typo on heading 'When you can claim Universal Credit' on /guidance/universal-credit-how-it-helps-you-into-work which prevented the scroll event from firing.

### DIFF
--- a/app/assets/javascripts/analytics/scroll-tracker.js
+++ b/app/assets/javascripts/analytics/scroll-tracker.js
@@ -65,7 +65,7 @@
       ['Percent', 75],
       ['Heading', 'Opening up work'],
       ['Heading', 'Support from your work coach'],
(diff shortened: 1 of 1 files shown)


erikse:
I wonder if in the future, it would be worth adding the links into the PR description, so we can click through and verify the target page exist, and that the headings exist.

For example:
 [URL on gov.uk](https://www.gov.uk/guidance/universal-credit-how-it-helps-you-into-work#when-you-can-claim-universal-credit)
  * Add scroll tracking to heading "When you can claim Universal Credit"

Even though it doesn't verify the heading text is matching, it will make it easier for the reviewer of the PR to go to the target page and check

vp2015:
Yep good idea, will do from now on

PR #753, opened 2016-03-01
Update Rails to security patched version 4.2.5.2

erikse:
Updates Rails to the latest version to address `[CVE-2016-2098]` for Possible remote code execution vulnerability in Action Pack, and `[CVE-2016-2097]` for Possible Information Leak Vulnerability in Action View.

Both vulnerabilities are exploited by code that allows a carefully set value in `params[:id]` when passed to the render method, which static doesn’t do, with access to `params[:template]`:
```ruby
  def raw_root_template
    render_raw_template("root", params[:template])
  end

  def raw_govuk_component_template
    render_raw_template("govuk_component", params[:template])
  end
```
actions in the `root_controller`, rendering files, eg:
```ruby
  private

  def render_raw_template(prefix, file_name)
    file_path = Rails.root.join("app", "views", prefix, "#{file_name}.raw.html.erb")
    error_404 and return unless File.exists?(file_path)
    render :text => File.read(file_path)
  end
```
and access to `params[:locale]`:
```ruby 
  def govuk_locales
    return error_404 unless params[:locale].match(/^[a-z]{2}(-[a-z0-9]{2,3})?$/)
    render_yaml_as_json("locales", "#{params[:locale]}.yml")
  end
```
in the `root_controller` action, being sanitised with a regex.

ChangeLog details for affected dependencies in Rails 4.2.5.2:
 * [Action Pack](https://github.com/rails/rails/blob/4-2-stable/actionpack/CHANGELOG.md#rails-4252-february-26-2016)
 * [Action View](https://github.com/rails/rails/blob/4-2-stable/actionview/CHANGELOG.md#rails-4252-february-26-2016)


benlovell:
Thanks @erikse :+1:

PR #755, opened 2016-03-08
Fix IE8 JavaScript error on GOV.UK

robinwhittleton:
IE8 is throwing an error on all of GOV.UK (reported by @maxf). Looking at the given line/column number reveals a use of `indexOf` in the webchat code. This isn’t supported by IE8, hence it’s throwing an error. We’ve got jQuery available, so the fix is to swap in `$.inArray()`.

### DIFF
--- a/app/assets/javascripts/webchat.js.erb
+++ b/app/assets/javascripts/webchat.js.erb
@@ -16,7 +16,7 @@
         '/government/organisations/hm-revenue-customs/contact/self-assessment-online-services-helpdesk'
       ];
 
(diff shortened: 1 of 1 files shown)


robinwhittleton:
My dev VM isn’t very happy, so could someone pull and test this before merging? I’ve manually checked it in IE8 but better safe than sorry.

erikse:
I've just checked this in a Win 7/IE8 VirtualBox and the JS console if free of errors for the homepage running locally.

PR #756, opened 2016-03-15
Add scroll tracking to Budget HTML doc

vp2015:
HMT have requested to have % scroll tracking on the html doc version of the budget to be published on 16/03/16


erikse:
@vp2015 - this looks fine, would be good to double check the URL in publisher if possible.

PR #757, opened 2016-03-15
Add sticky-element-container module.

tvararu:
This module allows you to stick an element to the bottom of its parent.

This is part of migrating `html_publications` from `whitehall` to `government-frontend`, and is meant to be used to implement the little "&uarr; Contents" button on http://www.gov.uk/government/publications/dwp-single-departmental-plan-2015-to-2020/dwp-single-departmental-plan-2015-to-2020.

Another PR will be opened to `government-frontend` to update the content item template and pass in the correct parameter to use this. (Update: https://github.com/alphagov/government-frontend/issues/122)

Relevant Trello ticket: https://trello.com/c/cJGrtnc1/286-5-html-publications-migration-front-end-work-large

Previous implementation in Whitehall: https://github.com/alphagov/whitehall/blob/master/app/assets/javascripts/application/back-to-content.js

### DIFF
--- a/app/assets/javascripts/modules/sticky-element-container.js
+++ b/app/assets/javascripts/modules/sticky-element-container.js
@@ -1,0 +1,136 @@
+/*
+  This module will cause a child in the target element to:
+  - hide when the top of the target element is visible;
+  - stick to the bottom of the window while the parent element is in view;
+  - stick to the bottom of the target when the user scrolls past the bottom.
+
+  Use 'data-module="sticky-element-container"' to instantiate, and add
+  `[data-sticky-element]` to the child you want to position.
+*/
+(function (Modules, root) {
+  'use strict';
+
+  var $ = root.$;
+  var $window = $(root);
+
+  Modules.StickyElementContainer = function () {
+    var self = this;
+
+    self._getWindowDimensions = function _getWindowDimensions () {
+      return {
+        height: $window.height(),
+        width: $window.width()
+      };
+    };
+
+    self._getWindowPositions = function _getWindowPositions () {
+      return {
+        scrollTop: $window.scrollTop()
+      };
+    };
+
+    self.start = function start ($el) {
+      var $element = $el.find('[data-sticky-element]');
+      var _hasResized = true;
+      var _hasScrolled = true;
+      var _interval = 50;
+      var _isDesktop = true;
+      var _windowVerticalPosition = 1;
+      var _startPosition, _stopPosition;
+
+      initialise();
+
+      function initialise () {
+        $window.resize(onResize);
+        $window.scroll(onScroll);
+        setInterval(checkResize, _interval);
+        setInterval(checkScroll, _interval);
+        checkResize();
+        checkScroll();
+        $element.addClass('govuk-sticky-element');
+      }
+
+      function onResize () {
+        _hasResized = true;
+      }
+
+      function onScroll () {
+        _hasScrolled = true;
+      }
+
+      function checkResize () {
+        if (_hasResized) {
+          _hasResized = false;
+          _hasScrolled = true;
+
+          var windowDimensions = self._getWindowDimensions();
+          _isDesktop = windowDimensions.width > 768;
+          _startPosition = $el.offset().top;
+          _stopPosition = $el.offset().top + $el.height() - windowDimensions.height;
+
+          var isMobile = !_isDesktop;
+          if (isMobile) {
+            unstick();
+          }
+        }
+      }
+
+      function checkScroll () {
+        if (_hasScrolled) {
+          _hasScrolled = false;
+
+          _windowVerticalPosition = self._getWindowPositions().scrollTop;
+
+          if (_isDesktop) {
+            updateVisibility();
+            updatePosition();
+          } else {
+            show();
+            unstick();
+          }
+        }
+      }
+
+      function updateVisibility () {
+        var isPastStart = _startPosition < _windowVerticalPosition;
+        if (isPastStart) {
+          show();
+        } else {
+          hide();
+        }
+      }
+
+      function updatePosition () {
+        var isPastEnd = _stopPosition < _windowVerticalPosition;
+        if (isPastEnd) {
+          stickToParent();
+        } else {
+          stickToWindow();
+        }
+      }
+
+      function unstick () {
+        $element.removeClass('govuk-sticky-element--stuck-to-parent');
+        $element.removeClass('govuk-sticky-element--stuck-to-window');
+      }
+
+      function stickToWindow () {
+        $element.addClass('govuk-sticky-element--stuck-to-window');
+        $element.removeClass('govuk-sticky-element--stuck-to-parent');
+      }
+
+      function stickToParent () {
+        $element.addClass('govuk-sticky-element--stuck-to-parent');
+        $element.removeClass('govuk-sticky-element--stuck-to-window');
+      }
+
+      function show () {
+        $element.removeClass('govuk-sticky-element--hidden');
+      }
+
+      function hide () {
+        $element.addClass('govuk-sticky-element--hidden');
+      }
+    };
+  };
+})(window.GOVUK.Modules, window);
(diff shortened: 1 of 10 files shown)


boffbowsh:
This looks ok to me, but I am not a JS developer

dsingleton:
Overall, :+1: to the feature. Having a generic way to sticky content is really useful, as it's a pattern we use in a few formats - and it'll be easier/quicker to experiment with sticky-content on other formats in the future.

Discussed offline: The commit structure is logical and easy to read, but the two big commits could do with more context in the commit message. Some good bits would be an explanation of why we're using BEM-style syntax in some places, the change to the z-index behaviour for the sticky content (vs whitehall) and the helpers around `window` to aid testing. I think that'll be useful for someone reviewing this, and someone in the future looking for historic context of the change.

app/views/govuk_component/govspeak_html_publication.raw.html.erb
```diff
@@ -2,7 +2,16 @@
   govspeak_locals = { content: content }
   govspeak_locals[:direction] = local_assigns.fetch(:direction) if local_assigns.include?(:direction)
   govspeak_locals[:rich_govspeak] = local_assigns.fetch(:rich_govspeak) if local_assigns.include?(:rich_govspeak)
+  sticky_footer_html = local_assigns.fetch(:sticky_footer_html) if local_assigns.include?(:sticky_footer_html)
```
dsingleton: As we've added a new argument to the `govspeak_html_publication` there should be a [corresponding fixture](https://github.com/alphagov/static/blob/master/app/views/govuk_component/docs/govspeak_html_publication.yml) to appear in the component guide.

tvararu:
@dsingleton fixing those now. :ok_hand:

tvararu:
All fixed.

app/views/govuk_component/govspeak_html_publication.raw.html.erb
tvararu: Fixed in 7978476.

benlovell:
LGTM :shipit:

dsingleton:
:tada:

app/views/govuk_component/govspeak_html_publication.raw.html.erb
```diff
@@ -2,7 +2,16 @@
...
 %>
-<div class="govuk-govspeak-html-publication">
+<div
+  class="govuk-govspeak-html-publication"
+  <%= "data-module=sticky-element-container" if sticky_footer_html %>
+>
   <%= render file: 'govuk_component/govspeak.raw', locals: govspeak_locals %>
+  <% if sticky_footer_html %>
```
fofr: I'm not sure that the sticky element should be integrated into this component. I think I'd have preferred the JS module to wrap the component call here: https://github.com/alphagov/government-frontend/blob/b5b6e9ccd68b52e426a8471827d123e7fabe82ab/app/views/content_items/html_publication.html.erb#L27-L32

fofr:
I'm sorry this feedback is coming after the merge.

It looks like this module has introduced a new JS style. eg Multiple `var` declarations, use of `_` to indicate private variables, `self = this` rather than `bind`, spacing between function name and parentheses `onScroll ()`, BEM style classes `govuk-sticky-element--stuck-to-window`.

For `govuk-sticky-element--stuck-to-parent` I think `&.stuck-to-parent` would be clearer – we can use Sass features to avoid the need for BEM style `--` class naming.

Regarding `govuk-sticky-element--hidden`, did the existing `.hidden`, `.js-hide` or `.visuallyhidden` classes not work?

app/views/govuk_component/govspeak_html_publication.raw.html.erb
fofr: The module could then be limited to `government-frontend` until it's needed by more than one frontend app.

app/assets/javascripts/modules/sticky-element-container.js
```diff
@@ -0,0 +1,136 @@
...
+      var _isDesktop = true;
+      var _windowVerticalPosition = 1;
+      var _startPosition, _stopPosition;
+
+      initialise();
+
+      function initialise () {
+        $window.resize(onResize);
+        $window.scroll(onScroll);
+        setInterval(checkResize, _interval);
```
fofr: Why `setInterval` rather than a `setTimeout` that sets itself to run again once it's reached the end?

If `checkResize` or `checkScroll` regularly takes longer than 50ms to run then the CPU might spike and the UI become less responsive (consider a low CPU mobile device). Similarly, if any of the code in those functions errors, the error would fire every 50ms, rather than stopping after the first error. For more details: http://stackoverflow.com/questions/729921/settimeout-or-setinterval

tvararu: Good points!

`setInterval` was carried over from the previous implementation (https://github.com/alphagov/whitehall/blob/master/app/assets/javascripts/application/back-to-content.js); I know about the perf trade-offs between it and `setTimeout`.

I actually wanted to use a `requestAnimationFrame` (with a `_.throttle` on the callback) call instead of either an interval or a timeout, but since we had no polyfill in place for it I decided to just leave it as is for now.

If you want to go full operational reasoning then there are some other perf wins; such as the window event listeners that are added for every call of this module and aren't delegated.

Since 99% of the time this interval callback will be called it's going to noop due to the fact that it hits a false conditional head-on, my hunch is that it won't cause performance issues.

Thanks again for the feedback. :+1:

app/assets/javascripts/modules/sticky-element-container.js
```diff
@@ -0,0 +1,136 @@
...
+    };
+
+    self._getWindowPositions = function _getWindowPositions () {
+      return {
+        scrollTop: $window.scrollTop()
+      };
+    };
+
+    self.start = function start ($el) {
+      var $element = $el.find('[data-sticky-element]');
```
fofr: This might be clearer if `$element` was `$stickyElement`. I confused it with `$el` that gets passed in to `start()`.

tvararu:
Thanks for the feedback @fofr, it would have been great to discuss these last week with you, as I wanted more input on the code styling choices.

The project doesn't have any `eslintrc` in-place, so I took the "ask for forgiveness" route and used [semistandard](https://github.com/Flet/semistandard). The reason I did is because it's reasonably well known and it was the least overhead for me (I just flipped a toggle in my editor). I would have picked [standard](http://standardjs.com/index.html) but I know that would not have gone down well due to the lack of semicolons. :)

In terms of code styling discussions I'm of the opinion that most of it is bikeshedding, and I don't care which way or the other we pick our semicolons, spaces, parantheses or variable naming conventions. However, I want it to be **consistent** and **enforced through very strict linting**. I'll gladly set up eslint or help set it up.

> I think `&.stuck-to-parent` would be clearer – we can use Sass features to avoid the need for BEM style `--` class naming.

I disagree here. The BEM class name is the whole point. Using a `.stuck-to-parent` rule would pollute the global CSS namespace, allowing other modules to change its properties by way of the cascade. E.g.

```
// module.css
.module.stuck-to-parent {
  position: fixed;
}

// some-other-library.css
.stuck-to-parent {
  position: relative !important;
}

// or-like-this.css
div[class].stuck-to-parent {
  position: relative !important;
}
```
It's arguable that `.stuck-to-parent` is sufficiently specific that the likelihood of that is very low. That being said, why risk it, and why introduce additional mental overhead over picking names that are less likely to conflict when you can just use BEM. Sass already helps you make it plenty DRY.

More thoughts on the subject welcome.

fofr:
We haven't had problems with a polluted global CSS namespace. The risk has typically been mitigated by scoping using Sass. It doesn't feel like a problem we need to solve, if solving it would mean re-writing the existing styles.

If however we find some demonstrable examples of this being a real world problem then we should reconsider.

tvararu:
> It doesn't feel like a problem we need to solve

But we do, at least for modules/components that are going to be used in foreign codebases. They should be written defensively and work well in all environments. BEM is a good fit and spares me the brain cycles and uncertainty of figuring out if `.my-module.hidden` is a good idea or not.

> if solving it would mean re-writing the existing styles

We don't need to do this. Rather, we can tighten our linters (https://github.com/brigade/scss-lint/blob/master/config/default.yml#L165-L167) and make it so that it won't be necessary to bikeshed class naming conventions for newly written code.

PR #759, opened 2016-03-21
Change URLs where HMRC web chat can be displayed

daibach:
HMRC have asked for their web chat trial to show on various tax credit related pages. 

At the same time, the self assessment trial has finished so we should remove those URLs.

### DIFF
--- a/app/assets/javascripts/webchat.js.erb
+++ b/app/assets/javascripts/webchat.js.erb
@@ -10,10 +10,15 @@
 
     shouldOpen: function(){
       var webchatEnabledPaths = [
-        '/government/organisations/hm-revenue-customs/contact/self-assessment',
-        '/check-if-you-need-a-tax-return/y/employed-even-if-just-for-part-of-the-year/no/no/no-neither-of-us-claimed-child-benefit/no/no/no/none-of-these',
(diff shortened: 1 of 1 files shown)


daibach:
cc. @robinwhittleton / @christophwong

robinwhittleton:
Can’t talk about whether the URLs are correct or not, but there’s no syntactic changes that’ll cause any problems there.

dsingleton:
This looks OK.

When adding new URLs it's worth knowing that WebChat will only work on applications using `wrapper` layout - https://trello.com/c/lReRyYJ4/38-include-hmrc-webchat-scss-in-all-static-layouts-not-just-wrapper

christophwong:
To add to @dsingleton 's point, the code looks right, as that's the whitelist array for any new urls.
However, these two urls wouldn't work:
- '/childcare-costs-for-tax-credits/y/yes/yes/weekly_diff_amount'
- '/childcare-costs-for-tax-credits/y/no/regularly_less_than_year/monthly_diff_amount'

As they don't have either the correct matching element for the [js to insert the webchat banner](https://github.com/alphagov/static/blob/change-hmrc-webchat-urls/app/assets/javascripts/webchat.js.erb#L59) as there is an extra element between the div#content and the nested header element.

daibach:
@christophwong Thanks. I'll look at fixing those ones in a separate PR.

daibach:
This is ready to be merged now

PR #760, opened 2016-03-21
Define .crest-eo for organisation_logos.

tvararu:
This will match the executive office crest styling to the single identity one, which will bring html publications on `government-frontend` more in line with the ones on `whitehall`.

Part of fixing this visual regression:

prime-minister s-office

Relevant Trello ticket: https://trello.com/c/cJGrtnc1/286-5-html-publications-migration-front-end-work-large


fofr:
Can we add a test fixture for the component to cover this case?

tvararu:
No automated tests to update since it's adding some additional CSS properties to a classname, but relevant documentation updated.

PR #761, opened 2016-03-22
Change govspeak header margin styling.

tvararu:
This brings it more in line with what is on whitehall, while respecting our new vertical rhythm conventions.

Before:

heading-padding-1
heading-padding-2
heading-padding-3

After (still outstanding differences due to added line-height consistency in government-frontend):

screen shot 2016-03-22 at 14 54 14

Relevant Trello ticket:
https://trello.com/c/cJGrtnc1/286-5-html-publications-migration-front-end-work-large

### DIFF
--- a/app/assets/stylesheets/govuk-component/_govspeak-html-publication.scss
+++ b/app/assets/stylesheets/govuk-component/_govspeak-html-publication.scss
@@ -92,6 +92,23 @@
         }
       }
     }
+
+    .stat-headline:first-child {
+      margin-top: $gutter;
+
+      @include media(tablet) {
+        margin-top: ($gutter * 2) + $gutter-two-thirds;
+      }
+    }
+
+    h2,
+    h3 {
+      margin-top: $gutter;
+
+      @include media(tablet) {
+        margin-top: ($gutter * 2) + $gutter-two-thirds;
+      }
+    }
   }
 
   .sticky-element {
--- a/app/assets/stylesheets/govuk-component/_govspeak.scss
+++ b/app/assets/stylesheets/govuk-component/_govspeak.scss
@@ -334,7 +334,13 @@
   }
 
   .stat-headline {
-    margin-bottom: $gutter;
+    margin-bottom: $gutter-half;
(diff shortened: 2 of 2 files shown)


fofr:
Could you include some before and after screenshots showing the affect of this change?

tvararu:
I already did, CMD+R. :+1:

fofr:
This change will affect all govspeak content. Headings on HTML publications in Whitehall are currently inconsistent with the rest of GOV.UK – by using the govspeak component it makes all headings consistent.

I'm not sure this is a necessary change.

tvararu:
I recognise that, but @dsingleton pointed out that they could use some specialised styling because of how dense their information is.

If that means scoping it to just HTML publications, I'll update the PR. If we decide against fragmenting the styling for consistency, then feel free to close this one.

fofr:
> I recognise that, but @dsingleton pointed out that they could use some specialised styling because of how dense their information is.

Cool, it'd be good to capture things like that in the PR to provide context.

If we're keeping this change it should probably live in https://github.com/alphagov/static/blob/master/app/assets/stylesheets/govuk-component/_govspeak-html-publication.scss

fofr:
We should check this with @markhurrell and @JenAllum to see if it's a design feature we want to keep.

tvararu:
Once we get feedback from them, I can update the PR's contents and fix the rules.

JenAllum:
Just discussed - retain the wider spacing for now. Format consolidation on long form text will review the approach later, more thoroughly.

tvararu:
@fofr updated. :white_check_mark:

fofr:
Code LGTM. I think the commit message could provide more detail: 
 * stat-headline now matches vertical rhythm of other block elements like paragraphs to improve readability
 * headings have increased spacing on HTML pubs to help break-up long format text and give natural resting points

ie. explain the differences that need the code, rather than describing the process of matching something that lives elsewhere

I think this will help future code spelunkers.

tvararu:
Rebased and updated.

PR #762, opened 2016-03-23
Add RTL support to html publications.

tvararu:
Part of the addressing the final unintended changes introduced with migrating HTML publications in https://trello.com/c/cJGrtnc1/286-5-html-publications-migration-front-end-work-large.

This fixes the positioning of the sticky element. The reason it's done on the parent element is because directly messing with the positioning of the element produces undesirable results.

### Before

screen shot 2016-03-23 at 19 06 18

### After

screen shot 2016-03-23 at 19 02 19

### DIFF
--- a/app/assets/stylesheets/govuk-component/_govspeak-html-publication.scss
+++ b/app/assets/stylesheets/govuk-component/_govspeak-html-publication.scss
@@ -5,6 +5,15 @@
   margin-bottom: $gutter * 1.5;
   z-index: 2;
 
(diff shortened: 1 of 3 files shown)


fofr:
Code looks good.

Minor comment: I'm wondering if it's worth indicating in the code somehow that this is necessary only for the correct RTL rendering of the sticky element. A developer could misinterpret this as duplication of what happens in the govspeak component.

tvararu:
Good thought, I'll add a comment for that.

fofr:
:+1:

PR #764, opened 2016-03-30
Add additional margin-top to govspeak contacts.

tvararu:
Makes them align better when around headers and other items which don't have an implicit margin-bottom.

Updates documentation with example where this change is visible.

Part of migrating HTML publications in https://trello.com/c/cJGrtnc1/286-5-html-publications-migration-front-end-work-large.

### Before

screen shot 2016-03-30 at 11 13 31

### After

screen shot 2016-03-30 at 11 14 23

### DIFF
--- a/app/assets/stylesheets/govuk-component/_govspeak.scss
+++ b/app/assets/stylesheets/govuk-component/_govspeak.scss
@@ -357,6 +357,7 @@
   .contact {
     @extend %contain-floats;
     margin-bottom: $gutter;
+    margin-top: $gutter;
     position: relative;
 
     .content {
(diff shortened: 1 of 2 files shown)


fofr:
How does this look on thinner viewports? I wonder if the smaller mobile spacing will look odd with the full gutter top margin.

tvararu:
Like this:

screen shot 2016-03-30 at 11 28 24

Could be halved, your call.

tvararu:
Discussed offline with @fofr: we are going to remove the bottom margin on the comments paragraph at the end of the contact card on desktop.

tvararu:
Updated!

fofr:
:+1:

PR #765, opened 2016-03-30
Allow change history to be expanded with JS

AndrewVos:
~~REVIEWERS: I'm not really happy with the way I've written these tests, and would appreciate any ideas how I can make them a bit better. Also, I'm not 100% happy with looping through the items twice in the view, but perhaps this has to be done because each set of items should be in a different unordered list.~~

This exists because we are going to show page history in Service Manual guide pages: https://trello.com/c/qzHbv9S9/181-show-change-notes-and-page-history-at-the-bottom-of-guidance-page

Allow change history to be expanded with JS

When using the document_footer component, you can now pass in
`expand_full_history`.

This will cause only the first history item to be shown, and a link will
allow you to show the rest of the history items. This uses javascript
toggle.

## Before expand has been clicked:

image

## After expand has been clicked:

image

[preview this locally with govuk_component_guide](http://localhost:3113/components/document_footer/fixtures/with_expandable_change_history/preview)

### DIFF
--- a/app/views/govuk_component/document_footer.raw.html.erb
+++ b/app/views/govuk_component/document_footer.raw.html.erb
@@ -8,6 +8,9 @@
   history ||= []
   history = Array(history)
 
+  hidden_history ||= []
+  history = Array(history)
+
   other ||= nil
 
   other_dates ||= nil
@@ -32,14 +35,31 @@
     <% if history.any? %>
       <div class="change-notes" id="history">
         <h3>Page history</h3>
-        <ol>
-          <% history.each do |change| %>
-            <li>
-              <time datetime="<%= change[:timestamp] %>" class="timestamp"><%= change[:display_time] %></time>
-              <%= change[:note] %>
-            </li>
-          <% end %>
-        </ol>
+        <% if history.any? %>
+          <ol>
+            <% history.each do |change| %>
+              <li>
+                <time datetime="<%= change[:timestamp] %>" class="timestamp"><%= change[:display_time] %></time>
+                <%= change[:note] %>
+              </li>
+            <% end %>
+          </ol>
+        <% end %>
+        <% if hidden_history.any? %>
+          <div data-module="toggle">
+            <a href="#change-notes-hidden-history" data-expanded="false" data-controls="change-notes-hidden-history">
+              + full page history
+            </a>
+            <ol id="change-notes-hidden-history" class="js-hidden">
+              <% hidden_history.each do |change| %>
+                <li>
+                  <time datetime="<%= change[:timestamp] %>" class="timestamp"><%= change[:display_time] %></time>
+                  <%= change[:note] %>
+                </li>
+              <% end %>
+            </ol>
+          </div>
+        <% end %>
       </div>
     <% end %>
   </div>
--- a/app/views/govuk_component/docs/document_footer.yml
+++ b/app/views/govuk_component/docs/document_footer.yml
@@ -69,3 +69,15 @@ fixtures:
     published: 20 January 2012
     from: <a href='/government/organisations/ministry-of-defence'>Ministry of Defence</a>
     part_of: <a href='/government/topics/energy'>Energy</a>
+  with_expandable_change_history:
+    history:
+    - display_time: 22 January 2012
+      timestamp: 22-01-2012Z14:19:00T
+      note: We updated the document
+    hidden_history:
+    - display_time: 24 January 2013
+      timestamp: 24-01-2012Z14:19:00T
+      note: We updated the document again
+    - display_time: 24 January 2014
+      timestamp: 24-01-2012Z14:19:00T
+      note: We updated the document again
--- a/test/govuk_component/document_footer_test.rb
+++ b/test/govuk_component/document_footer_test.rb
@@ -94,6 +94,39 @@ class DocumentFooterTestCase < ComponentTestCase
     assert_select '.change-notes li', count: 2
   end
 
+  test "renders document hidden history" do
+    render_component(
+      expand_full_history: true,
+      history: [
+          display_time: "26 October 2016",
+          timestamp: "26-10-2016Z14:19:00T",
+          note: "We updated the document"
+      ],
+      hidden_history: [
+        {
+          display_time: "22 January 2012",
+          timestamp: "22-01-2012Z14:19:00T",
+          note: "We updated the document"
+        },
+        {
+          display_time: "24 January 2012",
+          timestamp: "24-01-2012Z14:19:00T",
+          note: "We updated the document again"
+        }
+      ]
+    )
+
+    assert_timestamp_in('.change-notes #change-notes-hidden-history li', '22-01-2012Z14:19:00T', '22 January 2012')
+    assert_select '.change-notes #change-notes-hidden-history li', text: /We updated the document$/
+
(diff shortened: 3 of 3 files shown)


dsingleton:
For any change to a component it's good to include [updated fixtures](https://github.com/alphagov/static/blob/master/app/views/govuk_component/docs/document_footer.yml#L3-L71) and screenshots to demonstrate the changes - and makes them easier to review. Could you add those and i'll have a look at the actual changes?

AndrewVos:
@dsingleton Thanks, doing that now.

AndrewVos:
@dsingleton Updated docs and added two screenshots

fofr:
Having a toggle that shows content _above_ the link feels counterintuitive.

We have this pattern elsewhere:
screen shot 2016-03-30 at 17 10 27
screen shot 2016-03-30 at 17 10 31

https://www.gov.uk/government/consultations/property-ownership-and-public-contracting-by-foreign-companies-improving-transparency

AndrewVos:
@fofr I tend to agree. I'll move it.

dsingleton:
+1 to be consistent with existing GOV.UK patterns, even if they're not a component yet, they'll have been battle tested/researched/etc. It'd be worth checking if there are any other similar patterns of page history on other parts of GOV.UK too.

app/views/govuk_component/document_footer.raw.html.erb
```diff
@@ -30,9 +30,21 @@
...
+        <% rest_of_history = history[1..-1] %>
+        <% history = [history.first] %>
+      <% else %>
+        <% rest_of_history = [] %>
+      <% end %>
+
       <div class="change-notes" id="history">
         <h3>Page history</h3>
-        <ol>
+        <ol id="change-notes-history">
```
dsingleton: We try avoid `id`s in components as it can create specificity problems, and issues with the the component guide where multiple instances of a component are rendered on the same page.

(_`id="history"` is an exception as it's used as an anchor link, though that should probably be configurable as an optional parameter_).

It looks like the new `id` is to support testing, could that be re-written as a more meaningful class instead?

fofr: I thought the same, but I think these IDs are needed for `aria-` support and the toggle.

app/views/govuk_component/document_footer.raw.html.erb
```diff
@@ -30,9 +30,21 @@
       <% end %>
     <% end %>
     <% if history.any? %>
+      <% expand_full_history =
```
dsingleton: This is quite a logic for a component - we try and keep the logic inside components fairly straightforward, sticking to templating logic that could be replicated by simpler templating languages like `mustache`, as a longer term goal of GOV.UK Components is to make them available to wider government, and keeping them simpler will make it easier cross-compile to other languages (in theory). 

This tends to keep components simpler and easier to reason about too, which is nice. In this situation the list manipulation/splitting could be avoided by having a two `history` parameters, something like `history` and `history_hidden` (roughly). This would simplify the template and makes the component's programatic interface more explicit, and would support showing more than one piece of history as a default - but it does put more work on to the app calling the component to do that work, which I don't think is too bad. 

What do you think?

app/views/govuk_component/document_footer.raw.html.erb
```diff
@@ -40,6 +52,21 @@
             </li>
           <% end %>
         </ol>
+        <% if rest_of_history.any? %>
+          <div data-module="toggle">
+            <a href="#change-notes-rest-of-history" data-expanded="false" data-controls="change-notes-rest-of-history">
+              Show full history
+            </a>
+            <ol id="change-notes-rest-of-history" class="js-hidden">
```
dsingleton: Similar to the comment above, it would be good for this to be an optional parameter

dsingleton:
Added a few comments, hopefully they make sense, if not i'm happy to chat about them.

Some of the things I mentioned should definitely be documented, because they're not intuitive right now, and i'm hoping to find time to do that soon, so apologies :)

app/views/govuk_component/document_footer.raw.html.erb
dsingleton: Isn't `change-notes-rest-of-history` required for `aria-` support, rather than `change-notes-history`?

app/views/govuk_component/document_footer.raw.html.erb
AndrewVos: :100:

Sounds good to me!

AndrewVos:
@dsingleton Thanks. If you let me know the things you want documented I'll add them in when I get a chance!

dsingleton:
> @dsingleton Thanks. If you let me know the things you want documented I'll add them in when I get a chance!

No worries - it's actually on a team backlog to make space/time for me to do it, but i'd appreciate feedback when I do :)

app/views/govuk_component/document_footer.raw.html.erb
AndrewVos: @dsingleton Yes exactly. The latter was only added for testing, and I do feel bad about it :)

AndrewVos:
@dsingleton I've made the changes you suggested in a second commit, though please don't merge this because I would prefer squashing that commit with the former.

app/views/govuk_component/document_footer.raw.html.erb
```diff
@@ -29,17 +29,38 @@
         <p><%= title %>: <span class="other-date definition"><%= date %></span></p>
       <% end %>
     <% end %>
-    <% if history.any? %>
+    <%
```
dsingleton: This could be moved up to the top of the template, with the other argument defaults (and probably use the same default assignment style)

app/views/govuk_component/document_footer.raw.html.erb
```diff
@@ -29,17 +29,38 @@
...
+      hidden_history =
+          local_assigns.include?(:hidden_history) ? hidden_history : []
+    %>
+    <% if history.any? || hidden_history.any? %>
```
dsingleton: This would support having a page history header, but no (default) visible history. I think `history` should be required if providing `hidden_history` - or if we decide against that, then there should fixture covering that example.

dsingleton: Also, how that'd fit with the [history expansion example we have on whiteall](https://www.gov.uk/government/consultations/property-ownership-and-public-contracting-by-foreign-companies-improving-transparency)?

dsingleton:
Core Formats team have some [work in progress to migration detailed guides](https://trello.com/c/k6H9gNKF/316-6-detailed-guides-migration-front-end-work-small), which will need a pretty similar pattern for expanding change history ([example](https://www.gov.uk/guidance/apply-for-a-tier-4-sponsor-licence#history)). 

It'd be good for someone working on that to review this and make sure it supports both service manual and the version being migrated - consolidating any differences in the process. That'll most likely be @tvararu or @binaryberry.

@AndrewVos: That work should be happening soon, but let us know if waiting for Core to look at the component is going to hold you up?

dsingleton:
Had a chat offline and we agreed the easiest way forward will be for the service manual to duplicate the bits of `document_footer` it needs in it's frontend and modify them there - and not be dependant on the component right now.

This gives service manual team a bit more freedom, and core formats can add the functionality we need separately.

PR #769, opened 2016-04-15
[Do not merge] Add a banner to most of GOV.UK for the EU Referendum

boffbowsh:
There are two parts to this:
* The bar itself, its styles and some JS logic that counts how many times it's been seen
* JS logic that determines whether the bar should show, it runs in the header and toggles a CSS class before any styles are run (so the whole page doesn't jump). It's minified and inline, and doesn't use any libraries.
  * We've created a test spec that runs against the minified and unminified versions
  * A fake window object is used in the test spec so that `location.pathname` can be mocked

Rules about the bar:
* Shows when cookie bar shows
* Shows when browser upgrade bar shows
* Hidden when survey shows
* Hidden when emergency publishing in use
* Hidden on register to vote and done pages
* Hidden when JS disabled
* Stops showing after it's been seen twice
* Cookie expires after 28 days

screen shot 2016-04-15 at 12 07 29
screen shot 2016-04-15 at 12 04 55
screen shot 2016-04-15 at 12 06 59
screen shot 2016-04-15 at 12 06 35


app/assets/javascripts/modules/global-bar.js
```diff
@@ -0,0 +1,57 @@
...
+      function incrementViewCount(count) {
+        count = count + 1;
+        GOVUK.setCookie(GLOBAL_BAR_SEEN_COOKIE, count, {days: 84});
+
+        if (count == 2) {
+          track('Automatically dismissed');
+        }
+      }
+
+      function viewCount() {
```
tvararu: `retrieve/get/readViewCount` are ever so slightly more descriptive, but don't bother updating unless you also think it's worth to.

tvararu:
Looks good to me, found a missing semicolon I didn't bother to mention because I think there isn't anything else we *need* to do before this goes out. :+1:

PR #770, opened 2016-04-15
Don't print the global bar

fofr:
It’s content that’s not related to the thing being printed.

cc @dsingleton @boffbowsh


benilovj:
:+1:

PR #771, opened 2016-04-18
Add toggle behaviour to history in footer component

mgrassotti:
Add the full history link and the toggling behaviour to the document_footer component.
This changes are related to https://github.com/alphagov/government-frontend/pull/134

### DIFF
--- a/app/views/govuk_component/document_footer.raw.html.erb
+++ b/app/views/govuk_component/document_footer.raw.html.erb
@@ -17,7 +17,7 @@
 %>
 <div class="govuk-document-footer<%= direction_class %>">
   <h2 class="visuallyhidden">Document information</h2>
-  <div class="history-information">
+  <div class="history-information" id="history">
     <% if local_assigns.include?(:published) && published %>
       <p>Published: <span class="published definition"><%= published %></span></p>
     <% end %>
@@ -30,16 +30,20 @@
       <% end %>
     <% end %>
     <% if history.any? %>
-      <div class="change-notes" id="history">
-        <h3>Page history</h3>
-        <ol>
-          <% history.each do |change| %>
-            <li>
-              <time datetime="<%= change[:timestamp] %>" class="timestamp"><%= change[:display_time] %></time>
-              <%= change[:note] %>
-            </li>
-          <% end %>
-        </ol>
+      <div class="change-notes" data-module="toggle">
+        <p>
+          <a href="#full-history" data-controls="full-history" data-expanded="false">+ full page history</a>
+        </p>
+        <div class="js-hidden" id="full-history"><span class="arrow"></span>
+          <ol>
+            <% history.each do |change| %>
+              <li>
+                <time datetime="<%= change[:timestamp] %>" class="timestamp"><%= change[:display_time] %></time>
+                <%= change[:note] %>
+              </li>
+            <% end %>
+          </ol>
+        </div>
       </div>
     <% end %>
   </div>


fofr:
I'm not sure what's causing the test failures here.

Could the PR name be updated to something more descriptive? ie "Adding toggle behaviour to history in footer component"?

app/views/govuk_component/document_footer.raw.html.erb
```diff
@@ -30,16 +30,20 @@
...
-          <% history.each do |change| %>
-            <li>
-              <time datetime="<%= change[:timestamp] %>" class="timestamp"><%= change[:display_time] %></time>
-              <%= change[:note] %>
-            </li>
-          <% end %>
-        </ol>
+      <div class="change-notes" data-module="toggle">
+        <p>
+          <a href="#full-history" data-controls="full-history" data-expanded="false">+ full page history</a>
```
fofr: I'm not sure we want this link to show all the time. cc @dsingleton

app/views/govuk_component/document_footer.raw.html.erb
```diff
@@ -17,7 +17,7 @@
 %>
 <div class="govuk-document-footer<%= direction_class %>">
   <h2 class="visuallyhidden">Document information</h2>
-  <div class="history-information">
+  <div class="history history-information" id="history">
```
fofr: Do we need the extra `history` class?

mgrassotti: @fofr no, probably not, I don't think this class is used in government-frontend. I will remove it. Thanks

dsingleton:
> I'm not sure what's causing the test failures here.

There's something up with builds for me to, i think there's an issue with the CI instance these builds are running on. It's being looked into.

dsingleton:
> There's something up with builds for me to, i think there's an issue with the CI instance these builds are running on. It's being looked into.

The problems with CI are resolved, and i'm re-running the build for this branch

app/views/govuk_component/document_footer.raw.html.erb
dsingleton: We could hide it when there is no JS, or adapt the copy so it makes a bit more sense if it's always visible.

I know it's the copy whitehall uses, but this is a good point to re-evaluate it, see it we can do something better. Eg, I suspect this doesn't sound great in a screen reader - but it would be good to test.

app/views/govuk_component/document_footer.raw.html.erb
```diff
@@ -31,9 +31,14 @@
...
       <div class="change-notes" data-module="toggle">
-        <p>
+        <noscript>
+          <%# The following p appears only if javascript is disabled %> 
+          <p style="margin-bottom: 5px;">Full page history:</p>
+        </noscript>
+        <p id="full-history-link" style="display: none;">
           <a href="#full-history" data-controls="full-history" data-expanded="false">+ full page history</a>
         </p>
+        <script>$("#full-history-link").show();</script>
```
mgrassotti: @fofr @dsingleton, I tried to render the behaviour we want, but if you have better ideas on where I should include the js which shows the full-history-link and how to modify the "Full page history:" title style, just let me know. I tried to see it on a reader, but this whole section seems to be hidden...

app/views/govuk_component/document_footer.raw.html.erb
tvararu: Discussed offline with @fofr: we've decided to keep the same copy and behaviour from Whitehall, for now. We can amend it easily if need be.

app/views/govuk_component/document_footer.raw.html.erb
```diff
@@ -17,7 +17,7 @@
 %>
 <div class="govuk-document-footer<%= direction_class %>">
   <h2 class="visuallyhidden">Document information</h2>
-  <div class="history-information">
+  <div class="history-information" id="history">
```
fofr: Do we use this ID? It was there previously `<div class="change-notes" id="history">`, but I'm wondering if we can remove.

mgrassotti: The "see all updates" link in the metadata points to this anchor.

image

fofr: Ok :+1:

app/views/govuk_component/document_footer.raw.html.erb
```diff
@@ -30,16 +30,20 @@
...
-              <time datetime="<%= change[:timestamp] %>" class="timestamp"><%= change[:display_time] %></time>
-              <%= change[:note] %>
-            </li>
-          <% end %>
-        </ol>
+      <div class="change-notes" data-module="toggle">
+        <p>
+          <a href="#full-history" data-controls="full-history" data-expanded="false">+ full page history</a>
+        </p>
+        <div class="overlay js-hidden" id="full-history"><span class="arrow"></span>
```
fofr: We don't need the class `overlay` here.

mgrassotti: 👍 done

PR #772, opened 2016-04-18
Add register-to-vote promo on transaction done pages

dsingleton:
**Please don't merge, but give a 👍  so I can merge when ready**

Only on completed transaction format (though we'll extend it to transaction
pages in the near future). Exclude register to vote done page, as there is
not a lot of point promoing something you've just done.

As some done pages will have register-to-vote links in their related links
we need to exclude those, and if they're the only link in a group, the
group as well.

Filtering the register-to-vote link out of the related_artefacts before
grouping makes this much easier, but should only be done for pages where
we're showing the promo.

Note: register-to-vote is filtered out by content id, as the related
artefacts don't have path/slug field, and the full web_url is pretty
long, and can vary per environment, so use the content_id, which should
be stable.

Note: the changes to the tests are not super elegant, but this change is
likely to be reverted wholesale once the promo is complete, this isn't a
pattern we're planning to keep around.

image

### DIFF
--- a/app/views/root/related.raw.html.erb
+++ b/app/views/root/related.raw.html.erb
@@ -1,10 +1,34 @@
 <%# Whilst they are both in use, this and the test copy in slimmer should be kept in sync %>
-<% if artefact and (artefact.related_artefacts.any? or (artefact.related_external_links and artefact.related_external_links.any?)) %>
+
+<%
+  promo_target_formats = ['completed_transaction', 'transaction']
+  promo_excluded_regex = /register-to-vote/
+
+  has_promo_callout = (artefact and promo_target_formats.include?(artefact.format) and not promo_excluded_regex.match(URI.decode(artefact.slug)))
+%>
+<%# Whilst they are both in use, this and the test copy in slimmer should be kept in sync %>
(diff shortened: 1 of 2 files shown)


issyl0:
Just checked with @dsingleton how this differs from https://trello.com/c/dXgKpUBZ/340-register-to-vote-promotion-on-done-pages-3, which is in progress (see the ticket comments). This is short-lived and only going to be around for a few weeks, and this is applied to every done page and content designers can't choose. The changes in our story to make it be able to be chosen are still valid, but not as urgent, so we'll carry on with them at normal speed.

fofr:
Front-end bits :+1:

app/views/root/related.raw.html.erb
```diff
@@ -1,17 +1,44 @@
 <%# Whilst they are both in use, this and the test copy in slimmer should be kept in sync %>
-<% if artefact and (artefact.related_artefacts.any? or (artefact.related_external_links and artefact.related_external_links.any?)) %>
+
+<%
+  @promo_target_formats = ['completed_transaction']
```
boffbowsh: These shouldn't be instance variables if they're local to the view.

dsingleton:
@boffbowsh Updated. Otherwise are you happy with this?

boffbowsh:
Back end bits 👍

PR #773, opened 2016-04-20
[DISCUSS] Create description component

fofr:
Base styles on uses in government-frontend:
* https://github.com/alphagov/government-frontend/blob/master/app/assets/stylesheets/helpers/_description.scss
* https://github.com/alphagov/government-frontend/blob/b9f350d6a5da7d3f9206b80c323a55203c36e883/app/views/shared/_description.html.erb

Also creates a `right-to-left` mixin.

Example of switching to this component in government-frontend:
https://github.com/alphagov/government-frontend/compare/use-description-component

description - gov uk component guide 20160420

### DIFF
--- a/app/assets/stylesheets/govuk-component/_component.scss
+++ b/app/assets/stylesheets/govuk-component/_component.scss
@@ -19,3 +19,4 @@
 @import "breadcrumbs";
 @import "related-items";
 @import "organisation-logo";
+@import "description";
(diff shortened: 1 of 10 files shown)


dsingleton:
Generally, 👍 

My only concern confusion over it accepting input to be escaped, or raw HTML. Looking at the code and (some) examples is clearer, but not from the docs and it's not particularly intuitive if you're just using it quickly/haven't used components much.

All of which increases the risk of an XSS issue if someone naively passes in unescaped input.

This isn't a problem specific to this component, just more obvious in a new one and I don't have a good solution right now. Perhaps taking `description` and `escaped_description` parameters, and adding logic to use one of those? What do you think?

fofr:
What about `description_html`?

PR #775, opened 2016-04-22
Improve EU referendum homepage link copy for SRs.

tvararu:
Screen reader users have the option to navigate by jumping between anchor tags and reading out the content. "More information" does not provide a lot of context about what the link does.

There is another equivalent PR open on `frontend`: https://github.com/alphagov/frontend/pull/945

### After

screen shot 2016-04-22 at 15 22 26

### DIFF
--- a/app/views/root/_base.html.erb
+++ b/app/views/root/_base.html.erb
@@ -38,7 +38,7 @@
     <!--[if gt IE 7]><!-->
     <div id="global-bar" data-module="global-bar" class="dont-print">
       <div class="global-bar-message-container">
-        <p class="global-bar-message"><strong>EU Referendum</strong> On Thursday 23 June there will be a vote on the UK’s membership of the European Union. <a href="https://www.eureferendum.gov.uk/" rel="external noreferrer">More&nbsp;information</a></p>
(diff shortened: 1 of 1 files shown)


fofr:
What does the external link tracking event look like now? The event label might have changed?

tvararu:
I don't have enough experience using GA to know off the top of my head if it uses the anchor tag content or the `href` to group the events. Good question though, who might be in a better position to answer?

fofr:
If you right click the link to open in new tab, then check the original tabs network requests, you should see a post to `collect`. In the `payload` you'll see the request parameters, specifically we want to check what this changes to: `&el=More%C2%A0information`

tvararu:
That is indeed changed:

```
// snip
&ec=External%20Link%20Clicked
&ea=https%3A%2F%2Fwww.eureferendum.gov.uk%2F
&el=More%C2%A0information%20about%20the%20EU%20referendum
// snip
```

So both the address and the label is passed through on the request payload, and the label has changed.

fofr:
I think that's fine, it'll be a new event type but we should be able to combine them for numbers.

PR #777, opened 2016-04-25
Add political-status and publishing-government to analytics

fofr:
* Derive political status based on current-ness of government and political state of content
* Use government slug for publishing-government
* Only include analytics tags if both government and political keys are present

Based on political and government representations in content items, added in: https://github.com/alphagov/govuk-content-schemas/pull/290

And the existing logic in Whitehall:
https://github.com/alphagov/whitehall/blob/bf784b05c467d3ce576fc8729810d21078cbc69c/app/helpers/document_helper.rb#L272-L288
https://github.com/alphagov/whitehall/blob/bf784b05c467d3ce576fc8729810d21078cbc69c/test/functional/publications_controller_test.rb#L668-L691


tvararu:
👍

PR #779, opened 2016-04-28
Create new survey test - decrease frequency that survey shows for 48 hours

fofr:
* Remove old test, add new test, fix pending spec.
* Test will run from May 4 to May 5.

https://trello.com/c/YN1CU7jB/385-gov-uk-survey-test-for-may-small

### DIFF
--- a/app/assets/javascripts/user-satisfaction-survey.js
+++ b/app/assets/javascripts/user-satisfaction-survey.js
@@ -14,14 +14,6 @@
               '  </div>' +
               '</section>',
 
-    TEST_TEMPLATE: '<section id="user-satisfaction-survey" class="visible" aria-hidden="false">' +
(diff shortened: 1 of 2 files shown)


fofr:
Looks like Jenkins is having problems.

dsingleton:
> Looks like Jenkins is having problems.

Same issue as we saw the other day, one of the jenkins machines is unavailable and the job is running on a different one that is failing. Being looked at by @alexmuller at the moment.

dsingleton:
Changes look good to me, be happy to merge once the CI issues are resolved.

fofr:
Fixed tests: `surveyFrequency() { }` is fine when not minified, but causes a parse error when minified because it should be `surveyFrequency: function() {}`

PR #783, opened 2016-05-06
Check if links to Youtube should be embedded

gemmaleigh:
When using the [govspeak component](http://govuk-component-guide.herokuapp.com/components/govspeak), all links to YouTube are converted to an [embedded video player](http://govuk-component-guide.herokuapp.com/components/govspeak/fixtures/with_youtube_embed/preview).

This PR allows links to YouTube which aren't converted to an embedded player.

Check for a data attribute `data-youtube-player` if its value is `off`
then don’t convert that link into an embedded player.

For example:
```
<a href="http://www.youtube.com/youtubevidid" data-youtube-player="off">
Watch this related video
</a>
```

### DIFF
--- a/app/assets/javascripts/govuk-component/govspeak-enhance-youtube-video-links.js
+++ b/app/assets/javascripts/govuk-component/govspeak-enhance-youtube-video-links.js
@@ -1,52 +1,66 @@
 //= require libs/jquery/jquery-ui-1.10.2.custom
 //= require vendor/jquery/jquery.player.min
 
-(function ($) {
-  window.GOVUK = window.GOVUK || {};
+(function(Modules) {
+  "use strict";
 
-  function parseYoutubeVideoId(string){
-    if(string.indexOf('youtube.com') > -1){
-      var i, _i, part, parts, params = {};
-      parts = string.split('?');
-      if (parts.length === 1){
-        return;
+  Modules.YoutubeVideoLinks = function() {
+
+    this.start = function($el) {
+
+      function parseYoutubeVideoId(string){
+        if(string.indexOf('youtube.com') > -1){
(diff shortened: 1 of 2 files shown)


gemmaleigh:
cc. @AndrewVos @thehenster @fofr

fofr:
Why has this been made a module? It's not behaving like a module – ie auto-instantiated using a data attribure which leads to the start method being called. It seems to instead start itself?

If it did follow the patterns of a module I think it wouldn't be backwards compatible.

gemmaleigh:
I was trying to get it to work in a similar way to the option select component, which doesn't use a [data attribute to instantiate the module](https://github.com/alphagov/static/blob/master/app/assets/javascripts/govuk-component/option-select.js).

I understand this won't work as-is, without the data attribute, but the original version works in exactly the same way - [it starts itself](https://github.com/alphagov/static/blob/master/app/assets/javascripts/govuk-component/govspeak-enhance-youtube-video-links.js#L50).

gemmaleigh:
Like this?

```
(function ($) {
  "use strict";
  window.GOVUK = window.GOVUK || {};

  function parseYoutubeVideoId(string){
  ...
  }

  function shouldConvertToEmbeddedPlayer($link){
    ...
  }

  function enhanceYoutubeVideoLinks($el) {
  ...
  }

  GOVUK.enhanceYoutubeVideoLinks = enhanceYoutubeVideoLinks;

  var $element = $('.govuk-govspeak');
  GOVUK.enhanceYoutubeVideoLinks($element);

})(jQuery);
```

fofr:
> Is is okay to use the GOV.UK namespace, but not the module pattern in this context?
Yes.

gemmaleigh:
Closing in favour of #788.

PR #785, opened 2016-05-09
Fix display of external links in the metadata component.

tvararu:
Because the base font-size is different on this component, the external
link icon looks a bit off.

### Before

screen shot 2016-05-09 at 16 02 38

### After

screen shot 2016-05-09 at 16 03 40


fofr:
Could you add an external link example to here:
https://github.com/alphagov/static/blob/master/app/views/govuk_component/docs/metadata.yml

fofr:
Otherwise :+1:

PR #786, opened 2016-05-09
Create a test for video embedding within the govspeak component

gemmaleigh:
I've created a minimal test for the [govspeak component with YouTube video embed](http://govuk-component-guide.herokuapp.com/components/govspeak/fixtures/with_youtube_embed/preview). 

I'm getting the following error: 

```
  TypeError: 'null' is not an object (evaluating 'media.container') 
```

@fofr I saw your PR over on the repo for the [fork of the accessible media player](https://github.com/alphagov/Accessible-Media-Player/commit/94107068ed6cef422e92890d52e87b3133a71563). 

I think these issues might be related, as the YouTube player is also missing a file extension (it shouldn't be required).

### DIFF
--- a/app/assets/javascripts/govuk-component/govspeak-enhance-youtube-video-links.js
+++ b/app/assets/javascripts/govuk-component/govspeak-enhance-youtube-video-links.js
@@ -24,6 +24,14 @@
     }
   }
 
+  function checkPrototcolScheme(){
+    var scheme = document.location.protocol;
+    if (scheme == "file:") {
+      scheme = "https:";
+      return scheme;
+    }
+  }
+
   function enhanceYoutubeVideoLinks($el){
     $el.find("a[href*='youtube.com'], a[href*='youtu.be']").each(function(i){
       var $link = $(this),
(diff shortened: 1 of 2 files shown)


thehenster:
I've made this work by specifying the https version of the youtube API. I'll paste why from the commit message below.

Phantom's window object has a default location of `file:///` which has a protocol of `file:`.

In the AccessibleMediaPlayer the following snippet determine's which type of player to load.

```
if (config.url.match(/^(http|https)\:\/\/www\.youtube\.com/)) {
  // a youtube-y player
} else {
  // a non youtube-y player
}
```

So if we get to that point in a jasmine/phantomjs test and our protocol is `file:` rather than `http:` or `https:` it's going to attempt to load a non-youtube-y player reagardless.

This could be a changed a little to either call http:// or https:// but never file:// ?

thehenster:
Hmm. Or maybe it'd be better to write a function to calculate the protocol which is mockable in tests and then we can leave the behaviour the same in production? There's probably no meaningful difference since we won't ever have a protocol of file: in production..

fofr:
Nice sleuthing! :+1:

gemmaleigh:
Yup, good job! @thehenster. 

Closing in favour of #788.

PR #789, opened 2016-05-13
[Do not merge] Remove global banner bar

fofr:
__Deploy on May 25__
https://trello.com/c/npd5H4IW/405-sitewide-banner-remove-supporting-banner-code

* Delete inline class toggle JS and spec
* Delete global bar markup and module for dismissing/storing view count
* Remove styles

We don’t intend to use this again soon.

Reverts #769


gpeng:
:+1:

PR #790, opened 2016-05-13
Dedupe organisations in meta tags

tijmenb:
This element assumes that the `organisations`, `lead_organisations` and `supporting_organisations` links in the links hash never overlap. This is not the case, and definitely isn't since https://github.com/alphagov/govuk-content-schemas/pull/308. This commit dedupes them before generating the metatag.


fofr:
Should this include a test? eg https://github.com/alphagov/static/blob/master/test/govuk_component/analytics_meta_tags_test.rb#L36-L48

tijmenb:
I didn't include that because on my machine it's tested with the new govuk-content-schemas branch. https://github.com/alphagov/govuk-content-schemas/pull/308 fails without this change.

fofr:
:+1:

PR #791, opened 2016-05-17
Add missing print-base helpers to guides-print.scss

tvararu:
This will bump the text size and make printed text more legible.

Will affect pages such as https://www.gov.uk/after-a-death/print.

### DIFF
--- a/app/assets/stylesheets/guides-print.scss
+++ b/app/assets/stylesheets/guides-print.scss
@@ -17,6 +17,7 @@ $is-print: true;
 // OUTPUT
 // The following files and inline CSS will form the output of print.css
 @import "helpers/core-print";
+@import "helpers/print-base";
 
 section > header h1:after {
   content: "";


nickcolley:
This looks good to me, only thing I'd think about is that since this is _core_print it could be a dependency of another print stylesheet somewhere else, potentially that other dependant might rely on this smaller font-size so could be a breaking change?

tvararu:
The only `@import` call sites in this codebase appear to be in [guides-print.scss](https://github.com/alphagov/static/blob/master/app/assets/stylesheets/guides-print.scss#L19) and [static-print.scss](https://github.com/alphagov/static/blob/master/app/assets/stylesheets/static-print.scss#L6). The former is a special case with some overrides for guides. The latter is the generic case, which does affect all print stylesheets. Neither of those rely on the `font-size` or override it explicitly.

But since static is imported through loose contracts in a number of other codebases then yes, this change is potentially far-reaching.

@dsingleton?

dsingleton:
`static-print.css` is the print CSS file for the `wrapper` layout (called `static` internally 🙁) - which is used by most [mainstream/older frontend applications](https://github.com/search?utf8=%E2%9C%93&q=%22+slimmer_template+%27wrapper%27%22&type=Code&ref=searchresults). 

So this change wouldn't apply to content rendered with the [`header_and_footer_only` layout](https://github.com/search?utf8=%E2%9C%93&q=%22+slimmer_template+%27header_footer_only%27%22&type=Code&ref=searchresults) or `core_layout` (the default), but would affect all of the rendering apps using `wrapper`, not just guides.

Looking at some non-guide formats, removing the `font-size: 62.5%` rule from static would have no effect, as the same value is set in `govuk_template`'s [print stylesheet ](https://github.com/alphagov/govuk_template/blob/86106bfb332ff5c0346b92310a4117daeb2d8c8d/source/assets/stylesheets/govuk-template-print.scss).

It looks like the bug is with the only when something is [rendered with the static `print` layout](https://github.com/alphagov/frontend/blob/master/app/controllers/root_controller.rb#L196), which is `guide`s and `programme`s (programme is rendered in the same way as guide, [example](https://www-origin.integration.publishing.service.gov.uk/1619-bursary-fund/further-information), and that the bug fix should go in the CSS for that layout, rather than in the general purpose layouts?

tvararu:
> Looking at some non-guide formats, removing the `font-size: 62.5%` rule from static would have no effect, as the same value is set in `govuk_template`'s [print stylesheet ](https://github.com/alphagov/govuk_template/blob/86106bfb332ff5c0346b92310a4117daeb2d8c8d/source/assets/stylesheets/govuk-template-print.scss).

I wasn't aware of that, thanks! I asked @alext if there is any historical context why it was added but it appears that it was also ported over from somewhere else, most likely from back in the early days of the beta.

If I understand correctly, what you're proposing is no longer doing this in `_core_print.scss` and rather adding an override only for `guides` through `guides-print.scss`, perhaps like this:

```scss
html {
  font-size: inherit;
}
```

That wouldn't affect other formats.

But to the best of my understanding this is a bug with all formats, so we shouldn't scope the fix to just guides. So, @dsingleton, which way do you recommend?

dsingleton:
(Discussed offline with @nickcolley and @tvararu)

Adding ```@import "helpers/print-base";``` to https://github.com/alphagov/static/blob/print-stylesheet-a11y/app/assets/stylesheets/guides-print.scss#L20 will fix this reliably. 

The underlying problem is the `guides-print` SCSS, which is used by [`print` layout](https://github.com/alphagov/static/blob/master/app/views/root/print.html.erb#L7) was missing the [global print reset for elements](https://github.com/alphagov/static/blob/master/app/assets/stylesheets/helpers/_print-base.scss).

tvararu:
I've updated this PR. ✨

nickcolley:
Thanks @tvararu 👍

dsingleton:
👏

PR #792, opened 2016-05-17
Improve .report-a-problem-toggle:focus contrast

tvararu:
This link seems to override the base link classes, but produces a combination that has a poor contrast ratio when focused.

This commit updates the focus styles to fix the contrast.

### Before

screen shot 2016-05-17 at 14 44 51

### After

screen shot 2016-05-17 at 14 44 55


fofr:
Nice 👍

PR #794, opened 2016-05-19
Add HMRC webchat to ‘manage your tax credits’ page

daibach:
Request from HMRC

To be merged and deployed on Friday 20 May 2016.

### DIFF
--- a/app/assets/javascripts/webchat.js.erb
+++ b/app/assets/javascripts/webchat.js.erb
@@ -25,7 +25,7 @@
         '/vat-registration/changes-to-your-details',
         '/government/organisations/hm-revenue-customs/contact/national-insurance-numbers',
         '/lost-national-insurance-number',
-        '/manage-your-tax-credits',
+        '/manage-your-tax-credits'
       ];
 
       return ($.inArray(window.location.pathname, webchatEnabledPaths) >= 0);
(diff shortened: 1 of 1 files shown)


boffbowsh:
👍

issyl0:
This looks like what we did the last time we had to add it to a page, so :+1:.

app/assets/javascripts/webchat.js.erb
```diff
@@ -25,6 +25,7 @@
         '/vat-registration/changes-to-your-details',
         '/government/organisations/hm-revenue-customs/contact/national-insurance-numbers',
         '/lost-national-insurance-number',
+        '/manage-your-tax-credits',
```
robinwhittleton: Trailing comma will break in IE7+8. Can it be removed?

issyl0: We tend to keep trailing commas around to reduce noisy diffs when the next change comes (otherwise you'd have two lines changed because of adding a comma to the previous line) but I suppose so, if it breaks something...

fofr: JavaScript is the exception to this rule. Trailing commas break older IEs and are hard to track down.

boffbowsh:
FTFY

PR #796, opened 2016-05-25
Remove invalid breadcrumbs aria role from base template

fofr:
* Breadcrumbs is not a valid ARIA role
* It is not being used for styling

We previously removed from the component: https://github.com/alphagov/static/pull/737
Was added in: https://github.com/alphagov/static/pull/238


gemmaleigh:
👍  https://www.w3.org/TR/wai-aria/roles

tombye:
I think this change was also on the accessibility hack day list too so 👍

nickcolley:
Thanks @fofr 👍  🎉

PR #797, opened 2016-05-26
Remove stylesheets duplicated in header-footer-only

fofr:
`styleguide/conditionals2`, `helpers/footer` and `helpers/header` are
all already included in `header-footer-only`. This means their styles are duplicated in the generated CSS.

It seems odd that header and footer weren't removed when switching to `header-footer-only`, but I haven't found any explicit reason for them to be kept. Nothing appears to override those styles.

For an example, search for `#user-satisfaction` in https://assets.digital.cabinet-office.gov.uk/static/static-35b681259fdcd07f543889e5711c8dbcf7728a3fa318c4881d0ba2e3d8bb3ac0.css

### DIFF
--- a/app/assets/stylesheets/static.scss
+++ b/app/assets/stylesheets/static.scss
@@ -1,17 +1,12 @@
-/* govuk_frontend_toolkit includes */
 @import "header-footer-only";
 
+/* govuk_frontend_toolkit includes */
 @import "shims";
 @import "design-patterns/buttons";
-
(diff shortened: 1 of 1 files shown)


nickcolley:
This looks good to me, it is possible that the order in which the duplicates were imported could break something.

For example if one of the imports before the duplicate import is being overridden...

As far as I can tell that's not an issue. 👍 

cc @dsingleton - can you confirm the above? :)

PR #798, opened 2016-05-26
Turn on HMRC webchat for more tax credit URLS

daibach:
At request of HMRC.

Also re-orders the list of URLs to make it easier to find which URLs are enabled and removes a duplicate URL.

Probably best to review by commit.


robinwhittleton:
I wonder why the ‘a’ from the title is missing from `/tax-credits-if-you-have-baby`? Either way, looks good.

PR #800, opened 2016-05-26
Track when HMRC webchat is offered to users in analytics

daibach:
We currently track when users interact with HMRC’s webchat service (eg when they click the ‘accept’ or ‘reject’ buttons). This commit adds in an event to track when the webchat service is offered to users.

This event, alongside other data such as page views, will give us an idea of:

1. how many people are being presented with webchat but aren’t doing
anything with it
2. how many people are not being presented with webchat at all

cc @robinwhittleton @dsingleton @davidillsley


daibach:
Thanks @fofr!

PR #801, opened 2016-05-27
Bump template and toolkit

fofr:
* Updates the `footer-categories` class to work with new grid by extending `.grid-row` and removing side padding. This is now like-for-like with live.

# Update govuk_frontend_toolkit to 4.12.0
https://github.com/alphagov/govuk_frontend_toolkit/blob/master/CHANGELOG.md

## 4.12.0
Increase button padding to match padding from GOV.UK elements. If you have UI which depends on the padding set by the button mixin in the frontend toolkit and this is not overridden by button padding set by GOV.UK elements, this change will affect it.

## 4.11.0
Remove the GDS-Logo font-face definition
Move the @viewport statements to govuk_template. If you upgrade to this version of govuk_frontend_toolkit and you’re also using govuk_template you’ll need to upgrade that to at least 0.17.2 to maintain compatibility with desktop IE10 in snap mode.

# Update govuk_template to 0.17.3
https://github.com/alphagov/govuk_template/blob/master/CHANGELOG.md

## 0.17.3
Fix bug with heading colours in footer

## 0.17.2
Fix a bug with the skip-to-content link and iOS Voiceover
Migrate @viewport statement from govuk_frontend_toolkit

## 0.17.1
Reduce file size of template: removes HTML comments, type attributes on scripts, and uses HTML5 charset declaration.
Switch external link media query to be mobile first
Sass file cleanups
Replace old grid mixins with newer grid from frontend toolkit
Remove duplicate grey variables


fofr:
This still has the peculiar double meta charset issue:
```html
<meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
<meta charset="utf-8">
```

This doesn't break anything, and I think it's worth making the updates to get the latest changes in. This should be investigated separately, and probably involves some obscure Rails magic.

nickcolley:
I've added a commit based on a discussion offline with @dsingleton that describes what we need to refactor in addition to this bump later on.

PR #804, opened 2016-05-31
Remove `lead_organisations` & `supporting_organisations`

tijmenb:
These have been deprecated. We've ensured that all `lead_organisations` and `supporting_organisations` have been merged into `organisations` (https://github.com/alphagov/whitehall/pull/2585), so the data will be the same.


MatMoore:
The overall change makes sense to me but I don't know how the analytics work (eg are those tags arbitrary or does something need cleaning up elsewhere?).

👍  to merge if you are happy with it.

tijmenb:
This partial generates a meta tag, which is used by analytics. Nothing more to change. Happy for this to get merged.

Davidslv:
👍

PR #806, opened 2016-07-06
[Do not merge] Move unstuck logic to CSS so that breakpoints are better respected

nickcolley:
Since breakpoints in the grid are determined in CSS having
the logic that forces an element to be unstuck will mean
it can track the same variable that the grid system is using.


tvararu:
👍  looks good to me!

PR #807, opened 2016-07-07
[Do not merge] Disable webchat on the tax credits page

tvararu:
This page will be the first to transition to a new implementation that we want to test with HMRC.

This shouldn't be merged/deployed before the new implementation on alphagov/contacts-frontend is also ready to go: https://github.com/alphagov/contacts-frontend/pull/50


daibach:
This isn't going to be the first page for implementation anymore, although it'll come later in the transition.

Going to close this pull request for now. We can re-open it when this page is ready to transition.

Issue #808, opened 2016-07-15
Search box layout breaks when increasing font size


selfthinker:
WCAG2 1.4.4 says that to hit double-AA [text must be able to zoom to 200% of the default size without breaking the site layout](https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-scale.html). Currently the search box is fixed to 35px which makes the overlaid label unreadable when resized:

![screen shot 2016-07-15 at 14 56 21](https://cloud.githubusercontent.com/assets/108893/16876801/559c44c2-4a9d-11e6-9899-13918f9e258c.png)

We should allow the input to grow in size to contain the text, best removing the `height` completely.

(Sorry @robinwhittleton for copying most of your copy from alphagov/govuk_template#232.)

selfthinker:
This was fixed in #809 and got released today.

PR #809, opened 2016-07-18
Fix search box breaking when increasing font size

selfthinker:
Bug report: #808

This fixes the search box breaking when someone increases the font size. When something has a height (especially of px) it doesn't grow with the text (not when zooming but increasing just the font size).

Removing the height completely would be the ideal solution but that caused issues in IE9, IE10 and WindowsPhone due to the alignment with the submit button.
The `height` got substituted with `line-height` but that alone caused issues in Safari in many iOS versions and Chrome on some Android phones, the text box somehow had a height of 36px and not 35px which caused a white line to appear below the submit button.
That's why the height of all the things got increased to 36px which fixed it. (Although it is not clear what caused it.)

### DIFF
--- a/app/assets/stylesheets/helpers/_header.scss
+++ b/app/assets/stylesheets/helpers/_header.scss
@@ -63,7 +63,7 @@
     }
     label {
       float: left;
-      height: 35px;
+      min-height: 36px;
       line-height: 35px;
       text-indent: 15px;
       overflow: hidden;
@@ -88,10 +88,13 @@
       display: block;
       margin: 0;
       border: 0;
-      height: 35px;
+      min-height: 36px;
       padding: 6px 0;
-      @include ie-lte(7){
-        height: 23px;
+      @include ie-lte(8) {
+        min-height: 24px;
+      }
+      @include ie-lte(6) {
+        height: 24px;
       }
       @include appearance(none);
     }
@@ -122,8 +125,8 @@
       z-index: 4;
       right: 0;
       top: 0;
-      width: 35px;
-      height: 35px;
+      width: 36px;
+      height: 36px;
 
       border: 1px solid $mainstream-brand;
       border-width: 0 0 0 1px;


selfthinker:
Hmm, I don't have the right permissions to see details of the Jenkins errors. Who do I need to ask for that?

Davidslv:
@selfthinker Seems there was some problem connecting to github before, I've rebuilt it for you.

robinwhittleton:
This changes the default minimum size. Is that intentional?

jul-20-2016 15-30-25

selfthinker:
It's half-intentional. As I wrote above, when I changed the original `height: 35px` to `min-height: 35px`, several browsers would make the input's height 36px. To make it more consistent across all browsers, I increased it everywhere to 36px. There might be a better way to fix that (e.g. reducing the padding?), so I'd be happy with other solutions.

robinwhittleton:
@markhurrell thoughts?

markhurrell:
looks better at 36px - think it was meant to be that anyway, it's not vertically centered at 35px

good spot fixing the issue

app/assets/stylesheets/helpers/_header.scss
```diff
@@ -88,10 +88,13 @@
       display: block;
       margin: 0;
       border: 0;
-      height: 35px;
+      min-height: 36px;
       padding: 6px 0;
-      @include ie-lte(7){
-        height: 23px;
+      @include ie-lte(8) {
```
robinwhittleton: Previously IE8 was getting a 35px `height`, but now we’re setting a 24px `min-height`. Does its handling of `height` and `min-height` differ on inputs?

selfthinker: I changed it just because the dimensions were broken otherwise. But now I've dug a bit deeper into it. IE8 seems to have a problem with `box-sizing` together with `min-height`/`min-width` (and I guess `max-height`/`max-width` as well, but I haven't tested that). Unfortunately caniuse.com doesn't mention that, might be worth reporting. I found [evidence on SO](http://stackoverflow.com/questions/9508262/min-height-min-width-doesnt-respect-box-sizing-in-some-browsers). And I created [this reduced test case](https://dl.dropboxusercontent.com/u/4667354/gds/min-height-ie8.html) which demonstrates the issue reproducibly (if you open it in IE8).

robinwhittleton: Good enough for me. Maybe do a PR for caniuse? Data file for `box-sizing` is https://github.com/Fyrd/caniuse/blob/master/features-json/css3-boxsizing.json

PR #810, opened 2016-07-20
Refactor surveys javascript to support small test surveys

kalleth:
The Survey & Support team (Myself and @NeilvB) are working on a backlog including 4 or 5 "test" surveys that will replace the existing "User Satisfaction Survey" that runs across GOV.UK.

The intention of this PR is to take the existing survey code -- which couples the information about the survey being run to the implementation of how it is displayed -- and separate those concerns, making it easier to run small test surveys in future.

This PR is essentially a "no-op" change, keeping the existing user satisfaction survey running.

We provide a 'default survey', which is the existing user satisfaction survey, and an array which can contain "small surveys" which run for a time bounded period, with their own frequencies, and supersede the default survey.

An example of a "small survey" that could be added would be:

```ruby
smallSurveys: [
  {
    url: 'https://www.surveymonkey.com/s/SURVEY_CODE',
    template: TEMPLATE,  //inherit from the default template with no changes
    identifier: 'three_day_survey',
    frequency: 50, //show to 1 in 50 visitors
    startTime: new Date("May 4, 2016").getTime(),
    endTime: new Date("May 5, 2016 23:59:59").getTime()
  }
],
```

- [x] Deploy to integration
- [x] Product acceptance


app/assets/javascripts/surveys.js
```diff
@@ -0,0 +1,140 @@
...
+    isSurveyToBeDisplayed: function(survey) {
+      if (userSurveys.otherNotificationVisible() ||
+          GOVUK.cookie(userSurveys.surveyTakenCookieName(survey)) === 'true') {
+        return false;
+      } else if (userSurveys.userCompletedTransaction()) {
+        // We don't want any survey appearing for users who have completed a
+        // transaction as they may complete the survey with the department's
+        // transaction in mind as opposed to the GOV.UK content.
+        return false;
+      } else if ($('#user-satisfaction-survey-container').length <= 0) {
```
elliotcm: These surveys are still tied to this specific survey's container, are there plans to make this more generic?

app/assets/javascripts/surveys.js
```diff
@@ -0,0 +1,140 @@
...
+        userSurveys.displaySurvey(activeSurvey);
+      }
+    },
+
+    getActiveSurvey: function(defaultSurvey, smallSurveys) {
+      var activeSurvey = defaultSurvey;
+
+      $.each(smallSurveys, function(_index, survey) {
+        if(userSurveys.currentTime() >= survey.startTime && userSurveys.currentTime() <= survey.endTime) {
+          activeSurvey = survey;
```
elliotcm: This means that if more than one survey is running at once, only the last one added will ever be shown to users.

elliotcm:
Other than a couple of questions this seems fine.  I know a lot of the code has just been moved, but could you check your new stuff against https://github.com/alphagov/styleguides/blob/master/js.md if you haven't already?  Much thanks.

app/assets/javascripts/surveys.js
kalleth: We identified this limitation during development; a requirement to have multiple surveys "active" at once exists but not until test #7 (We're currently on test #4) and we have a plan for how to address it when we get there.

For the moment, the desired user need is to have the test survey "supersede" the default survey.

app/assets/javascripts/surveys.js
NeilvB: @kalleth and I had a chat about this - since the content of all the upcoming surveys are related to user satisfaction, we thought that this ID name is OK as it is.

elliotcm:
Merge when ready.

PR #812, opened 2016-07-21
Fix transaction bug introduced in 8f1f835

kalleth:
In 8f1f835 when attempting to refactor the transaction detection logic, I inverted the match (returning true if inside a transaction URL, not false if *not* inside a transaction URL) but forgot to invert the if statement to "OR" from "AND".

This commit fixes that, and adds some tests to ensure the mistake can't happen again.

Credit to @knewter who spotted this when I showed him my GOV.UK PR :(.

👍 for coding in the open, though!

### DIFF
--- a/app/assets/javascripts/surveys.js
+++ b/app/assets/javascripts/surveys.js
@@ -106,15 +106,15 @@
     },
 
     userCompletedTransaction: function() {
-      var currentURL = window.location.pathname;
+      var path = userSurveys.currentPath();
 
       function stringContains(str, substr) {
         return str.indexOf(substr) > -1;
       }
 
-      if (stringContains(currentURL, "/done") &&
-          stringContains(currentURL, "/transaction-finished") &&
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/surveys.js
```diff
@@ -106,15 +106,15 @@
     },
 
     userCompletedTransaction: function() {
-      var currentURL = window.location.pathname;
+      var currentPath = userSurveys.currentUrl();
```
elliotcm: Is this a path or a URL?

elliotcm:
Small consistency fix needed, the actual code change looks good.  Sorry I missed it during review.

app/assets/javascripts/surveys.js
kalleth: `/path-to-resource`, so I'll fix the helper method to be currentPath rather than URL too. Good spot.

kalleth:
@elliotcm Updated to include the consistency fix (`currentUrl => currentPath`).

PR #814, opened 2016-07-22
Add translations for all govuk components

nickcolley:
Continues the work done in https://github.com/alphagov/static/pull/505

Moves the language translation config files back into the conventional place rather than in the view directory.

### DIFF
--- a/app/controllers/root_controller.rb
+++ b/app/controllers/root_controller.rb
@@ -24,14 +24,15 @@ class RootController < ApplicationController
   end
 
   def govuk_available_locales
-    locale_files = Rails.root.join('app', 'views', 'locales', '*.yml')
+    locale_files = Rails.root.join("config", "locales", "*.yml")
     locales = Dir[locale_files].map { |file| File.basename(file, '.yml') }
     render :json => locales
   end
 
   def govuk_locales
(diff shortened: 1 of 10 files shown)


app/controllers/root_controller.rb
```diff
@@ -24,14 +24,15 @@ def govuk_component_docs
...
 
   def govuk_available_locales
-    locale_files = Rails.root.join('app', 'views', 'locales', '*.yml')
+    locale_files = Rails.root.join("config", "locales", "*.yml")
     locales = Dir[locale_files].map { |file| File.basename(file, '.yml') }
     render :json => locales
   end
 
   def govuk_locales
     return error_404 unless params[:locale].match(/^[a-z]{2}(-[a-z0-9]{2,3})?$/)
```
Davidslv: That regular expression will allow something like `pt-p1p`, which ISO-639 are we using and which ISO is that regex trying to accommodate?

https://en.wikipedia.org/wiki/ISO_639

nickcolley: I'm not familiar with the internals of the i18n for Rails so not sure, this seems like a sanity check but maybe something like https://github.com/tigrish/iso#tags would be more appropriate.

gpeng: There is a bit of a mixture of ISO-639 standards and one or two non-standard locales. I think this is reasonable as a simple check.

PR #816, opened 2016-07-26
Add documentation on the reasoning behind components

dsingleton:
Try and explain the why behind components, as well as the how.

I'm looking for feedback on
- how understandable it is
- if it's helpful to someone new to GOV.UK Components
- phrasing
- spelling

(in that order)

### DIFF
--- a/doc/govuk-components.md
+++ b/doc/govuk-components.md
@@ -9,7 +9,7 @@ The available components and their documentation are exposed by an API at `/temp
 
 * a [Partial View](../app/views/govuk_component) - The template logic and markup, also defines the arguments expected
 * a [SCSS module](../app/assets/stylesheets/govuk-component) - The styling of the component
-* a Javascript module - no examples yet.
+* a [Javascript module](../app/assets/javascripts/govuk-component) - JS behaviour of the component
 * [Documentation](../app/views/govuk_component/docs) - a `.yml` per component, describing the component and containing fixture data.
 * a [unit test](../test/govuk_component) - testing the component renders correctly based on parameters passed in
 
--- a/doc/govuk-components.md
+++ b/doc/govuk-components.md
@@ -13,6 +13,60 @@ The available components and their documentation are exposed by an API at `/temp
 * [Documentation](../app/views/govuk_component/docs) - a `.yml` per component, describing the component and containing fixture data.
 * a [unit test](../test/govuk_component) - testing the component renders correctly based on parameters passed in
 
+## Principles behind components
+
+### Encapsulate common UI
+
+GOV.UK Components aim to hide the complexity of a piece of common UI. Putting the templating logic, HTML, CSS and JS behind an abstraction, so that an app rendering a component only needs to know about the parameters to render it with. This is similar to how [React components](https://facebook.github.io/react/index.html) work.
+
+Think of it as a function call in a programming language, as long as the [interface](https://en.wikipedia.org/wiki/Application_programming_interface) doesn't change, then code rendering a component doesn't need to care about the changes.
+
+Here's an example of calling a component. The parameter is `title` with a value of `"My page title"`.
+
+```ruby
+<%= render partial: 'govuk_component/title', locals: { title: "My page title" } %>
+```
+
+This will output something like:
+
+```html
+<div class="govuk-title length-average">
+  <h1>My page title</h1>
+</div>
+```
+
+Apps should treat the output of a component like a [black box](https://en.wikipedia.org/wiki/Black_box), not relying on the internal implementation as it may change (eg, additional CSS based on the `govuk-title` or `h1` selector)
+
+If additional behavior in the component is needed then it should be made in the component itself, then used in an application.
+
+### Distribute via the network
+
+One of the goals of GOV.UK Components is to have consistency between the many different frontend applications. Because of this components live in a central place (the `static` application) and are distributed to the different frontends over the network. This is done via Slimmer, in a similar way to [layout templates](slimmer_templates.md).
+
+This allows a component to be changed and deployed within `static` and each frontend application will automatically fetch a new version of the [component template code](../app/views/govuk_component).
+
+The automatic updating allows changes to be made to every GOV.UK frontend app, without having to manually update and deploy them. With 20+ frontend apps, this makes updating components much easier.
+
+However, this approach does have some costs. It's less explicit, and not as clear how the common component code is included or updated. Changes to the component interface/parameters require [forwards-only migrations](http://engineering.skybettingandgaming.com/2016/02/02/how-we-release-so-frequently/#forward-only-migrations), which for components, would look like this:
+
+- Release a version of static that introduces backwards compatible change to a component
+- Migrate all frontends to use new API and stop depending on old one
+- Release version of static to remove old implementation for a component
+
+For components of established UI patterns this should be less of an issue.
+
+### Avoid complexity
+
+The parameters to a component should be simple data. This is to make components easier to understand and easier to test. Try to avoid passing complex objects into a component, as this complicates the interface and makes the component harder to reuse. A good rule of thumb for keeping the template logic simple is: "can the template logic be implemented in a simpler templating language like `mustache`", ie, avoiding ERB/ruby features.
+
+A long-term goal is that a component can be extracted out of the GOV.UK Publishing specific `static` repo and moved to a repo usable by wider government, like `govuk_template` or `govuk_frontend_toolkit`.
+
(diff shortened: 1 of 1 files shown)


carolinegreen:
Seems like a good explanation to me.  It's nice as well having the comparisons to other things.  I've been around for a while but haven't used components.  Probably want someone who's actually new to look at it too though.

doc/govuk-components.md
```diff
@@ -9,10 +9,58 @@ The available components and their documentation are exposed by an API at `/temp
...
 * [Documentation](../app/views/govuk_component/docs) - a `.yml` per component, describing the component and containing fixture data.
 * a [unit test](../test/govuk_component) - testing the component renders correctly based on parameters passed in
 
+## Principles behind components
+
+### Encapsulate common UI
+
+GOV.UK Components aim to hide the complexity of a piece of common UI. Putting the templating logic, HTML, CSS and JS behind an abstraction, so that an app rendering a component only needs to know about the parameters to render it with. This is similar to how [React components](https://facebook.github.io/react/index.html) work.
+
+Think of it as a function call in a programming language, as long as the [interface](https://en.wikipedia.org/wiki/Application_programming_interface) doesn't change, then code rendering a component doesn't need to care about the changes.
```
nickcolley: Do we want components to be pure functions, for example with no side effects?

doc/govuk-components.md
```diff
@@ -9,10 +9,58 @@ The available components and their documentation are exposed by an API at `/temp
...
+
+```html
+<div class="govuk-title length-average">
+  <h1>My page title</h1>
+</div>
+```
+
+Apps should treat the output of a component like a [black box](https://en.wikipedia.org/wiki/Black_box), not relying on the internal implementation as it may change (eg, additional CSS based on the `govuk-title` or `h1` selector)
+
+If additional behavior is required in the component is needed then it should be made in the component itself, then used in an application.
```
selfthinker: Is that a typo? I guess only `is required` or `is needed` should be used?

doc/govuk-components.md
```diff
@@ -9,10 +9,58 @@ The available components and their documentation are exposed by an API at `/temp
...
+
+### Distribute via the network
+
+One of the goals of GOV.UK Components is to have consistency between the many different frontend applications. Because of this components live in a central place (the `static` application) and are distributed to the different frontends over the network. This is done via Slimmer, in a similar way to [layout templates](slimmer_templates.md).
+
+This allows a component to be changed and deployed within `static` and each frontend application will automatically fetch a new version of the [component template code]()
```
selfthinker: Link is missing, as is a period.

doc/govuk-components.md
```diff
@@ -9,10 +9,58 @@ The available components and their documentation are exposed by an API at `/temp
...
+
+The automatic updating allows changes to be made to every GOV.UK frontend app, without having to manually update and deploy them. With 20+ frontend apps, this makes updating components much easier.
+
+However, this approach does have some costs. It's less explicit, and not as clear how the common component code is included or updated. Changes to the component interface/parameters must be backwards compatible, which can require multi-step changes, but with established UI patterns this should be less of an issue.
+
+### Avoid complexity
+
+The parameters to a component should be simple data. This is to make components easier to understand and easier to test. Try to avoid passing complex objects into a component, as this complicates the interface and makes the component harder to reuse. A good rule of thumb is if the template logic can be implemented in a simpler templating language like `mustache`, avoiding ERB/ruby features.
```
selfthinker: "A good rule of thumb [...]" is not a complete sentence.

doc/govuk-components.md
```diff
@@ -9,10 +9,58 @@ The available components and their documentation are exposed by an API at `/temp
...
+
+A long-term goal is that a component can be extracted out of the GOV.UK Publishing specific `static` repo and moved to a repo usable by wider government, like `govuk_template` or `govuk_frontend_toolkit`.
+
+There are some cases where components can expect a complex object as a parameter, so that logic is not duplicated in each application calling the component, for example the [analytics component](https://github.com/alphagov/static/blob/master/app/views/govuk_component/analytics_meta_tags.raw.html.erb#L2-L33) has complex logic to extract the analytics information from a content item, which can be updated without updating each application.
+
+The component is split into two parts, the complex logic can be thought of as a Presenter and the [simpler part](https://github.com/alphagov/static/blob/master/app/views/govuk_component/analytics_meta_tags.raw.html.erb#L36-L38) as the traditional template (ie, could be extracted out).
```
selfthinker: I don't know what a Presenter is. Maybe a link would be helpful.

selfthinker:
I would say most of this is very understandable and helpful. Apart from my comments, the other thing which could be improved is making some sentences shorter (or splitting them into two or three). But that's not super important.

dsingleton:
@selfthinker updated the bits around your comments, thanks!

doc/govuk-components.md
```diff
@@ -9,10 +9,58 @@ The available components and their documentation are exposed by an API at `/temp
...
+
+Apps should treat the output of a component like a [black box](https://en.wikipedia.org/wiki/Black_box), not relying on the internal implementation as it may change (eg, additional CSS based on the `govuk-title` or `h1` selector)
+
+If additional behavior in the component is needed then it should be made in the component itself, then used in an application.
+
+### Distribute via the network
+
+One of the goals of GOV.UK Components is to have consistency between the many different frontend applications. Because of this components live in a central place (the `static` application) and are distributed to the different frontends over the network. This is done via Slimmer, in a similar way to [layout templates](slimmer_templates.md).
+
+This allows a component to be changed and deployed within `static` and each frontend application will automatically fetch a new version of the [component template code]((../app/views/govuk_component).
```
selfthinker: This is not a link due to a syntax error (two opening brackets).

doc/govuk-components.md
```diff
@@ -9,10 +9,58 @@ The available components and their documentation are exposed by an API at `/temp
...
+
+The automatic updating allows changes to be made to every GOV.UK frontend app, without having to manually update and deploy them. With 20+ frontend apps, this makes updating components much easier.
+
+However, this approach does have some costs. It's less explicit, and not as clear how the common component code is included or updated. Changes to the component interface/parameters must be backwards compatible, which can require multi-step changes, but with established UI patterns this should be less of an issue.
+
+### Avoid complexity
+
```
selfthinker: Typo: two "can"s in `can the template logic can be implemented`.

doc/govuk-components.md
```diff
@@ -9,10 +9,58 @@ The available components and their documentation are exposed by an API at `/temp
...
+```html
+<div class="govuk-title length-average">
+  <h1>My page title</h1>
+</div>
+```
+
+Apps should treat the output of a component like a [black box](https://en.wikipedia.org/wiki/Black_box), not relying on the internal implementation as it may change (eg, additional CSS based on the `govuk-title` or `h1` selector)
+
+If additional behavior in the component is needed then it should be made in the component itself, then used in an application.
```
tvararu: I think [higher-order functions/components](https://facebook.github.io/react/blog/2016/07/13/mixins-considered-harmful.html#higher-order-components-explained) (functions that augment other functions and return a new, better one) is also a good pattern to re-use certain types of behaviours between multiple components.

doc/govuk-components.md
```diff
@@ -9,10 +9,58 @@ The available components and their documentation are exposed by an API at `/temp
...
+
+### Distribute via the network
+
+One of the goals of GOV.UK Components is to have consistency between the many different frontend applications. Because of this components live in a central place (the `static` application) and are distributed to the different frontends over the network. This is done via Slimmer, in a similar way to [layout templates](slimmer_templates.md).
+
+This allows a component to be changed and deployed within `static` and each frontend application will automatically fetch a new version of the [component template code]((../app/views/govuk_component).
+
+The automatic updating allows changes to be made to every GOV.UK frontend app, without having to manually update and deploy them. With 20+ frontend apps, this makes updating components much easier.
+
+However, this approach does have some costs. It's less explicit, and not as clear how the common component code is included or updated. Changes to the component interface/parameters must be backwards compatible, which can require multi-step changes, but with established UI patterns this should be less of an issue.
```
tvararu: By multi-step I assume you mean something like [forward-only migrations](http://engineering.skybettingandgaming.com/2016/02/02/how-we-release-so-frequently/)?

FTFA:

> Take dropping a column as an example; how do you release that change? Easy:
> 
> - Release a version of the code that doesn’t use that column; ensure it is stable / won’t be rolled back.
> - Do a second release that has a migration to remove the column.

Applied to changing a component's public API:

- Release a version of static that introduces a new parameter.
- Migrate frontends to use new API and stop depending on old one.
- Release version of static to remove old parameter.

dsingleton: Yep, just like that - incidentally, thats how most GOV.UK do DB migrations, so it's a familiar pattern.

Your bullet list makes it much clearer, so i've incorporated that into the doc :)

dsingleton:
Addressed comments, hopefully it's more useful now ;)

tvararu:
LGTM👌

dsingleton:
Great - anyone happy to hit merge?

gemmaleigh:
👍

PR #818, opened 2016-08-01
Add missing govspeak example callout styles

steventux:
https://govuk.zendesk.com/agent/tickets/1386394

The example callout styles are missing from the govspeak component and giving a different rendering when comparing private-frontend to public frontends.

screenshot from 2016-08-01 17 22 35

### DIFF
--- a/app/assets/stylesheets/govuk-component/_govspeak.scss
+++ b/app/assets/stylesheets/govuk-component/_govspeak.scss
@@ -530,4 +530,14 @@
       margin-right: (($thumbnail-width * -1) - $gutter-half);
     }
   }
+
+  .example {
+    border-left: 10px solid $panel-colour;
+    padding-left: 1.5em;
+    margin: 2.5em 0;
+
+    strong {
+      display: block;
+    }
+  }
 }
--- a/app/views/govuk_component/docs/govspeak.yml
+++ b/app/views/govuk_component/docs/govspeak.yml
@@ -507,3 +507,11 @@ fixtures:
           </div>
         </div>
       </section>
+  example:
+    content: |
+      <div class="example">
+        <p>
+          This is an indented example block.<br/>
+          It may span multiple lines, <a href="#">contain links</a>.
+        </p>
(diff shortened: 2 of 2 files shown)


gpeng:
Looks good to me. Maybe hang on for @nickcolley to have a look before merging.

app/assets/stylesheets/govuk-component/_govspeak.scss
```diff
@@ -530,4 +530,18 @@
...
+    border-left: 10px solid $panel-colour;
+    padding-left: 1.5em;
+    margin: 2.5em 0;
+
+    strong {
+      display: block;
+    }
+
+    @include media(mobile) {
+      margin: 0.75 0;
```
nickcolley: With no units on the 0.75 this won't work, this looks to be a problem upstream?

nickcolley:
@steventux can you comment if these styles were found somewhere else? Just wondering if we can reuse styles here instead of moving them here..

steventux:
@nickcolley yes these were lifted from [here](https://github.com/alphagov/static/blob/master/app/assets/stylesheets/helpers/_text.scss#L327)

nickcolley:
When I add the correct units on smaller viewports the callouts look like the following (current on the left, after on the right)

This has been broken since 2015 so going to check if we actually want the tighter whitespace now..

image

gemmaleigh:
We have an ["inset text" style in GOV.UK elements](http://govuk-elements.herokuapp.com/typography/#typography-inset-text) that looks similar to this. 

It would be good if we could use the same style for both, or have a reason why this callout style should be different.

nickcolley:
We also need to update https://github.com/alphagov/static/blob/master/app/views/govuk_component/docs/govspeak.yml with this usecase too..

steventux:
thanks @nickcolley and @gemmaleigh, updated

robinwhittleton:
Looks OK to me (and @nickcolley / @dsingleton) so I’ll merge.

PR #817, opened 2016-07-28
Fix scroll tracking events for GA Campaigns and Cosmetic changes

ikennaokpala:
[Trello card](https://trello.com/c/ZRqYEHab/57-suppress-duplicate-scroll-tracking-events)

## Description

This PR contains three sets of changes:

* Cosmetic changes: Remove duplicate function names (c2281c492).
* Handling path comparison better (16866febe12e).
* Reducing the scroll delay. By reducing the timeout delay the the scroll tracking is fired more frequently and making it hard to miss out on faster scrolling users. This at the moment works and deemed experimental (03906389).

PR #819, opened 2016-08-03
Quick fix scroll tracking events for GA campaigns and cosmetic changes

selfthinker:
This is a continuation of closed #817.
[Trello card](https://trello.com/c/ZRqYEHab/57-suppress-duplicate-scroll-tracking-events)

## Description

This PR contains five sets of changes:

* Cosmetic changes: Remove duplicate function names.
* Handling path comparison better.
* Reducing the scroll delay. By reducing the timeout delay the the scroll tracking is fired more frequently and making it hard to miss out on faster scrolling users. This at the moment works and deemed experimental. 
* To reduce the effect the reduction of the scroll delay has for users, this removes most paths which we don't need to track anymore.
* Remove all code related to percent scroll tracking as that proved not to be useful.

## Changes to previous PR

1. Renamed `concatPath` to `normalisePath` as suggested by @dsingleton.
2. @nickcolley looked at potential performance issues regarding the reduction of the scroll delay to 10 ms and found it's probably not that bad. And we initially said we will just go with it and look into potential issues later.
3. To reduce those potential issues we removed 16 out of the previous 20 paths this script would run on.
4. Remove code related to percent scroll tracking and changed tests accordingly.

@nickcolley later mentioned that a much better fix would be to use throttling instead of a timeout. As this task has been going around since 22 July and has a deadline (8 Aug) that will be properly fixed in a subsequent PR.

### DIFF
--- a/app/assets/javascripts/analytics/scroll-tracker.js
+++ b/app/assets/javascripts/analytics/scroll-tracker.js
@@ -154,7 +154,9 @@
 
   ScrollTracker.prototype.getConfigForCurrentPath = function (sitewideConfig) {
     for ( var path in sitewideConfig ) {
-      if ( window.location.pathname == path ) return sitewideConfig[path];
+      if (this.normalisePath(window.location.pathname) == this.normalisePath(path)) {
+        return sitewideConfig[path];
+      }
     }
   };
 
@@ -169,6 +171,10 @@
     }
 
     return nodes;
+  };
+
+  ScrollTracker.prototype.normalisePath = function (path){
+    return path.split("/").join("");
   };
 
   ScrollTracker.prototype.onScroll = function () {
--- a/app/assets/javascripts/analytics/scroll-tracker.js
+++ b/app/assets/javascripts/analytics/scroll-tracker.js
@@ -4,34 +4,6 @@
   window.GOVUK = window.GOVUK || {};
 
   var CONFIG = {
-    '/': [
-      ['Heading', 'Services and information'],
-      ['Heading', 'More on GOV.UK'],
-      ['Percent', 80] //To track 'Services and information' section in footer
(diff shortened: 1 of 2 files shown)


nickcolley:
Just to clarify from trello

"I had a think about how we can improve the reliability of the events fired.

Currently we're debouncing events on scroll, this means that when a new scroll event fires it will resets the timer each time until the user stops scrolling.

We could consider throttling instead which would mean there is always consistent events firing at the interval we expect.

This should allow us to keep a respectable delay and make sure that the events fire across the page.

See http://drupalmotion.com/article/debounce-and-throttle-visual-explanation for an example of debouncing and throttling mouseevents."

In summary it seems the underlying problem was in debouncing events rather than the interval in which they were debounced.

https://trello.com/c/ZRqYEHab/57-review-time-delay-for-header-scroll-tracking-events#comment-57a1ba1fac54b5272731748b

selfthinker:
Sorry, I tried to fix the failing test, will look at it again tomorrow. I'd be grateful If someone could give me a hint.

tvararu:
It does look like https://github.com/alphagov/static/pull/819/files#diff-5a474506625a3d56d10f933b501c0d4cR82 is better suited to a throttle than a debounce pattern.

That being said, with the impending deadline, I think we should ship this and revise it at a later date.

LGTM. 👍

selfthinker:
FYI, there won't be any subsequent PR as implementing throttling did not solve the original issue. I tested the implementation (34943b90f7655b75664a2e43a15a1cb950cab4cc) and found it doesn't log the scroll tracking reliably over a delay of a meagre 48ms.
We should rather keep on questioning if the scroll tracking makes sense in general and answers the right questions.

PR #821, opened 2016-08-04
Update template and toolkit to latest versions

robinwhittleton:
- Fix logo issues and document using tabindex to fix skiplink (template 0.18.3)
- Degrade gracefully when external JS can’t be loaded (template 0.18.2)
- Fix a .visually-hidden bug on GOV.UK (template 0.18.1)
- Slims down the page weight for IE8 (template 0.18.0)
- Amend GOVUK.stickAtTopWhenScrolling to "unstick" for smaller screens (toolkit 4.18.3)
- Remove `@media` width query from print CSS and use npm script rather than grunt-shell (toolkit 4.18.2)
- Fix error in IE from shimLinksWithButtonRole JS (toolkit 4.18.1)
- Add GOVUK.ShowHideContent JavaScript (toolkit 4.18.0)
- Add class to selection buttons per radio and checkbox (toolkit 4.17.0)
- Fix anchor-buttons.js (toolkit 4.16.1)
- Add Department for International Trade organisation (toolkit 4.16.0)
- Normalise keyboard behaviour between buttons and links with a button role (toolkit 4.15.0)
- Fixes a bug with the sizing of tabular numbers (toolkit 4.14.1)
- Makes headings block-level by default (toolkit 4.13.0)

I’m somewhat worried about the [last change to headings](https://github.com/alphagov/govuk_frontend_toolkit/pull/200). I can’t think of a place on GOV.UK where we use inline headings on purpose, but I don’t have a comprehensive knowledge of the system. Would like a review by someone on the GOV.UK team.


nickcolley:
When I updated whitehall we had an issue with the new `display: inline` behaviour here https://github.com/alphagov/whitehall/pull/2704/commits/c77f836b821ec513f25e36501023a20540ec7449

So we should probably do a similar with the consumers of Static..

robinwhittleton:
Oops, of course. So yes, no problem there.

robinwhittleton:
Rebased and bumped to the latest versions. Also edited my original comment with the new stuff it’s pulling in.

selfthinker:
As this was already quite old, I just bumped both gems to the very latest version, rebased and force-pushed.

tvararu:
Sorry for taking so long to get around to reviewing this. Have reviewed, tested locally, all seems well.

Great work! 👍

nickcolley:
Thanks @tvararu 👍  🎉

selfthinker:
Woohoo! Thanks!

PR #823, opened 2016-08-08
Migrate link href printing from template

robinwhittleton:
[govuk_template](https://github.com/alphagov/govuk_template) prints relative links with [a forced gov.uk prefix](https://github.com/alphagov/govuk_template/blob/master/source/assets/stylesheets/_core-print.scss#L24). This works for gov.uk but not for other services that consume the template. First step to fix this is to insert the prefix as part of static. After this is live we can remove the prefix override from the template (i.e. change the link href printing selector to `a[href^="/"]:after, a[href^="http://"]:after, a[href^="https://"]:after`).

### DIFF
--- a/app/assets/stylesheets/helpers/_print-base.scss
+++ b/app/assets/stylesheets/helpers/_print-base.scss
@@ -4,6 +4,11 @@ a {
   text-decoration: none;
 }
 
+// Override relative link printing to include gov.uk
+a[href^="/"]:after {
+  content: " (https://www.gov.uk" attr(href) ")";
+}
+
 h1,
 h2,
 h3,


nickcolley:
Could we make the suffix just `attr(href)` upstream?

robinwhittleton:
Yes, we will do. This maintains the existing gov.uk behaviour for gov.uk though.

nickcolley:
I mean, if we simplify upstream we don't need to be specific here. I'd rather not duplicate the code here if we can.

robinwhittleton:
@nickcolley pointed out a better way of doing this that will require a different PR.

robinwhittleton:
Reopened: this is the current best way.

nickcolley:
For reference to others, I talked to @robinwhittleton about having the domain specified as a SASS variable that could be set by consumers. The idea being we don't duplicate any code where possible.

Doesn't look like this is so simple right now so will revisit later on.

LGTM 👍

PR #824, opened 2016-08-09
Contract "we are" to "we're" on error page

alexmuller:
This is GOV.UK style. Suggested by Tom Hughes.


nickcolley:
LGTM thanks @alexmuller (and Tom :P) 👍

PR #828, opened 2016-08-23
Remove webchat banner on VAT Enquiries page

daibach:
This page will be getting the new style of HMRC webchat content. So the banners are no longer required.

Related to a PR on `contacts-frontend` to enable the new webchat there. Must be deployed before, or alongside that change.


tvararu:
👍  looks good to me, shouldn't need anything else on `static`.

PR #835, opened 2016-09-13
Feedback form event tracking

selfthinker:
[Trello card](https://trello.com/c/mGyllSZu)

To make it easier to gather statistics on which areas get the most feedback, we'd like to track the feedback intent of users by tracking the revealing of the form and its submit (not success or failure).
While I was working on those file I also fixed some linting issues (but two are still in there).

### DIFF
--- a/app/assets/javascripts/report-a-problem-form.js
+++ b/app/assets/javascripts/report-a-problem-form.js
@@ -8,7 +8,7 @@
     $form.on('submit', this.submit.bind(this));
 
     this.appendHiddenContextInformation();
-  }
+  };
 
   ReportAProblemForm.prototype.appendHiddenContextInformation = function() {
     // form submission for reporting a problem
@@ -46,7 +46,7 @@
 
   ReportAProblemForm.prototype.setUrl = function() {
     this.$form.find('input#url').val(window.location);
-  }
+  };
 
   ReportAProblemForm.prototype.postFormViaAjax = function() {
     $.ajax({
--- a/app/assets/javascripts/report-a-problem-form.js
+++ b/app/assets/javascripts/report-a-problem-form.js
@@ -65,6 +65,7 @@
   ReportAProblemForm.prototype.submit = function(evt) {
     this.hidePrompt();
     this.setUrl();
+    this.trackEvent('GOVUK Send Feedback');
     this.disableSubmitButton();
     this.postFormViaAjax();
 
@@ -77,6 +78,12 @@
 
   ReportAProblemForm.prototype.triggerSuccess = function(data) {
     this.$form.trigger('reportAProblemForm.success', data);
+  };
+
+  ReportAProblemForm.prototype.trackEvent = function(action){
+    if (GOVUK.analytics && GOVUK.analytics.trackEvent) {
+      GOVUK.analytics.trackEvent('Onsite Feedback', action, { label: '(not set)' });
+    }
   };
 
   GOVUK.ReportAProblemForm = ReportAProblemForm;
--- a/spec/javascripts/report-a-problem-form-spec.js
+++ b/spec/javascripts/report-a-problem-form-spec.js
@@ -33,9 +33,17 @@ describe("form submission for reporting a problem", function () {
       $form.triggerHandler('submit');
 
       expect($.ajax).toHaveBeenCalled();
-
       args = $.ajax.calls.mostRecent().args;
       expect(args[0].url).toBe('ajax-endpoint');
+    });
+
+    it("should track the event", function() {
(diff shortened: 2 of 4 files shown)


spec/javascripts/report-a-problem-form-spec.js
```diff
@@ -30,9 +30,13 @@ describe("form submission for reporting a problem", function () {
     it("should submit using ajax to the action specified by the form", function() {
       var args;
       spyOn($, "ajax");
+      spyOn(GOVUK.analytics, "trackEvent");
       $form.triggerHandler('submit');
 
       expect($.ajax).toHaveBeenCalled();
+      expect(GOVUK.analytics.trackEvent).toHaveBeenCalledWith(
```
pmanrubia: I would create a separate test for tracking events.

spec/javascripts/report-a-problem-spec.js
```diff
@@ -14,11 +14,20 @@ describe("form submission for reporting a problem", function () {
 
   describe("clicking on the toggle", function(){
     it("should toggle the visibility of the form", function() {
+      spyOn(GOVUK.analytics, "trackEvent");
       expect($form).toBeVisible();
+
       $('.js-report-a-problem-toggle').click();
       expect($form).toBeHidden();
+      expect(GOVUK.analytics.trackEvent).toHaveBeenCalledWith(
```
pmanrubia: Same here. I would create a unit test for the event that is sent when a form is closed.

spec/javascripts/report-a-problem-spec.js
```diff
@@ -14,11 +14,20 @@ describe("form submission for reporting a problem", function () {
...
+        'Onsite Feedback', 'GOVUK Close Form', jasmine.any(Object)
+      );
+
       $('.js-report-a-problem-toggle').click();
       expect($form).toBeVisible();
+      expect(GOVUK.analytics.trackEvent).toHaveBeenCalledWith(
+        'Onsite Feedback', 'GOVUK Open Form', jasmine.any(Object)
```
pmanrubia: Same here. I would create a unit test for the event that is sent when a form is open.

app/assets/javascripts/report-a-problem-form.js
```diff
@@ -79,5 +80,11 @@
     this.$form.trigger('reportAProblemForm.success', data);
   };
 
+  ReportAProblemForm.prototype.trackEvent = function(action){
+    if (GOVUK.analytics && GOVUK.analytics.trackEvent) {
```
pmanrubia: I am curious about why we need to test for the existence of `GOVUK.analytics` if this is already exercised in the test.

nickcolley: Due to the way we load modules there's a chance ReportAProblemForm could be loaded without the Analytics module.

spec/javascripts/report-a-problem-form-spec.js
selfthinker: The reason why we didn't do that is because `GOVUK.analytics.trackEvent` is already tested in `analytics/static-analytics-spec.js`.

selfthinker: Or did you mean just separating those lines out into a new section?

pmanrubia: Yes, sorry if my comment was not clear enough. I would do the following change:

* From:

```javascript
it(“should submit using ajax to the action specified by the form”) {
// Test the form is being sent
// Test the event is sent
}
```

* To:

```javascript
it(“should submit using ajax to the action specified by the form”) {
// Test the form is being sent
}

it(“it should track the event: GOVUK Send Feedback”) {
// Test the event is sent
}
```

This way, when you run tests individually, you can read what the actual behaviour of the function / Module / Object is. If you decide, for example, to stop sending those events, you just need to remove the unit test, and not updating an existing one.

selfthinker: That makes sense. I have just updated the tests accordingly.

pmanrubia:
@selfthinker This LGTM 👍 , not sure if you need a review from another Frontend developer to merge into master 
cc/ @nickcolley

selfthinker:
Thanks @pmanrubia. I think that will be fine. It's not a very complex change.
I forgot what the rules are for the frontend repos, if the author or the reviewer merges? I think it was the reviewer, but I will just go ahead and merge.

selfthinker:
Sorry, I chose "rebase and merge" instead of "merge pull request". I somehow thought that would be doing something different to what it did.

PR #842, opened 2016-10-12
Remove need IDs meta tag

rubenarakelyan:
This commit removes the need IDs meta tag. The performance analysts confirmed at a meeting on 12/10/2016 that this tag is no longer required.


nickcolley:
LGTM 👍

PR #843, opened 2016-10-18
Replace "eg" with "like" for report a problem

alexmuller:
https://insidegovuk.blog.gov.uk/2016/07/20/changes-to-the-style-guide-no-more-eg-and-ie-etc/

> Terms like eg, ie and etc, while common, make reading difficult for some.


nickcolley:
Thanks @alexmuller 👍  🎉

Issue #845, opened 2016-10-25
govuk_frontend_toolkit 5.0.0 will remove the ‘accordian’ icons


robinwhittleton:
The spelling has been fixed to be ‘accordion’. These are currently used in [_publisher.scss](https://github.com/alphagov/static/blob/9e8795b262c948a65b798029864cbe04f193262a/app/assets/stylesheets/helpers/_publisher.scss) and [_core.scss](https://github.com/alphagov/static/blob/9e8795b262c948a65b798029864cbe04f193262a/app/assets/stylesheets/helpers/_core.scss), and will need to be updated when toolkit is bumped.

PR #846, opened 2016-10-29
Remove external link icons

robinwhittleton:
This bumps govuk_frontend_toolkit and govuk_template to the latest versions (removing the external link icons and code), and removes references to external links from the stylesheets in static.

It also fixes the spelling of accordian to accordion.

The one non-trivial change is the removal of `display:inline` from external links in `info-notice`. This was added in https://github.com/alphagov/static/commit/2a6a299fce3368ba0752636dc64f9fa54f4a7747 , but I can’t find any reason for it to still be there: there aren’t any bases styles this is overriding any more.


robinwhittleton:
Fixes https://github.com/alphagov/static/issues/845

PR #849, opened 2016-11-04
Add styling for financial tables

selfthinker:
This adds styling for complex financial tables in HTML publications (e.g. for the Budget or Autumn Statement).
Those will be added directly as HTML.

The tables are differentiated from normal tables by the `financial-data` class and can have 3 more classes for table rows (`section-heading`, `total` and `subtotal`) and 1 more class for table cells (`numeric`).

The very same changes were already added to Whitehall (alphagov/whitehall#2796) but as HTML publications will soon be moved to government-frontend, the CSS needs to live somewhere else. It could potentially also live within government-frontend but it makes more sense to be available in any app using the govspeak_html_publication component.

The changes won't affect anything yet, as there are no such tables with the `financial-data` class within HTML publications yet.
Markup for a table you can test with is here: http://output.jsbin.com/kokuyub
This could be tested on e.g. the last budget: /government/publications/budget-2016-documents/budget-2016#table-23-total-managed-expenditure-1

<del>If HTML publications are moved over before the Autumn Statement (23 Nov), then this needs to be live well before that date as well.</del> The HTML publications will be moved over **after** the Autumn Statement, but probably relatively shortly afterwards.

### Without these changes

financial-table-desktop-before
financial-table-mobile-before

### With these changes

financial-table-desktop-after
financial-table-mobile-after


selfthinker:
I fixed all the linting issues except for the 7 occurrences of "QualifyingElement". I will explain my reasons for keeping them in comments on here. I'd like to challenge the very existence of that linting option, but if the consensus is that we should keep it, I'd like to know this should best be rewritten.

app/assets/stylesheets/govuk-component/_govspeak-html-publication.scss
```diff
@@ -109,6 +109,76 @@
         margin-top: ($gutter * 2) + $gutter-two-thirds;
       }
     }
+
+    table.financial-data {
```
selfthinker: I could potentially use `.financial-data-table` here instead. But these styles will always be only for tables and having `table.financial-data` makes that clearer and also easier to read. They are here to distinguish from a bare `table` and should be seen as a modifier to it.

app/assets/stylesheets/govuk-component/_govspeak-html-publication.scss
```diff
@@ -109,6 +109,76 @@
...
+
+      // needs padding due to occasional background colour
+      // is intentionally small and not balanced with other padding
+      // as tables can be quite wide
+      tr > :first-child {
+        padding-left: $gutter-one-third / 2;
+      }
+
+      // add spacing so that groupings are clearer
+      tr.section-heading > * {
```
selfthinker: This is my way of saving a few characters and making writing this less error-prone. Instead of `tr.foo th, tr.foo td` it saves the extra line by compressing that to `tr.foo > *`. If this was the two lines I could just write `.foo th, .foo td` but the single line only works if `.foo` is a `tr`. That's why I wouldn't want to change that to `.foo > *` as that makes it usable on the wrong elements.
So, the only way I could rewrite this to fit the linting configuration is by writing this instead:
```
.section-heading th,
.section-heading td {
}
```
But I'd prefer to disable the linting option.
This is the same for the other 4 occurrences of `tr > *` below.

app/assets/stylesheets/govuk-component/_govspeak-html-publication.scss
```diff
@@ -109,6 +109,76 @@
...
+      }
+
+      tr.total > *,
+      tbody tr.subtotal > * {
+        font-weight: bold;
+      }
+
+      // the total is usually in the tfoot, so already has that background colour
+      // but when it's used inside the tbody, it should also be highlighted
+      tr.total {
```
selfthinker: I would want the `.total` class only to be applicable to a `tr` as it's important for the design to be applied to the whole row.
But this is the instance I feel least strongly about and could change it to `.total`.

fofr:
Can you provide some before and after screenshots?

selfthinker:
@fofr, there is no "before" as the tables don't exist yet. For the "after" see the example link I provided on JSbin. Or do you mean you would like to see how a generic table looks on government-frontend?

fofr:
@selfthinker The before screenshot would be how the complex table is rendered without any of the new styles applied, ie existing table styles.

The JSBin is great, but I would expect to see screenshots of the new styles rendered in context and at different viewport widths.

fofr:
An example of the markup could be added to the HTML publication govspeak component documentation here: https://github.com/alphagov/static/blob/master/app/views/govuk_component/docs/govspeak_html_publication.yml

Which would render here:
http://govuk-component-guide.herokuapp.com/components/govspeak_html_publication

selfthinker:
@fofr, I've added some screenshots to the description and added an example table to the documentation.

I am not 100% sure if that is the right place, though. I mean I am not sure if the CSS itself is in the right place. Personally, I feel this should rather be either in the Govspeak component or maybe even somewhere more global. But in case the CSS ever "moves up", it will be easy to move the documentation somewhere else alongside it as well. And I guess that should happen whenever the need arises for financial tables to be used outside of HTML publications.

fofr:
Thanks @selfthinker, the tables look good. I think both are in the right place.

fofr:
Can you add the linting exceptions and we can merge this PR? Happy to continue the linting debate elsewhere, perhaps in the frontend meeting on Wednesday? I'd rather it didn't hold up this change.

(The toolkit has been versioned with your font variable changes, so that could also be updated)

selfthinker:
I have done all the changes, including the update of the frontend toolkit.

When we briefly discussed the linting issue at the frontend meeting before last, Robin said it would be better to discuss in a PR (that is alphagov/govuk-lint#62 not this one) because it will reach a wider audience and won't be a decision of a few. But unfortunately it's not really working like that either.

fofr:
Thanks @selfthinker, looks good to me. Needs a rebase before it can be merged.

Issue #676, opened 2015-11-19
Active items in guides are outdented vs other parts


edwardhorsford:
Active items in gov.uk guides are outdented vs the other parts. This as the effect of making the parts jump around if you click between the parts.

![screen shot 2015-11-19 at 13 43 25](https://cloud.githubusercontent.com/assets/2204224/11273774/1ae0a26a-8ecd-11e5-89f6-b1c87aad7936.png)

nickcolley:
Not sure why this was done: https://github.com/alphagov/static/commit/11302c0274a45ba193c0148b4c41cdd092ee9736

edwardhorsford:
Can we fix it?

nickcolley:
@edwardhorsford yes, I'm going to do it :)

nickcolley:
Thanks for raising, once we deploy static you should see this fixed in production 😄

PR #858, opened 2016-12-01
Fix inconsistent spacing in page navigation

nickcolley:
Fixes https://github.com/alphagov/static/issues/676

I made this change inline and was not able to run this locally, raising so this it's more likely to be picked up...


fofr:
LGTM :+1:

PR #859, opened 2016-12-02
Avoid rendering the side content when there's no content

Davidslv:
This commit aims to superseed https://github.com/alphagov/frontend/pull/1069
We believe that is the component responsability to make sure the requirements are met
in order to do the the job, in this case, to render the content.

What this commit suggest is that we do our components with [Design by Contract](https://en.wikipedia.org/wiki/Design_by_contract) in mind,
what it means is that the component becomes responsible to dictate how it should be used.

Also there's a certain advantage of doing this at the component level,
this avoids extra work on the developer to chase all the rendering applications to make sure they should or not render the component.

So keeping it on the component side is not only beneficial code wise but also more efficient.

Trello: https://trello.com/c/5TCdhKVN/320-fix-the-hovering-blue-sidebar-on-pages-without-related-links

screen shot 2016-12-01 at 14 57 01

screen shot 2016-12-01 at 14 57 12

### DIFF
--- a/test/govuk_component/related_items_test.rb
+++ b/test/govuk_component/related_items_test.rb
@@ -7,8 +7,7 @@ class RelatedItemsTestCase < ComponentTestCase
 
   test "no error if no parameters passed in" do
     assert_nothing_raised do
-      render_component({})
-      assert_select ".govuk-related-items"
+      assert_empty render_component({})
     end
   end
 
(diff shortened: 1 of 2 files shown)


test/govuk_component/related_items_test.rb
```diff
@@ -8,7 +8,6 @@ def component_name
   test "no error if no parameters passed in" do
     assert_nothing_raised do
       render_component({})
-      assert_select ".govuk-related-items"
```
fofr: Some components use this:
`assert_empty render_component({})`

I think that would be better now.

Davidslv: ah! Thanks, that makes much more sense!

PR #861, opened 2016-12-07
Add and init button role shim

### LINKED ISSUE #825: Implement keyboard events for links with a role of button
govuk_frontend_toolkit 4.15.0 will include [a shim](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/javascripts/govuk/anchor-buttons.js) for fixing keyboard events on links with a role of `button`. In the frontend meeting it was decided that we’d implement this in static initially to test, and then decide whether to migrate it over to govuk_template. @cfq also requested that this usage was recorded in analytics too so that we can see how many people this affects.

nickcolley:
Closes #825

Shims any applications that are using `[role="button"]` so that using the spacebar will click them as expected.

# How to test

Check out this branch locally and also a frontend application with a button (for example a start page).

Tab to the button, then press the spacebar, the button should be clicked in the same behaviour as pressing enter.

### DIFF
--- a/app/assets/javascripts/core.js
+++ b/app/assets/javascripts/core.js
@@ -25,11 +25,13 @@ $(document).ready(function() {
     var buttonsSelector = "label.selectable input[type='radio'], label.selectable input[type='checkbox']";
     new GOVUK.SelectionButtons(buttonsSelector);
 
+    if (GOVUK.shimLinksWithButtonRole) {
+      GOVUK.shimLinksWithButtonRole.init();
+    }
+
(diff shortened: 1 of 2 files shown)


nickcolley:
I've tested this with @fofr on integration for the start pages and it looks good.

fofr:
A couple of non-blocking caveats:
* Holding down space bar loads the next page then scrolls to the bottom of it
* Shift + Space to scroll back up also loads the page despite an intention not to

PR #863, opened 2016-12-14
Add tracking data attributes to breadcrumbs

carvil:
These will be used to report tracking events to Google Analytics (cc/ @fofr )

PR to add JS tracker coming shortly. On that PR I will add the `data-module` attribute to specify which module to call.

For more details, please see https://github.com/alphagov/govuk_frontend_toolkit/pull/359

The new markup looks like this (for https://www.gov.uk/benefits-calculators as an example):

```
<ol>
    <li>
        <a data-track-category="breadcrumbClicked" data-track-action="1" data-track-label="/" data-track-dimension="Home" href="/">Home</a>
    </li>
    <li>
        <a data-track-category="breadcrumbClicked" data-track-action="2" data-track-label="/browse/benefits" data-track-dimension="Benefits" href="/browse/benefits">Benefits</a>
    </li>
    <li>
        <a data-track-category="breadcrumbClicked" data-track-action="3" data-track-label="/browse/benefits/entitlement" data-track-dimension="Benefits entitlement" href="/browse/benefits/entitlement">Benefits entitlement</a>
    </li>
  </ol>
```

Trello: https://trello.com/c/buYUPz1l/247-enable-tracking-of-breadcrumb-clicks-across-applications-rendering-education-content

PR #864, opened 2016-12-14
Track breadcrumb clicks

carvil:
This PR adds a new tracker that tracks clicks on breadcrumb links. When a click occurs, it:
- sets a custom dimension in order to report the title of the link clicked;
- calls the `TrackLink` module to track a custom event (`trackEvent`) with the remaining data attributes (category, action and label).

**NOTE**:

Before this can be merged, we need to:
- [x] - change the dimension ID to a correct value (we agreed on 29);
- [x] - merge https://github.com/alphagov/static/pull/863

This branch was created off https://github.com/alphagov/static/pull/863, hence the first 2 commits being the same.

### DIFF
--- a/app/views/govuk_component/breadcrumbs.raw.html.erb
+++ b/app/views/govuk_component/breadcrumbs.raw.html.erb
@@ -13,7 +13,8 @@
             track_category: 'breadcrumbClicked',
             track_action: index + 1,
             track_label: crumb[:url],
-            track_dimension: crumb[:title]
+            track_dimension: crumb[:title],
+            module: 'track-breadcrumb-click'
           }
         ) %>
       <% else %>
--- a/app/assets/javascripts/modules/track-breadcrumb-click.js
+++ b/app/assets/javascripts/modules/track-breadcrumb-click.js
@@ -1,0 +1,23 @@
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function(Modules) {
+  "use strict";
+
+  Modules.TrackBreadcrumbClick = function () {
+    this.start = function (element) {
+      element.on('click', trackBreadcrumbClick);
+
+      var titleDimensionId = 29,
+          dimension = element.data('track-dimension'),
+          trackClick = new GOVUK.Modules.TrackClick();
+
+      trackClick.start(element);
+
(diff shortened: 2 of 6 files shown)


fofr:
This looks good to me.

My only query is the order of the click handlers on the link – is it possible that the analytics gets sent to Google before the custom dimension is set?

app/views/govuk_component/breadcrumbs.raw.html.erb
```diff
@@ -3,10 +3,20 @@
...
       <% if crumb[:url] %>
-        <a href="<%= crumb[:url] %>"><%= crumb[:title] %></a>
+        <%= link_to(
+          crumb[:title],
+          crumb[:url],
+          data: {
+            track_category: 'breadcrumbClicked',
+            track_action: index + 1,
+            track_label: crumb[:url],
+            track_dimension: crumb[:title],
```
fofr: You could infer this in `track-breadcrumb-link` using `element.text()`

carvil: In the initial implementation I inferred most of the data via Javascript, but now that we are adding data attributes I thought it would be clearer to leave it there even if it looks a bit repetitive. This way I think it's clearer for someone in the future having a look. Does that make sense?

carvil:
@fofr good point about the order of events. The tests I ran locally seem to send the events in the expected order. I think the reason for this is that the order we set listeners on the element matters.

When a page is loaded, it activates the `TrackBreadcrumbClick` module via the `data-module` attribute for each of the links. That module first sets an event listener on the link with:

```
element.on('click', trackBreadcrumbClick);
```

And then calls `start` on the `TrackClick` module, which in turn sets its event listener for the same element. This means that when a `click` event is triggered, they will be called in order.

I will deploy this to integration and make sure the reporting in analytics is correct, though.

carvil:
@fofr thanks for reviewing this PR. I added 1 more commit last week to deal with an issue we had with `data()`, could you please have a look? If that's ok I will push this live today.

fofr:
ebe69aa looks good, good catch. It's a shame the GA API isn't more lenient in what it accepts.

carvil:
Thanks!

PR #867, opened 2016-12-16
Fix invalid value for rel attribute

selfthinker:
There is a typo in the previous_and_next_navigation component as the "previous" link has an invalid [link type](https://www.w3.org/TR/html5/links.html#linkTypes).
This corrects `rel="previous"` to the correct `rel="prev"`.


fofr:
Good catch! It looks like we've got this problem in a few other places too:
https://github.com/search?q=user%3Aalphagov+rel%3D%22previous%22&type=Code&utf8=%E2%9C%93

PR #873, opened 2017-01-05
Add a new layout that excludes footer links

36degrees:
The links that appear in the top half of the footer (.footer-categories) replicate the home page of GOV.UK and represent the needs of the citizens using the site.

As the primary users of the Service Manual (and forthcoming Service Toolkit) are people within government building a services these links are not relevant and may even compete (could be confused with Service Manual content).

### DIFF
--- a/app/views/root/_base.html.erb
+++ b/app/views/root/_base.html.erb
@@ -44,46 +44,48 @@
 <% end %>
 
 <% content_for :footer_top do %>
-  <div class="footer-categories">
-    <div class="footer-explore">
-      <h2>Services and information</h2>
+  <% unless local_assigns[:hide_footer_links] %>
+    <div class="footer-categories">
+      <div class="footer-explore">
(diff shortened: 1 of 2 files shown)


nickcolley:
Asked Oliver on Slack about the possibility of redefining `footer_top`.

> Oliver: *Might* be able to flush it after?
> <% content_for :footer_top, flush: true do
  " "
end %>
> it’s perhaps less obvious?

36degrees:
I *think* I’m inclined to go with the current way (using `hide_footer_links`) because 1) it’s more obvious and 2) there’s [prior art for it in `hide_nav`](https://github.com/alphagov/static/blob/ec8ef20c036caef89a5fbd27fe3053ac13f6e59a/app/views/root/_base.html.erb#L33-L39)…

nickcolley:
👍  I think it's definitely clearer your intent with this way

PR #875, opened 2017-01-06
Track Related Item clicks

alecgibson:
This change will add tracking for clicks on the "Related Item" links.

Trello card: https://trello.com/c/ybdOyNSu/288-enable-tracking-of-related-links-clicks-across-all-gov-uk-applications

The change is almost a direct read-across of the same change for Breadcrumbs:
* https://github.com/alphagov/static/pull/863
* https://github.com/alphagov/static/pull/864

Note that we are reusing the "Link Text" Custom Dimension 29 (See: https://gov-uk.atlassian.net/wiki/display/GOVUK/Analytics+on+GOV.UK#AnalyticsonGOV.UK-customDimensionsCustomdimensions&suk=ff808081595cb8c4015963b402330000)


alecgibson:
Please review, but don't merge yet - I'm going to run this past @carvil when he's back on Monday.

nickcolley:
I'm not that well versed with the Analytics code yet but it seems from first glance that we could add the dimensions to the already existing `track-click` module and remove the breadcrumb module?

I'll add @fofr as a reviewer :)

app/assets/javascripts/modules/track-link-click.js
```diff
@@ -3,17 +3,20 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+      element.on('click', onClick);
+
+      var titleDimensionId = LINK_TEXT_CUSTOM_DIMENSION_SLOT,
           dimension = element.data('track-dimension'),
           trackClick = new GOVUK.Modules.TrackClick();
 
       trackClick.start(element);
 
-      function trackBreadcrumbClick(e) {
+      function onClick(e) {
```
fofr: This method name should be descriptive of what the function does, not the circumstance that it's called in

fofr:
I agree with @nickcolley, I don't see the need for both `track-click` and `track-link-click`.
It's essentially, "track-click" with a custom dimension set.

app/assets/javascripts/modules/track-click.js
```diff
@@ -8,16 +8,26 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       element.on('click', trackClick);
 
       var options = {},
-          category = element.attr('data-track-category'),
-          action = element.attr('data-track-action'),
-          label = element.attr('data-track-label');
+          category = element.data('track-category'),
```
carvil: I think we should use `attr()` instead of `data()` when fetching data from these data attributes, more details here: https://github.com/alphagov/static/commit/ebe69aa77528fd669ff34ce1f093fa45026cd451#diff-5649f4c1766856649d723e2ef9a026b3

app/views/govuk_component/breadcrumbs.raw.html.erb
```diff
@@ -14,7 +14,8 @@
             track_action: index + 1,
             track_label: crumb[:url],
             track_dimension: crumb[:title],
-            module: 'track-breadcrumb-click'
+            track_custom_dimension: 29,
```
carvil: We could perhaps set this in a static variable with a name that explains when this is in case people wonder what the number 29 is all about. Something like: 

```
LINK_TEXT_CUSTOM_DIMENSION=29
```

And then use it throughout. What do you think?

spec/javascripts/modules/track-click.spec.js
```diff
@@ -24,6 +25,59 @@ describe('A click tracker', function() {
...
+
+    expect(GOVUK.analytics.setDimension).toHaveBeenCalledWith(29, 'Home');
+    expect(GOVUK.analytics.trackEvent).toHaveBeenCalledWith(
+      'breadcrumbClicked',
+      1,
+      { label: '/' }
+    );
+  });
+
+  it('tracks related item click events', function() {
```
carvil: Could you please also add a test that makes sure we call `GOVUK.analytics.trackEvent` but don't call `GOVUK.analytics.setDimension` when the custom dimension attribute isn't present?

app/views/govuk_component/breadcrumbs.raw.html.erb
alecgibson: Sounds good to me - where should it live?

spec/javascripts/modules/track-click.spec.js
alecgibson: Aren't the changes to `'tracks click events'` sufficient? (lines 11 & 28). Or did you want it to be more explicit than that?

app/assets/javascripts/modules/track-click.js
alecgibson: Will do. Although the dimension has always been a number rather than a string - should we keep it that way? (Ie use `element.data('track-custom-dimension')` where all the others are `.attr`?

spec/javascripts/modules/track-click.spec.js
alecgibson: As discussed, this is fine.

app/assets/javascripts/modules/track-click.js
alecgibson: As discussed, custom dimension should remain data because Google Analytics expects an integer.

app/assets/javascripts/modules/track-click.js
```diff
@@ -10,14 +10,24 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
       }
 
       function trackClick() {
-        if (GOVUK.analytics && GOVUK.analytics.trackEvent) {
+        if (!GOVUK.analytics) {
+          return;
+        }
+
+        if (customDimension && GOVUK.analytics.setDimension) {
+          GOVUK.analytics.setDimension(customDimension, dimension);
```
nickcolley: If `dimension` is not set, could this break?

alecgibson: I've added a check and some tests for this.

app/assets/javascripts/modules/track-click.js
```diff
@@ -10,14 +10,24 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       var options = {},
           category = element.attr('data-track-category'),
           action = element.attr('data-track-action'),
-          label = element.attr('data-track-label');
+          label = element.attr('data-track-label'),
+          dimension = element.attr('data-track-dimension'),
+          customDimension = element.data('track-custom-dimension');
```
fofr: Are these two variables the custom dimension index and the custom dimension value?

app/assets/javascripts/modules/track-click.js
```diff
@@ -10,14 +10,24 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
       }
 
       function trackClick() {
-        if (GOVUK.analytics && GOVUK.analytics.trackEvent) {
+        if (!GOVUK.analytics) {
+          return;
+        }
+
+        if (customDimension && dimension && GOVUK.analytics.setDimension) {
+          GOVUK.analytics.setDimension(customDimension, dimension);
```
fofr: I think this dimension will remain set for all subsequent tracked-clicks.

eg If I cmd+click and open a page in a new tab. Then on the old page if I click an external link or a download – (two types of links that get tracked), those will be tracked with this dimension.

app/assets/javascripts/modules/track-click.js
alecgibson: Yes I believe so - I can rename if you want? (I'm still only just getting my head around everything!)

app/assets/javascripts/modules/track-click.js
alecgibson: Very possible - I have no idea how this is put together. How do I "unset" it? Set it to undefined?

alecgibson: (I guess technically for compatibility it should be set to `null`?)

fofr: Rather than setting the dimension I think we should pass it as an option into the event itself:

See: https://github.com/alphagov/govuk_frontend_toolkit/blob/c4acf649030d8a7be1e1533117d36f9bebbe038b/javascripts/govuk/analytics/analytics.js#L37-L45

And:
https://developers.google.com/analytics/devguides/collection/analyticsjs/custom-dims-mets

eg
```
GOVUK.analytics.trackEvent(category, action, {dimension29: "dimensionValue"});
```

That _should_ work.

alecgibson: As a side-note, I'm not sure we do track downloads? For example the PDF on the [Computing Curriculum](https://www.gov.uk/government/publications/national-curriculum-in-england-computing-programmes-of-study) page doesn't have any tracking markup, and I don't see any events fired off to GA in my Network tab. Unless it's a specific type of download?

alecgibson: Similarly, I don't see any tracking markup on the link to the DHS Website on [this page](https://www.gov.uk/foreign-travel-advice/usa/terrorism)? Are you aware of an actual case I can test?

alecgibson: (The above solution looks good to me, but I'd like to confirm on a test case if you can find one?)

fofr: That's configured here:
https://github.com/alphagov/static/blob/9e8795b262c948a65b798029864cbe04f193262a/app/assets/javascripts/analytics/static-analytics.js#L26-L28

And here:
https://github.com/alphagov/govuk_frontend_toolkit/blob/master/javascripts/govuk/analytics/download-link-tracker.js

Though you're right, I can't see anything being tracked.

app/assets/javascripts/modules/track-click.js
fofr: I think it'd be clearer if renamed

alecgibson:
This pull request is now dependent upon the deployment of https://github.com/alphagov/govuk_frontend_toolkit/pull/365

alecgibson:
@fofr @nickcolley this should now be good to go. It was waiting on the Frontend Toolkit Gem getting bumped (which it now has). I've tested this change end-to-end with my changes to the Frontend Toolkit Gem.

carvil:
I've tried it in integration and I can see the GA events being fired correctly. 👍

PR #874, opened 2017-01-05
Revert "Upgrade jQuery to 1.12.4"

gpeng:
Reverts alphagov/static#869 Breaks things... Related to https://github.com/jquery/jquery/issues/2824

PR #876, opened 2017-01-06
Add unit test for jQuery tabs plugin

fofr:
This is not an exhaustive test. It tests the fundamental behaviour of the app in its two modes; accordion and tabs.

It aims to test that the jQuery tabs plugin:
* Runs set up code without error
* Configures tab links and panes
* Hides and shows panes when clicking links

The lack of a unit test meant we missed an error produced when jQuery was upgraded: 
https://github.com/alphagov/static/pull/874
https://github.com/jquery/jquery/issues/2824

When upgrading jQuery again, this test will fail with the error:
```
Error: Syntax error, unrecognized expression: a[href$=#]
```

cc @nickcolley @gpeng

### DIFF
--- a/spec/javascripts/jquery-tabs-spec.js
+++ b/spec/javascripts/jquery-tabs-spec.js
@@ -1,0 +1,163 @@
+describe("jQuery tabs", function () {
+  var TABS = '\
+        <div class="tabs-container">\
+          <div class="js-tabs nav-tabs">\
+            <ul>\
+                <li><a href="#tab-1">Tab 1</a></li>\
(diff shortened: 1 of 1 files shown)


fofr:
When writing this test I noticed that the aria attributes it generates eg `aria-controls` and `aria-labelledby` don't appear to take into account that the plugin itself is altering the ID of tab pane elements (it appends `-enhanced`). I believe this means the aria attributes on tabs will be pointing at an element ID that does not exist on the page, making the tabs inaccessible.

@cfq can we test this?

Issue #871, opened 2017-01-04
[target=_blank] to external websites vulnerability 


nickcolley:
We're using [target=_blank] which has a vulnerability (https://mathiasbynens.github.io/rel-noopener/)

- https://github.com/alphagov/static/blob/caac6ba9a5768957c22f6440ea01bf865100db55/app/assets/javascripts/surveys.js#L12

This has been fixed in government-frontend with the following PR alphagov/government-frontend#217

nickcolley:
Thanks a lot @h-lame 🎉  👍  💃

PR #877, opened 2017-01-06
Environment user research panel survey

h-lame:
For: https://trello.com/c/kA2QXSPq/91-gov-uk-survey-environment-user-research-panel-sign-up-set-of-9-surveys

Also includes a fix for #871.
Also extends the family visa survey by 4 weeks and removes other expired surveys.


h-lame:
Flagged as DNM because it removes a survey that runs until the 8th, so we shouldn't merge + deploy until at least then.

app/assets/javascripts/surveys.js
```diff
@@ -22,6 +22,45 @@
...
+              + /check-flood-risk/.source
+              + /fishing-licences\/buy-a-fishing-licence/.source
+              + /fishing-licences\/when-you-need-a-licence/.source
+              + /topic\/environmental-management\/environmental-permits/.source
+              + /topic\/environmental-management\/waste/.source
+              + /hunting\/overview/.source
+              + /report-dead-animal/.source
+              + /browse\/environment-countryside\/flooding-extreme-weather/.source
+              + /sign-up-for-flood-warnings/.source
+              + ')'
```
h-lame: Is this going to be performant?  I wanted to use a single regexp for all the paths, but it looked unweildy on one line.

app/assets/javascripts/surveys.js
```diff
@@ -53,6 +53,9 @@
 
       var $surveyLink = $('#take-survey');
       var surveyUrl = survey.url;
+      if (typeof surveyUrl === "function") {
+        surveyUrl = surveyUrl();
+      }
```
h-lame: Is this the best way to check if an object is callable?

app/assets/javascripts/surveys.js
nickcolley: Do we need to use `.source`?

nickcolley:
```js

var timeBeforeExecution = performance.now(); 
var pathMatchingExpr = new RegExp('/(?:' +
  + /check-flood-risk/.source
  + /fishing-licences\/buy-a-fishing-licence/.source
  + /fishing-licences\/when-you-need-a-licence/.source
  + /topic\/environmental-management\/environmental-permits/.source
  + /topic\/environmental-management\/waste/.source
  + /hunting\/overview/.source
  + /report-dead-animal/.source
  + /browse\/environment-countryside\/flooding-extreme-weather/.source
  + /sign-up-for-flood-warnings/.source
  + ')'
);

pathMatchingExpr.test(window.location.pathname);
var timeAfterExecution = performance.now(); 
console.log((timeAfterExecution - timeBeforeExecution) + " milliseconds.");
```

> 0.18500000005587935 milliseconds.

app/assets/javascripts/surveys.js
nickcolley: Is this required? The href="javascript:" should mean the expression is executed?

app/assets/javascripts/surveys.js
```diff
@@ -23,114 +23,44 @@
...
-
-          function organisationMatches() {
-            var orgMatchingExpr = /<D6>|<D106>|<D109>|<EA243>|<EA86>|<EA242>|<EA541>/;
-            var metaText = $('meta[name="govuk:analytics:organisations"]').attr('content') || "";
-            return orgMatchingExpr.test(metaText);
-          }
-
-          return (sectionMatches() || organisationMatches() || breadcrumbMatches());
+        url: function() {
+          return "https://signup.take-part-in-research.service.gov.uk/?utm_campaign="+
```
nickcolley: Nit: Spaces between string and +

app/assets/javascripts/surveys.js
h-lame: I didn't think you could add regexps together, so that's why we use `.source`.  Well you can add regexps together directly, but `/foo|bar/ + /baz/` generates `"/foo|bar//baz"` whereas `/foo|bar/.source + /baz/.source` giving us `"foo|barbaz"` which is (in this case at least) what we want when pushed into `new RegExp` (the regexp adding version gives us `/\/foo|bar\/\/baz\//` and the `.source` version gives us `/foo|barbaz/`).

I guess that timing is ok though, so maybe we don't need to worry?

app/assets/javascripts/surveys.js
h-lame: The `javascript:void;` href is a placeholder, it's replaced by the value of `surveyUrl` (maybe there's a better way?).

app/assets/javascripts/surveys.js
nickcolley: Yeah 👍

app/assets/javascripts/surveys.js
h-lame: Realised we didn't need to support functional urls at all (at least for this survey) so we've dropped this entirely.

h-lame:
Removed DNM flag as we now want to extend the "expired" survey (change included in this PR).

NeilvB:
Looks good 👍

PR #880, opened 2017-01-16
Add back removed CSS for sidebar positioning

tijmenb:
In https://github.com/alphagov/static/pull/879 I removed CSS for breadcrumbs and related links. The `related-container` CSS should not have been deleted, because it determines the size of the sidebar (rather than the appearance of it, which is determined by the govuk component).

## Before

screen shot 2017-01-16 at 16 20 21

## After

screen shot 2017-01-16 at 16 19 58

https://trello.com/c/7wohUbyo

### DIFF
--- a/app/assets/stylesheets/helpers/_core.scss
+++ b/app/assets/stylesheets/helpers/_core.scss
@@ -18,6 +18,7 @@ body {
   }
 }
 
+// Needed for the positioning of the sidebar. Deprecated.
 .related-container {
   @include grid-column(1/3, $full-width: desktop);
   margin-top: 40px;
--- a/app/assets/stylesheets/helpers/_core.scss
+++ b/app/assets/stylesheets/helpers/_core.scss
@@ -18,6 +18,10 @@ body {
   }
 }
 
+.related-container {
+  @include grid-column(1/3, $full-width: desktop);
+  margin-top: 40px;
+}
 
 p {
   @include core-19;


nickcolley:
Is it worth also including some information to suggest that this selector is now deprecated?

We're in the nice position with this that we know why some legacy CSS exists and how it will be removed so it'd be good to document that for the future. 💃

carvil:
Looks good to me 👍 

It would be nice to link to the preferred way of doing this going forward so we know how to change it. We will be changing a few content pages with the new navigation and I'm sure we will come across this exact issue.

tijmenb:
@nickcolley @carvil the reason I ended up not commenting is that I don't know what the upgrade path is and what the new thing looks like. If there's more explaining you can do, could you add that in a PR?

carvil:
@nickcolley is the preferred way to have something like this?

```
<div class="grid-row">
  <div class="column-two-thirds">content 2/3 on the left</div>
  <div class="column-one-third">content at 1/3 on the right</div>
</div>
```

I am not sure either, but would be nice to clarify it.

nickcolley:
We'd need to make sure the application is using the '[core_layout](https://github.com/alphagov/static/blob/master/doc/slimmer_templates.md#core_layout)' layout.

Then indeed there's access to the grid by default https://github.com/alphagov/static/blob/master/app/assets/stylesheets/helpers/_layouts.scss#L8

carolinegreen:
@carvil I would have thought that is the preferred way.

carvil:
Ok, good to know. I don't think we need to update the message further, this discussion should be enough for now I think.

PR #881, opened 2017-01-18
[WIP] Add accordion with descriptions component

alecgibson:
This accordion has been copied and pasted from `service-manual-frontend` to be re-used in `collections`.

Corresponding Pull Request for `service-manual-frontend`: https://github.com/alphagov/service-manual-frontend/pull/97

However, we need to think about how to unit test this component in the apps that use it, because it currently won't be rendered.

We also need to think about the change process, since Finding Things are currently doing user research and testing, which may mean the component could be volatile in the short term.

Trello: https://trello.com/c/FPzDHPRc/324-make-accordionwithdescriptions-available-outside-the-service-manual-frontend-app-as-a-component


carvil:
We will pick this work up again once the new navigation is live and we have agreed on an implementation for it.

PR #882, opened 2017-01-20
Improve print styles

fofr:
* Indicate end of content and beginning of footnotes
* Ensure headings are bold when printing
* Hide sticky contents link when printing
* Use decimal list styles for ordered lists - fixes https://govuk.zendesk.com/agent/tickets/1593796
* Update mainstream guide print file to use modern typography mixins
   * Delete old and un-used typography mixins

| Feature | Old | New |
| --------| ----| -----|
| Mainstream guide | mainstream_guide_print_old | mainstream_guide_print_new
| Mainstream | screen shot 2017-01-20 at 13 02 44 | screen shot 2017-01-20 at 13 02 24
| HTML publication | screen shot 2017-01-20 at 13 16 53 | screen shot 2017-01-20 at 13 17 31


app/assets/stylesheets/styleguide/_typography.scss
```diff
@@ -1,218 +0,0 @@
-@import "_print.scss";
-
```
nickcolley: Are we sure these mixins were not being used? Just thinking these could be out of date and be used, so us removing them could be a regression.

fofr: `_print` sets:

```
@mixin print {
  @if $is-print {
    @content;
  }
}
```

The only use of `@include print` was in `styleguide/_typography`. If something had been using it I would have expected to see a compile error from sass.

nickcolley: I was referring to the type mixins, is that the same story?

fofr: I've remove the only reference to `styleguide/_typography.scss`. There are no Sass compile errors.

PR #883, opened 2017-01-20
Allow tracking of all trackable links within a container

fofr:
Creating a single event listener is more efficient than one for every link that could be clicked. It also reduces the number of unique modules that must be started.

* Use new pattern on breadcrumbs and related links
* I've tested the correct events are generated in frontend

Will make https://trello.com/c/62Gy8pyj/77-track-interactions-with-contents-lists-1-s more efficient.

cc @tijmenb @alecgibson @nickcolley

### DIFF
--- a/app/assets/javascripts/modules/track-click.js
+++ b/app/assets/javascripts/modules/track-click.js
@@ -5,24 +5,36 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
   Modules.TrackClick = function () {
     this.start = function (element) {
-      element.on('click', trackClick);
+      var trackable = '[data-track-category][data-track-action]';
 
-      var options = {transport: 'beacon'},
-          category = element.attr('data-track-category'),
-          action = element.attr('data-track-action'),
-          label = element.attr('data-track-label'),
-          dimension = element.attr('data-track-dimension'),
-          dimensionIndex = element.attr('data-track-dimension-index');
-
-      if (label) {
-        options.label = label;
+      if (element.is(trackable)) {
+        element.on('click', trackClick);
+      } else {
+        element.on('click', trackable, trackClick);
       }
 
-      if (dimension && dimensionIndex) {
-        options['dimension' + dimensionIndex] = dimension;
-      }
+      function trackClick(evt) {
+        var $el = $(evt.target),
+            options = {transport: 'beacon'};
 
-      function trackClick() {
+        if ( ! $el.is(trackable)) {
+          $el = $el.parents(trackable);
+        }
+
+        var category = $el.attr('data-track-category'),
+            action = $el.attr('data-track-action'),
+            label = $el.attr('data-track-label'),
+            dimension = $el.attr('data-track-dimension'),
+            dimensionIndex = $el.attr('data-track-dimension-index');
+
+        if (label) {
+          options.label = label;
+        }
+
+        if (dimension && dimensionIndex) {
+          options['dimension' + dimensionIndex] = dimension;
+        }
+
         if (GOVUK.analytics && GOVUK.analytics.trackEvent) {
           GOVUK.analytics.trackEvent(category, action, options);
         }
--- a/app/views/govuk_component/breadcrumbs.raw.html.erb
+++ b/app/views/govuk_component/breadcrumbs.raw.html.erb
@@ -1,7 +1,7 @@
 <%
   breadcrumbs ||= []
 %>
-<div class="govuk-breadcrumbs">
+<div class="govuk-breadcrumbs" data-module="track-click">
   <ol>
   <% breadcrumbs.each_with_index do |crumb, index| %>
     <li>
@@ -15,7 +15,6 @@
             track_label: crumb[:url],
             track_dimension: crumb[:title],
             track_dimension_index: 29,
-            module: 'track-click',
           }
         ) %>
       <% else %>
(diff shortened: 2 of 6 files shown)


fofr:
Closing to add a further feature.

app/assets/javascripts/modules/track-click.js
```diff
@@ -5,24 +5,36 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
-          dimension = element.attr('data-track-dimension'),
-          dimensionIndex = element.attr('data-track-dimension-index');
-
-      if (label) {
-        options.label = label;
+      if (element.is(trackable)) {
+        element.on('click', trackClick);
+      } else {
+        element.on('click', trackable, trackClick);
       }
```
alecgibson: Wouldn't it be easier to register a listener on all "trackable" descendants?
`element.find(trackable).click(trackClick);` (You'd then be able to drop the slightly weird `parents` check, in `trackClick`).

fofr: That would create more event listeners and be less efficient.

alecgibson: Is the performance difference actually noticeable? I'm not convinced that it's worth obscuring the code like this.

nickcolley: I'm not sure how sophisticated jQuery's event delegation is however if it works as I think it does then using `.find` will add a listener for each element, rather than one listener.

There might be a way we can delegate all links however which would make the code simpler to understand and be performant.

alecgibson: Oh I think it will almost certainly add a listener to each node. My point is: does it matter? This feels a little like premature optimisation. I do like the simplicity of a single module instantiation, but I think optimising away a handful of extra event listeners on a page is surely overkill?

fofr: Here's our guidance on it:
https://github.com/alphagov/govuk_frontend_toolkit/blob/master/docs/javascript.md#declare-event-listeners-at-the-start

alecgibson: Yeah, I'm fine with it in principle, I just think it's making this particular bit of code less readable. Of course fine to leave - just my two cents.

fofr: Here's a perf test for the various patterns:
http://jsperf.com/click-perf

fofr: I could move the logic into a named function so it's clear what's happening?

nickcolley: I'd agree if this wasn't generic, but since this could be used anywhere and with any amount of links, I'd rather us just do this now since memory leaks are hard to debug and crop up at the worst times.

alecgibson: How about scrapping this `if` statement (ie call the handler on any click inside this element). Then in the handler, you could simply say `var $el = $(evt.target).closest(trackable)` and return early if no trackable element was found?

nickcolley:
I've tested this branch alongside government-frontend and it's working as expected, do we need to update some docs to indicate the new usage?

fofr:
I don't think there are docs at the moment.

alecgibson:
@fofr How well does this cope with nesting? What happens in these cases?
1. Two tracked items are nested within one another (eg this possibly happens in the `service-manual-frontend` accordion?)
2. Two tracking modules are nested inside each other. For example, let's say so many elements are tracked on a page that someone decides to move the tracking module up to some wrapper. However, that wrapper might contain, eg breadcrumbs, which also instantiate the module. Would we get a double tracking event fired?

fofr:
@alecgibson The modules shouldn't be nested.

1. The most nested item would be tracked
2. Click events would be tracked twice

Do you have an example of either of these happening intentionally?

If more advanced tracking of multiple features with a single click was a required feature we'd develop for it when it was needed.

alecgibson:
@fofr My point is more that it's easy to accidentally nest modules (for example, you might forget that breadcrumbs already instantiate the module). Given that the whole point of this module is to accurately track how many times things were clicked on, I think this is important to guard against.

As for complex element tracking, see https://github.com/alphagov/service-manual-frontend/blob/master/app/assets/javascripts/modules/accordion-with-descriptions.js#L284

These are both nasty side-effects of this event delegation pattern (if listeners are attached to all the DOM nodes instead, then an event is fired every time an event propagates through them, as expected).

If this is a pattern that we intend to stick to, I'd suggest creating some sort of well-tested JavaScript Event Delegator object, who takes a callback and a selector, and makes sure that things like nesting are gracefully handled. This sort of refactoring could certainly be left for the next time we implement this pattern, but I definitely think we should be guarding against weird nesting behaviour, because it's highly unintuitive and could be hard to track down.

fofr:
> it's easy to accidentally nest modules (for example, you might forget that breadcrumbs already instantiate the module).

If we see this become a problem we can fix it then, and build an advanced event delegator when we need to. I agree there is some risk, but also that the risk is small and similar to misuse of the module pattern we already have.

We have a similar pattern in place on admin tools and haven't had these issues:
https://github.com/alphagov/govuk_admin_template/blob/master/app/assets/javascripts/govuk-admin-template/modules/track_click.js

alecgibson:
@fofr sounds okay to me - the thought just occurred to me and I thought I'd flag it. If you think it's an acceptable risk, then that's fine.

nickcolley:
Thanks @fofr and @alecgibson 👍  😄

PR #889, opened 2017-01-31
Fix breadcrumb and section matchers for Education survey

h-lame:
In #879 we replaced the `#global-breadcrumb` container for breadcrumbs
with the govuk component rendered version which uses `.govuk-breadcrumbs`
as the container instead.  This broke our `breadcrumbMatches()` function
so we change the function to grab the correct selector.

We also change the section matcher to use a regexp test instead of direct
equality to check for a matching section.  It's not clear if the section
names have recently changed, or if the matcher never worked but it wasn't
detected because the breadcrumbs worked instead.

### DIFF
--- a/app/assets/javascripts/surveys.js
+++ b/app/assets/javascripts/surveys.js
@@ -80,13 +80,13 @@
         frequency: 10,
         activeWhen: function() {
           function breadcrumbMatches() {
-            var text = $('#global-breadcrumb').text() || "";
+            var text = $('.govuk-breadcrumbs').text() || "";
             return (/Education/i.test(text) || /Childcare/i.test(text) || /Schools/i.test(text));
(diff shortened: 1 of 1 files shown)


fofr:
We should have some tests around this.

h-lame:
@fofr I think you're right, but I'm not sure what they should be and how best to protect us.  For example, the [current tests for surveys](https://github.com/alphagov/static/blob/d00183fe0c82938023e3c01f31d1712d6a1bec35/spec/javascripts/surveys-spec.js) use fixtures, which wouldn't have protected us in this case.  Integration tests don't render components and even if they did the surveys themselves don't run 100% of the time so we need to factor that into any integration-style test.  Do you have time to talk this through somewhere?

PR #890, opened 2017-02-02
Add block of Google Analytics dimensions for A/B tests

suzannehamilton:
The GA dimensions from 40-49 have been allocated to A/B tests. These may be reserved for any A/B test for the duration of the test.

This commit identifies which dimension to use for the A/B test based on an extra `data` attribute added to the `meta` tag. Only dimensions in the allowed range are permitted. Any others are ignored and are not reported to GA.

It would be good to have some feedback on this approach. I'm aiming to make it as simple as possible for new teams to add an A/B test. With this proposal, they would have to do the following:

- Request one of the dimensions in the 40-49 block from the performance analysts for the duration of the A/B test. They will assign one which isn't currently being used.
- Set a meta tag in the header like `<meta name="govuk:ab-test" content="NameOfTest:#{bucket}" data-analytics-dimension="43">`.
  - Ideally, this would be done through the [AB test gem](https://github.com/alphagov/govuk_ab_testing). The application code would just need to pass the experiment name and dimension to the gem.

https://trello.com/c/g9OyC7Pn/310-tell-google-analytics-about-a-b-tests

Before this can be merged:

- [x] Merge #888 and change the merge base of this PR
- [ ] Add the data attribute to the `meta` tag on the A/B example page in frontend
- [ ] Add the data attribute to the `meta` tag in the [A/B testing gem](https://github.com/alphagov/govuk_ab_testing) (alphagov/govuk_ab_testing#5)

### DIFF
--- a/app/assets/javascripts/analytics/static-analytics.js
+++ b/app/assets/javascripts/analytics/static-analytics.js
@@ -13,6 +13,7 @@
     setPixelDensityDimension();
     setHTTPStatusCodeDimension();
     this.setDimensionsFromMetaTags();
+    this.setAbTestDimensionsFromMetaTags();
     this.callMethodRequestedByPreviousPage();
 
     // Track initial pageview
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/analytics/static-analytics.js
```diff
@@ -94,9 +95,26 @@
...
+          // This is the block of dimensions assigned to A/B tests
+          minAbTestDimension = 40,
+          maxAbTestDimension = 49;
+
+      $abMetas.each(function() {
+        var $meta = $(this),
+        dimension = parseInt($meta.data('analytics-dimension')),
+        testNameAndBucket = $meta.attr('content');
+
+        if (dimension >= minAbTestDimension && dimension <= maxAbTestDimension) {
```
suzannehamilton: Is it OK to silently ignore invalid or out-of-range dimensions? I'd prefer to throw an error here to alert the developer to the bad configuration, but I don't want people to accidentally break all analytics and not notice.

tijmenb:
Closing this temporarily to focus on the education A/B testing. We'll revisit after that's live. 

Trello: https://trello.com/c/QoJsEjgw/381-better-a-b-testing

suzannehamilton:
@tijmenb - I'd like to reopen this because we need to add another dimension for the Education Navigation A/B test, and I think this change is the easiest way to do that. Is that OK with you?

The alternative would be to give the Example and EducationNavigation `meta` tags different names, but that won't be compatible with more general A/B test analytics which will make the switchover harder when we release this change.

36degrees:
_I've changed the required status checks for this repo. You'll need to rebase this branch against master so that this branch gets a Jenkinsfile can be built against new CI._

PR #891, opened 2017-02-03
Add new environment short survey

NeilvB:
Note that at the time of writing this survey is running concurrently
with another survey that affects a small number of pages. It was decided
to retain the existing survey for these pages rather than choose between
multiple candidate surveys per page. By placing the new survey at the top of the
array we ensure that it will be "overridden" by a survey that also
applies where that survey is found later in the array.

Trello: https://trello.com/c/J8PdbakJ/108-gov-uk-survey-environment-short-survey

### DIFF
--- a/app/assets/javascripts/surveys.js
+++ b/app/assets/javascripts/surveys.js
@@ -23,6 +23,34 @@
     },
     smallSurveys: [
       {
+        url: 'https://www.surveymonkey.co.uk/r/6QC8RV8',
+        identifier: 'environment_survey',
+        template: '<section id="user-satisfaction-survey" class="visible" aria-hidden="false">' +
+                  '  <div class="wrapper">' +
+                  '    <h1>Why did you come to GOV.UK today?</h1>' +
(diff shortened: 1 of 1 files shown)


nickcolley:
Hey @NeilvB 

When we update the JavaScript we force any users that had cached versions to have to redownload so I think we should be keeping config out of our bundles if possible.

Do you know how regular surveys are going to be?

If they're going to be as regular as they have been so far we should consider putting this config baked into markup (even if it's the static templates).

This way the survey code can look for the metadata in the page and display if needed, without us having to update the JavaScript.

There might be other solutions but just a thought.

NeilvB:
Thanks @nickcolley for the info. It looks like we'll continue to have new surveys incoming at a similar rate so your tips sound good.

h-lame:
@nickcolley - I've written this up in our trello board and tagged you to it (so we remember to talk to you when we get round to doing something about it.  I don't mean to suggest that you do it!)

PR #892, opened 2017-02-07
Tidy up the GovSpeak component

36degrees:
The styles for the GovSpeak component are currently in a bit of a mess - a 500 line CSS file that doesn't appear to be in any logical order (at least, not that I could fathom).

This PR moves the baseline typography and table styles and the various different GovSpeak 'extensions' into individual files inside a govspeak subdirectory.

I have not attempted to refactor, consolidate or remove any of the CSS.

I have:

- Compiled assets whilst on both this branch and the `master` branch, taken header-footer.css from each (which includes the GovSpeak component), sorted the files alphabetically and diffed them. [The only differences were in the comments that report where a style has come from, and the order of some selectors](https://gist.github.com/36degrees/9f72d5a7fecc33ff9a3402e8e180db6b).
- Used wraith to capture screenshots of the [GovSpeak component page](http://govuk-component-guide.herokuapp.com/components/govspeak) whilst on both branches. Wraith reports no differences.

I suggest reviewing on a commit-by-commit basis.


app/assets/stylesheets/govuk-component/govspeak/_magna-charta.scss
```diff
@@ -0,0 +1,306 @@
...
+  &.mc-negative .mc-td.mc-key-cell {
+    padding-right: 25px;
+  }
+}
+
+.mc-outdented .mc-bar-cell.mc-bar-negative {
+  text-align: left;
+}
+
+// .mc-bar-cell.mc-bar-negative.mc-bar-indented {
```
robinwhittleton: Can we tidy up commented out code if we’re losing direct blame relations anyway?

fofr:
Really happy to see this, @nickcolley and I had been planning something similar.

fofr:
Did you come across any CSS styles that weren't documented in the component guide during your refactoring?

app/assets/stylesheets/govuk-component/govspeak/_example.scss
```diff
@@ -0,0 +1,11 @@
+// Scoped to govspeak container when included
```
fofr: Is it worth explaining what each feature is used for at the top of each file?

I'm not sure what `.example` is for, for instance.

nickcolley:
This is really great thanks 👍 

I spoke to Oliver and suggested we intentionally scope each include so that the decision to reuse, any of these styles requires more than just importing them. This is so that a discussion is more likely to be had, is my hope 😄

36degrees:
@nickcolley @fofr @robinwhittleton happy that those more recent commits address your concerns / comments?

fofr:
:+1: Great docs and updates.

After scoping to govspeak container, presumably wraith still shows no change?

Errors are linting problems: `[W] Comment: Use `//` comments everywhere`

36degrees:
Wraith still shows no change; additionally, the production CSS comes out at exactly the same size.

nickcolley:
Nice work @36degrees 👍  🎉

PR #895, opened 2017-02-08
Fix draft watermark for short content

gpeng:
Some content is clipping the draft background watermark. This commit adds `background-size: contain` to avoid this.

Before:
screen shot 2017-02-08 at 18 20 36

After:
screen shot 2017-02-08 at 18 21 07

Before:
screen shot 2017-02-08 at 18 25 03

After:
screen shot 2017-02-08 at 18 25 20

I've not tested the effect on IE so would welcome an opinion on that.

/cc @fofr (thanks for the pointer on how to fix), @nickcolley

### DIFF
--- a/app/assets/stylesheets/helpers/_draft.scss
+++ b/app/assets/stylesheets/helpers/_draft.scss
@@ -2,4 +2,5 @@
   background-image: image-url("draft-watermark.png");
   background-repeat: repeat-y;
   background-position: 50% 0;
+  background-size: contain;
 }


nickcolley:
You can see that `contain` is only supported in IE9+ (https://developer.mozilla.org/en/docs/Web/CSS/background-size)

So I'd look at using a fallback some variation of `background-size: 100%`

Also, how does this work with print styles, is this respected?

fofr:
@nickcolley Printing does print background images.

36degrees:
_I've changed the required status checks for this repo. You'll need to rebase this branch against master so that this branch gets a Jenkinsfile can be built against new CI._

fofr:
Pragmatically I think this is an enhancement and we don't _need_ a fallback. IE8 has functional support.

nickcolley:
Since the biggest issue is on smaller viewports and they tend to be mobile clients that have support for contain, I'd say :shipit:

gpeng:
Thanks all. Post approval commit is just a rebase to pick up @36degrees's Jenkins business...

PR #896, opened 2017-02-09
Use govuk.sassLinter

36degrees:
Having made [changes to the govuk.sassLinter method](https://github.com/alphagov/govuk-puppet/pull/5459) to allow it to take a path or set of paths as an argument, we can now use it whilst passing our custom path.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -2,11 +2,6 @@
 
 REPOSITORY = 'static'
 
-def sassLinter() {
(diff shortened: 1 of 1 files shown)


36degrees:
I _think_ this is currently failing because although the Puppet changes have been deployed, Puppet hasn't actually run on integration yet.

Going to give it half an hour and try re-building.

PR #899, opened 2017-02-10
Add callouts to Govspeak component; Remove information-block

36degrees:
## Background

We would like to be able to use callouts in the Service Manual. The CSS for the callouts is currently only defined in [`helpers/_text.scss`](https://github.com/alphagov/static/blob/master/app/assets/stylesheets/helpers/_text.scss), which is not included when using the `core_layout` template in static.

This adds styling for the various [Govspeak callouts](http://govspeak-guide.herokuapp.com/#callouts) to the Govspeak component, which is included in core_layout.

As part of this work, I have also **removed the `information-block` class** – after a fair amount of digging we can't find anywhere this is used. See the [commit message](https://github.com/alphagov/static/commit/943d1e1437b8625c3295dd0c1a4ef27d873471d1) for more details.

## Design

I worked with Mia on the template consolidation team, as she's been doing some work to rationalise the different Govspeak callouts we have.

> The image with the blue m’s shows the spacing (in ems!) that I am proposing.
> The address callout uses a box the same width as the other callout borders, with a grey left border (the box and the other borders are white or transparent).

screen shot 2017-02-08 at 11 13 34

I have not touched the address element, as it's a fairly complicated chunk of CSS and I think it's distinct enough to be out of scope for this work. 

## Result

screen shot 2017-02-10 at 10 02 03gmt


app/views/govuk_component/docs/govspeak.yml
```diff
@@ -121,15 +121,12 @@ fixtures:
         <p>My quote</p>
         <p class="last-child">about things</p>
       </blockquote>
-  call_out_blocks:
+  information_block:
```
dsingleton: As `.information-block` was removed in https://github.com/alphagov/static/pull/899/commits/943d1e1437b8625c3295dd0c1a4ef27d873471d1, can the fixture be removed now?

app/assets/stylesheets/govuk-component/govspeak/_call-to-action.scss
```diff
@@ -0,0 +1,25 @@
...
+// http://govuk-component-guide.herokuapp.com/components/govspeak/fixtures/call_to_action
+//
+// Support:
+//
+// - alphagov/whitehall: ✔︎
+// - alphagov/govspeak: ✔︎
+
+.govuk-govspeak {
+  .call-to-action {
+    margin: 2em 0;
```
fofr: Do left and right margins need to be set to 0?

app/assets/stylesheets/govuk-component/govspeak/_information-callout.scss
```diff
@@ -0,0 +1,22 @@
...
+// - alphagov/govspeak: ✔︎
+
+.govuk-govspeak {
+  .info-notice {
+    border-left: 1em solid $panel-colour;
+    padding: 1em 0 1em 1em;
+    margin: 2em 0;
+
+    // Remove margin from the last element
+    p:last-child,
```
fofr: Feels like this _could_ be a mixin and shared between the callouts.

36degrees:
I've messed around with the service manual frontend to get it to render the Govspeak styles using the 'old' wrapper layout, so you can see what it would look like:

screen shot 2017-02-10 at 10 43 10gmt

Note that there are no styles for `call-to-action` in helpers/text – that's something that only existed in the Govspeak component.

Using 'Call to Action' (`$CTA`) also isn't documented in the [GovSpeak guide](http://govspeak-guide.herokuapp.com/).

For that matter, neither are advisories (`@`) – another present in the Govspeak gem which has no styles in this component and is not documented.

app/assets/stylesheets/govuk-component/govspeak/_information-callout.scss
36degrees: Yeah, I was debating it. I kinda feel it's in few enough places that it doesn't warrant hiding things behind abstractions, but I could be persuaded.

nickcolley: Does this need to be qualified? Would `:last-child {}` work?

app/assets/stylesheets/govuk-component/govspeak/_call-to-action.scss
36degrees: I _guess_ not as that should be the default… this is a bit more defensive I guess. And I think I just prefer it to setting `margin-top`, `margin-bottom` and leaving the rest to chance?

app/assets/stylesheets/govuk-component/govspeak/_example.scss
```diff
@@ -7,9 +7,16 @@
 
 .govuk-govspeak {
   .example {
-    border-left: 10px solid $panel-colour;
-    padding-left: 1.5em;
-    margin: 2.5em 0;
+    border-left: 1em solid $panel-colour;
+    padding: 1em 0 1em 1em;
```
nickcolley: I'm 👍  for using `em` in general but is this consistent with our spacing as defined here: https://govuk-elements.herokuapp.com/layout/?

> spacing between elements is usually 5px, 10px, 15px or multiples of 15px


I'd prefer we redefine our spacing system to use `em` rather than be inconsistent.

36degrees: ^ @miaallers thoughts?

app/assets/stylesheets/govuk-component/govspeak/_information-callout.scss
36degrees: Good question. Probably not? It's hard to know exactly what elements are going to be used inside the callout and so what this could affect, but at the same time that's also a good argument for making it a generic `:last-child`…

fofr: @36degrees When I use comments to explain something it usually makes sense to refactor into a named function:

eg `// Remove margin from the last element`
becomes `@include remove-last-element-margin;`

I prefer the selector to be qualified. It's faster and less risky (eg otherwise it would apply to all of the last list items in a list).

If we use a mixin it's easy to add another case if found.

app/assets/stylesheets/govuk-component/govspeak/_example.scss
dsingleton: > I'd prefer we redefine our spacing system to use em rather than be inconsistent.

Is this a recommendation for GOV.UK Publishing, or to all services?

fofr: > I'd prefer we redefine our spacing system to use em rather than be inconsistent.
That's a big change.

For now, using ems in this case improves legibility and makes spacing consistent with other content.

We're using the grid classes to define vertical margins on paragraphs etc though:
https://github.com/alphagov/static/blob/master/app/assets/stylesheets/govuk-component/govspeak/_typography.scss#L16-L17

36degrees: We could move to using multiples of `$gutter-half` (it’s essentially a difference of 15px vs a computed 16px).

It'd be good to better understand Mia's intentions. I'm aware Mia's out of the office today, so let's hold merging this till Monday, and discuss it then.

nickcolley: I know it's a big change that's why I'm wondering why we'd change it only here?

Or is this something we don't keep to standard for on GOV.UK Publishing?

nickcolley: I thought we were using those spacing rules but it seems like we're not so I'm 👍  to merge, we can have a conversation about consistency another time.

nickcolley:
Thanks @36degrees and @miaallers, this is top 👍

PR #900, opened 2017-02-13
Extract TLS Version from cookie and send to GA

h-lame:
For: https://trello.com/c/JX541tka/113-add-tls-version-information-to-google-analytics

The TLS Version supported by the browser will be set as a cookie by the
CDN and so will be available when viewing the page.  We extract the value
from the cookie and send it as the value for custom dimension 16 to
google analytics as part of the initial page view tracking.  Doing it
here has caused some churn in the tests (it's now the 4th arg instead
of the trackPageView message), but it means we don't have to send a
separate GA request so it shouldn't impact performance for the users.

This relies on https://github.com/alphagov/govuk-cdn-config/pull/27 to provide the cookie.

### DIFF
--- a/app/assets/javascripts/analytics/static-analytics.js
+++ b/app/assets/javascripts/analytics/static-analytics.js
@@ -12,6 +12,7 @@
 
     setPixelDensityDimension();
     setHTTPStatusCodeDimension();
+    setTLSVersionDimension();
     this.setDimensionsFromMetaTags();
     this.setAbTestDimensionsFromMetaTags();
     this.callMethodRequestedByPreviousPage();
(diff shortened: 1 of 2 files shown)


fofr:
Is the TLS cookie being added to our list of cookies on https://www.gov.uk/help/cookies?
Has the dimension been configured in Google Analytics?

h-lame:
@fofr: 

re: cookies info - yes, it is on our acceptance criteria which is in the trello card (which I neglected to put into the PR - bad me, rectified now)
re: dimensions - also yes, we got it from Tim LB (its also on the trello card, double apologies for not including it)

h-lame:
Ready to merge now that the CDN config has been deployed.

NeilvB:
Looks good to me

PR #901, opened 2017-02-13
Style breadcrumb anchor links

alecgibson:
This style any breadcrumbs linking to anchors on the current page. This is used in conjunction with https://github.com/alphagov/govuk_navigation_helpers/pull/30 to show users which page they are currently on, when browsing the new taxonomy.

Before:
govuk_frontend_toolkit-current-breadcrumb-before

After:
govuk_frontend_toolkit-current-breadcrumb-after

Trello: https://trello.com/c/MVM5bueS/386-add-current-page-to-taxonomy-breadcrumbs

### DIFF
--- a/app/assets/stylesheets/govuk-component/_breadcrumbs.scss
+++ b/app/assets/stylesheets/govuk-component/_breadcrumbs.scss
@@ -10,7 +10,7 @@
 
   @include breadcrumbs;
 
-  a[href^="#"] {
+  .breadcrumb-for-current-page {
     color: $secondary-text-colour;
     text-decoration: none;
   }
--- a/app/assets/stylesheets/govuk-component/_breadcrumbs.scss
+++ b/app/assets/stylesheets/govuk-component/_breadcrumbs.scss
@@ -9,4 +9,9 @@
   }
 
   @include breadcrumbs;
+
+  a[href^="#"] {
+    color: $secondary-text-colour;
+    text-decoration: none;
+  }
 }
--- a/app/views/govuk_component/breadcrumbs.raw.html.erb
+++ b/app/views/govuk_component/breadcrumbs.raw.html.erb
@@ -5,16 +5,27 @@
   <ol>
   <% breadcrumbs.each_with_index do |crumb, index| %>
     <li>
-      <% if crumb[:url] %>
+      <%
+        is_link = crumb[:url].present? || crumb[:is_current_page]
+        path = crumb[:is_current_page] ? '#content' : crumb[:url]
+        aria_current = crumb[:is_current_page] ? 'page' : 'false'
+        css_class = crumb[:is_current_page] ? 'breadcrumb-for-current-page' : ''
+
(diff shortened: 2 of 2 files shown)


fofr:
> We currently don't display the current page we are on in the breadcrumbs. We have included the current page in the breadcrumbs in lab tests and we decided to keep it.
Are there more details about why this is a good thing?

alecgibson:
@fofr I think @alextea should know more.

alextea:
@fofr we included the current breadcrumb in our latest round of research and there were no negative reactions to the inclusion. Leonie Watson recently posted on Twitter about [aria attributes for breadcrumbs](http://tink.uk/using-the-aria-current-attribute/), her examples included the current page and [I asked her thoughts on this](https://twitter.com/LeonieWatson/status/829755001216049160) – she said it was generally a good idea.

This will included as part of our private beta and won't be released on the whole of GOV.UK to begin with. We will analyse feedback and our on going user research to see if there are any negative effects.

app/assets/stylesheets/govuk-component/_breadcrumbs.scss
```diff
@@ -9,4 +9,9 @@
   }
 
   @include breadcrumbs;
+
+  a[href^="#"] {
```
fofr: Is it possible to do this with a class? `a[href^="#"]` is a little obtuse.

alecgibson: Indeed possible. We'd have to:
- Update the breadcrumb component to accept an arbitrary class (or a flag)
- Update navigation helpers to feed this class/flag in
- Add this class to static
- Possibly undo all the above if we wind up styling "normal" (ie non-taxon) breadcrumbs too (since we could do this in a more universal way in that case)

Is it worth adding this machinery? Or is it over-engineering?

nickcolley: I'd also prefer if this were a class since this is vague to how it's used. But pragmatically let's leave it as is for now.

fofr: Let's leave  as-is for now 👍

alecgibson:
@fofr @nickcolley I've gone for the flag approach so we can get an aria attribute on there too.

nickcolley:
Nice work @alecgibson, good to see the attention to `aria` 👍

carvil:
👍

nickcolley:
Had a peak at https://ljwatson.github.io/design-patterns/aria-current/ and it looks like Léonie is using `aria-current="location"` for breadcrumbs 😱

alecgibson:
@nickcolley the [standard](https://www.w3.org/TR/wai-aria-1.1/#aria-current) is pretty hazy on what `location` means:
> Represents the current location within an environment or context.

I'm not sure this is better than `page`:
> Represents the current page within a set of pages.

PR #905, opened 2017-02-21
Add a taxonomy sidebar component

mobaig:
Intended for use on content pages tagged to the GOV.UK taxonomy.
Initially this should only link to the list of taxons the current page is
tagged to, along with descriptions of each taxon.

### DIFF
--- a/app/assets/stylesheets/govuk-component/_component.scss
+++ b/app/assets/stylesheets/govuk-component/_component.scss
@@ -19,3 +19,4 @@
 @import "breadcrumbs";
 @import "related-items";
 @import "organisation-logo";
+@import "taxonomy-sidebar";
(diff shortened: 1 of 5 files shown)


fofr:
Do you have some screenshots of this in action?

mobaig:
@fofr - just added one. Looks very similar to the related links sidebar at the moment.

carvil:
Looks good to me 👍

nickcolley:
Are we planning on deprecating related links? We should be clear why we need two similar components.

carvil:
@nickcolley related links will stay as-is for quite some time (we are moving slices of data across to the new navigation slowly, starting with education content).

However, the new sidebar will evolve based on user research as the new navigation moves forward. Right now, we have taxon descriptions as a new add-on to it, but we are already experimenting with different behaviours that will diverge quite a bit (currently in the prototype).

I think it would be worthwhile to keep it as a separate component to give us the freedom to iterate on it until the new navigation becomes the standard way to navigate in GOV.UK pages.

What do you think?

nickcolley:
Approach seems great to me, just it feels like it could be confusing having two similar components, with no way to know which would should be used.

carvil:
@nickcolley do you think we should add more documentation to http://govuk-component-guide.herokuapp.com/components/taxonomy_sidebar  ?

We could be explicit that that's a beta feature to be used only with the new navigation work. Or do you think there would be a better way to make it clearer to developers?

PR #907, opened 2017-02-22
Fall back to "en" locale for "en-GB"

tijmenb:
Smart answers uses a "en-GB" locale, but we don't have that in this app. This causes a lot of 404's to be returned and missing translations in smart answers (presumably).

Kibana reports about 3.3 million `404` on this route over the last 24 hours:

screen shot 2017-02-22 at 12 02 16

This is also caused by smart-answers cache being broken, which is fixed in https://github.com/alphagov/slimmer/pull/183.

https://trello.com/c/D9HmkJwI

### DIFF
--- a/app/controllers/root_controller.rb
+++ b/app/controllers/root_controller.rb
@@ -30,6 +30,11 @@ class RootController < ApplicationController
   end
 
   def govuk_locales
+    # Smart Answers uses "en-GB" instead of "en". We don't have a specfic en-GB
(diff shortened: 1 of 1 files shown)


mcgoooo:
looks good 🚢 :shipit:

fofr:
Should smart answers use `en`? Or is `en-GB` correct?

tijmenb:
@fofr I've added an issue here: https://github.com/alphagov/smart-answers/issues/2925

PR #908, opened 2017-02-23
Track values

alecgibson:
This will pass an event value through to the trackers, where it is handled in [`govuk_frontend_toolkit`](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/javascripts/govuk/analytics/google-analytics-universal-tracker.js#L78-L87)

This change is required to pass an event value when tracking navigation grid clicks.

Trello: https://trello.com/c/btaOSNax/426-track-position-of-clicks-in-the-new-taxon-grid-pages

### DIFF
--- a/spec/javascripts/modules/track-click.spec.js
+++ b/spec/javascripts/modules/track-click.spec.js
@@ -90,6 +90,28 @@ describe('A click tracker', function() {
     expect(GOVUK.analytics.trackEvent).toHaveBeenCalledWith('category', 'action', {label: 'Foo', transport: 'beacon'});
   });
 
+  it('tracks clicks with values', function() {
+    spyOn(GOVUK.analytics, 'trackEvent');
+
+    element = $(
+      '<a data-track-category="category" \
+          data-track-action="1" \
+          data-track-label="/" \
+          data-track-value="9" \
+          href="/">Home</a>'
+    );
+
+    tracker.start(element);
+
+    element.trigger('click');
+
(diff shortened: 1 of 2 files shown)


spec/javascripts/modules/track-click.spec.js
```diff
@@ -90,6 +90,28 @@ describe('A click tracker', function() {
...
+    );
+
+    tracker.start(element);
+
+    element.trigger('click');
+
+    expect(GOVUK.analytics.trackEvent).toHaveBeenCalledWith(
+      'category',
+      '1',
+      { label: '/', value: '9', transport: 'beacon' }
```
suzannehamilton: Do we have to worry about converting the value to an integer here or will that happen in `GOVUK.analytics`? What will happen if the value is invalid?

alecgibson: Validation all happens in [`govuk_frontend_toolkit`](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/javascripts/govuk/analytics/google-analytics-universal-tracker.js#L78-L87)

fofr: (If the toolkit didn't do the conversion analytics wouldn’t accept it)
This feels like we should use `data-` rather than `attr` as the coercion to an integer is useful.

suzannehamilton: OK, so it just silently ignores invalid values, which is probably sensible in this case. I was just worried that we might need to do some extra exception handling here.

PR #910, opened 2017-02-24
Add analytics for user navigation journeys

suzannehamilton:
Add Google Analytics dimensions:
 - Whether a user is viewing a navigation or a content page, i.e. where they are in their journey through the site
 - What type of navigation page they visit, e.g. a taxon page or a mainstream browse page

Don't merge this until we have confirmation of the GA dimensions to use. I've temporarily set them to values outside of the GA range.

Trello: https://trello.com/c/SCsj2A9D/425-add-custom-dimension-to-page-view-tracking-on-new-navigation-pages


suzannehamilton:
We've confirmed the GA dimensions as 32 and 33, so this is now ready to go.

PR #912, opened 2017-02-24
Add email address survey type

h-lame:
For: https://trello.com/c/dfSufy85/114-develop-a-new-survey-delivery-method

We want a new survey type that invites the user to enter their email address and then submits this to an endpoint in feedback that will send them an email with more details about the survey they're being invited to take part in.

Flagged as DNM because we expect the backend implementation in feedback (see: https://github.com/alphagov/feedback/pull/192) to be merged and deployed first.

We have a separate card (https://trello.com/c/8YSfHXc2/110-reduce-need-to-change-static-js-when-deploying-surveys) to investigate making new survey's not require invalidating the JS bundle, so I've not thought about that much here.


app/assets/stylesheets/helpers/_header.scss
```diff
@@ -284,5 +284,36 @@
...
+        width: 940px;
+      }
+    }
+
+    label {
+      @include core-14;
+      margin: 0 0 0 20px;
+    }
+
+    div.actions {
```
nickcolley: I don't think we need to over qualify this.

https://github.com/CSSLint/csslint/wiki/Disallow-overqualified-elements

h-lame: Thanks for looking @nickcolley, fixed in a rebase (see: 24ccdefda23574daffea59854083f079cdd687bb)

doc/surveys.md
```diff
@@ -53,6 +59,40 @@ An HTML fragment representing the contents of the survey box. This **MUST** conf
 </section>
 ```
 
+The `#survey-no-thanks` and `#take-survey` elements are required as the surveys code will expect these elements for setting cookies and tracking user behaviour.
+
+#### email survey template
+An HTML fragment representing the interactive UI for entering an email address.  We expect somethign like the following:
```
elenatanasoiu: something*

h-lame: thanks

h-lame:
Removed DNM as feedback is merged and a deploy for both is booked for Thursday.

PR #913, opened 2017-02-27
[Discuss] Limit line length when printing content

fofr:
On pages that use header and footer only, text in paragraphs spread all across the page, making content hard to read when printed.

On core layouts we limit this to 80%. This doesn’t take into account the printed text size, the size of the paper or the orientation (eg line length still bad in landscape, or very short on envelope sizes)

* Use em so that line length is always appropriate
* Use max-width so it’s never longer than the width of the page
* Apply to the main element to constrain all content

This has some negative affects:
* Printing uses more paper 🌲 
* Tables don't spread full width and complex tables don't print nicely

Looking at this as part of https://trello.com/c/MluRcP1W/. Printed foreign travel advice is currently limited to 80% width, other government-frontend documents aren't.

The most printed pages are mainstream content, which is 80% width. This is the most printed page: https://www.gov.uk/national-minimum-wage-rates

## Screenshots

|Feature|Old|New|
|--|--|--|
|Consultation A4 portrait|screen shot 2017-02-27 at 11 17 04|screen shot 2017-02-27 at 11 17 15|
|HTML publication A4 portrait|screen shot 2017-02-27 at 11 20 07|screen shot 2017-02-27 at 11 20 39
|HTML publication A4 landscape|screen shot 2017-02-27 at 11 22 31|screen shot 2017-02-27 at 11 22 45|
|Case study A5 portrait|screen shot 2017-02-27 at 11 27 18|screen shot 2017-02-27 at 11 27 32|
|[HTML publication with tables (second most printed government document)](https://www.gov.uk/government/publications/tax-and-tax-credit-rates-and-thresholds-for-2017-18/tax-and-tax-credit-rates-and-thresholds-for-2017-18)|screen shot 2017-02-27 at 11 32 52|screen shot 2017-02-27 at 11 33 12|

cc @robinwhittleton @nickcolley @markhurrell

### DIFF
--- a/app/assets/stylesheets/helpers/_print-base.scss
+++ b/app/assets/stylesheets/helpers/_print-base.scss
@@ -1,5 +1,9 @@
 /* Global print stylesheet */
 
+main {
+  max-width: 36em;
+}
+
 a {
   text-decoration: none;
 }


nickcolley:
I'd propose we only limit line-length on elements that make sense to for example paragraphs, then we'd get best of both worlds?

markhurrell:
hey both - I agree this is better, but we stopped limiting line-lengths on print styles because we had loads of complaints from government users a while back about paper being wasted by having all that whitespace on the right-hand side

I'd prefer to think about this we get to the print stylesheets on the consolidated templates and are thinking about the designs as a whole

fofr:
@markhurrell I'll proceed with travel advice being rendered at full width rather than 80% then. This will match other templates.

PR #917, opened 2017-02-27
Don't open print links in a new window

fofr:
Print links should use target attribute not JS if they want to open in a new window

* Target _blank has a vulnerability:
https://mathiasbynens.github.io/rel-noopener/ and
https://github.com/alphagov/whitehall/issues/2942
* Links don’t need to be in a new window

### DIFF
--- a/app/assets/javascripts/core.js
+++ b/app/assets/javascripts/core.js
@@ -1,6 +1,4 @@
 $(document).ready(function() {
-  $('.print-link a').attr('target', '_blank');
-
   var $searchFocus = $('.js-search-focus');
   $searchFocus.each(function(i, el){
     if($(el).val() !== ''){


nickcolley:
So I know some of our print pages do funky stuff on a new page.

For example:

1. Visit https://www.gov.uk/become-lorry-bus-driver
2. Scroll to the bottom
3. Click 'Print entire guide'
4. New page opens, with print dialogue.

Would this break that functionality?

It's also worth noting the print links should not go to an external domain.

fofr:
@nickcolley Nothing breaks. The link would act like a normal in-page link. The auto-print uses a `window.onload` function and is unaffected.

1. Visit https://www.gov.uk/become-lorry-bus-driver
2. Scroll to the bottom
3. Click 'Print entire guide'
4. ~~New~~ page opens, with print dialogue.

nickcolley:
I don't feel so comfortable with this change because I don't have lots of devices to test this against, printing behaves very different depending on the operating system so I worry this would introduce an odd behaviour.

That being said, I've not seen an evidence of that testing to make this change in the first place.

Will let someone else also have a look. 👍

dsingleton:
My strength of opinion on this is low, 3/10 for context.

1. Is it still a security issue if the opened link is controlled by us? Or is this about consistently avoiding `target="_blank"`?

2. Is there anyone with context of the [original change](https://github.com/alphagov/static/commit/0909ff1f46f7020731bc87167e289676be065d66)? Or anyone who's observed user research around print since then?

My reckon would be that it opens in a new page to reduce the chance of users getting stuck on the print page - as we know some users have difficulty navigating via the back button. 

I'm not sure  the reasons given for the change are convincing enough to make it. 

> - Target _blank has a vulnerability: https://mathiasbynens.github.io/rel-noopener/ and alphagov/whitehall#2942
> - Links don’t need to be in a new window

There might be other benefits/context i'm not aware of though.

36degrees:
> Links don’t need to be in a new window

I get the whole 'give users control over whether things open in a new window or not', but when it comes to print links specifically I think there's a _pretty_ common pattern for doing so. I'm not convinced that we should remove it without further research.

fofr:
Main focus of this PR should have been:
> Print links should use target attribute not JS if they want to open in a new window

Closing this until we have PRs to add attributes where needed.

PR #920, opened 2017-02-28
Display related links in taxonomy sidebar

alecgibson:
This change updates the taxonomy sidebar to accept a list of related content items under each taxon, and updates the styling to match the latest designs.

This is a **breaking change** which will align the API to the slimmed-down version in [this update to `govuk_navigation_helpers`](https://github.com/alphagov/govuk_navigation_helpers/pull/44). Since taxonomy display is still in alpha, and this is a novel component being used only by the navigation team this breakage is assumed to be acceptable.

# Screenshots

## Before

sidebar-before

## After

sidebar-after

# Trello

https://trello.com/c/EJXC60k1/429-add-more-like-this-results-to-side-navigation-in-frontend-apps

### DIFF
--- a/test/govuk_component/taxonomy_sidebar_test.rb
+++ b/test/govuk_component/taxonomy_sidebar_test.rb
@@ -11,55 +11,103 @@ class TaxonomySidebarTestCase < ComponentTestCase
     end
   end
 
-  test "renders a related items block" do
+  test "renders a taxonomy sidebar" do
     render_component(
-      sections: [
+      items: [
         {
-          title: "Section title",
-          items: [
+          title: "Item title",
+          url: "/item",
+          description: "An item",
+        },
+        {
+          title: "Other item title",
+          url: "/other-item",
+          description: "Another item",
+        },
+      ]
+    )
+
+    assert_select ".govuk-taxonomy-sidebar > :nth-of-type(1) h2", text: "More about Item title"
+    assert_link_with_text_in(".govuk-taxonomy-sidebar > :nth-of-type(1)", "/item", "See everything in Item title")
+    assert_select(".govuk-taxonomy-sidebar > :nth-of-type(1) p", "An item")
+
+    assert_select ".govuk-taxonomy-sidebar > :nth-of-type(2) h2", text: "More about Other item title"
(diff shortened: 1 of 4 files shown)


test/govuk_component/taxonomy_sidebar_test.rb
```diff
@@ -13,53 +13,101 @@ def component_name
 
   test "renders a related items block" do
```
mobaig: This was actually my doing from a previous PR, but could you update this test description to say 'taxonomy sidebar' instead?

test/govuk_component/taxonomy_sidebar_test.rb
```diff
@@ -13,53 +13,101 @@ def component_name
...
+          related_content: [
+            {
+              title: "Related link 1",
+              link: "/related-link-1",
+            },
             {
-              title: "Item title",
-              url: "/item",
-              description: "An item",
+              title: "Related link 2",
```
mobaig: Can we remove data for Related link 2 and 3 if we aren't making assertions about them below? Unnecessary test setup.

alecgibson: Sorry stupid copy-paste error.

test/govuk_component/taxonomy_sidebar_test.rb
alecgibson: Will do

alecgibson:
@mobaig changes pushed - can you please re-review?

fofr:
Are those list link styles intentional?

alecgibson:
@fofr you mean the bullets out to the left? Those are.

fofr:
I don't know how the front-end community should review this and other recent fronted PRs.

We're having trouble knowing:
* What to treat as production ready
* What is something purely for prototyping, user research, integration only or AB testing on live
* What users will see these changes
* How to test/view the PR itself (ie in this example, the component being used in context)
* How to consider these changes inline with our other patterns
* How to review the design elements
* How people in the future will know to remove code that's not live

Neither the PR or the story give this information.

tijmenb:
As discussed offline, agree it's probably hard to follow the process from outside of the team. We'll have a think and discuss tomorrow IRL!

PR #921, opened 2017-02-28
Add a default value for the `user-journey-stage` dimension

suzannehamilton:
Probably easiest to review as individual commits:

- Refactor test of analytics dimensions.
- Add a default value for the `user-journey-stage` dimension. Pages which show navigation rather than content will add the meta tag for this dimension. All other pages should be tracked as 'thing', which identifies pages as containing content rather than navigation.

Marked as DO NOT MERGE until the meta tag has been added to all the navigation page types. Full list in Trello: https://trello.com/c/SCsj2A9D/425-add-custom-dimension-to-page-view-tracking-on-new-navigation-pages

### DIFF
--- a/app/assets/javascripts/analytics/static-analytics.js
+++ b/app/assets/javascripts/analytics/static-analytics.js
@@ -93,6 +93,7 @@
       dimensions[key] = value;
     });
 
+    this.setUserJourneyStage(dimensions['user-journey-stage']);
     this.setSectionDimension(dimensions['section']);
     this.setFormatDimension(dimensions['format']);
     this.setResultCountDimension(dimensions['search-result-count']);
@@ -102,7 +103,6 @@
     this.setWorldLocationsDimension(dimensions['analytics:world-locations']);
     this.setRenderingApplicationDimension(dimensions['rendering-application']);
     this.setNavigationPageTypeDimension(dimensions['navigation-page-type']);
-    this.setUserJourneyStage(dimensions['user-journey-stage']);
   };
 
   StaticAnalytics.prototype.setAbTestDimensionsFromMetaTags = function() {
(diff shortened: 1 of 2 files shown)


suzannehamilton:
All the high-priority pages have been updated to include the `navigation` version of this meta tag, so this PR is ready go to.

app/assets/javascripts/analytics/static-analytics.js
```diff
@@ -188,7 +188,12 @@
   };
 
   StaticAnalytics.prototype.setUserJourneyStage = function(pageType) {
-    this.setDimension(33, pageType);
+    // Track the stage of a user's journey through GOV.UK. If the page does not
+    // set a page type in a meta tag, default to "thing", which identifes a page
+    // as containing content rather than some kind of navigation.
+    var defaultUserJourneyStageValue = "thing";
+    var dimensionValue = (typeof pageType === "undefined") ? defaultUserJourneyStageValue : pageType;
```
alecgibson: Is there any reason in particular you're checking specifically for `undefined` rather than using `!pageType` ? Or do you specifically want to allow eg `pageType = null` or `pageType = ''` (and if so, what do these mean?)

alecgibson: Also can you please rename `pageType` to `journeyStage` (or similar)? This looks like a copy-paste from dimension 32, which actually is page type?

alecgibson: And if you do just use the falsiness, remember you can write:

```javascript
var dimensionValue = journeyStage || 'thing';
```

suzannehamilton: > Is there any reason in particular you're checking specifically for `undefined` rather than using `!pageType ?`

It was to be consistent with line 135, which checks for the presence of dimension values in general. But on reflection, it probably makes sense to just check for falsiness here because we know what kind of values to expect (`finding` or `thing` or some other non-empty value). Whereas line 135 can't make that kind of assumption - there might be a dimension for which an empty string is meaningful. I've updated it, thanks.

> Also can you please rename `pageType` to `journeyStage` (or similar)?

Done. I left it as `pageType` because it is a type of page, but if it looks like a mistake then it's worth changing.

PR #928, opened 2017-03-06
Design updates to taxonomy sidebar component

mobaig:
Apply the latest set of design changes.

## Government frontend

screen shot 2017-03-06 at 16 40 18

## Smart Answers

screen shot 2017-03-07 at 15 02 17

https://trello.com/c/yFhLdgMX/471-style-and-copy-changes-to-the-taxonomy-side-bar

### DIFF
--- a/app/views/govuk_component/taxonomy_sidebar.raw.html.erb
+++ b/app/views/govuk_component/taxonomy_sidebar.raw.html.erb
@@ -2,19 +2,21 @@
   <aside class='govuk-taxonomy-sidebar' data-module='track-click' role='complementary'>
     <% items.each_with_index do |item, item_index| %>
 
-      <div class="sidebar-taxon">
-        <h2>More about <%= item[:title] %></h2>
+      <div class='sidebar-taxon'>
+        <h2>
+          <%= link_to item[:title], item[:url], class: 'taxon-link' %>
+        </h2>
 
-        <p>
+        <p class='taxon-description'>
           <%= item[:description] %>
         </p>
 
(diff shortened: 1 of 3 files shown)


app/views/govuk_component/taxonomy_sidebar.raw.html.erb
```diff
@@ -1,20 +1,36 @@
 <% if local_assigns[:items] && !items.blank? %>
   <aside class='govuk-taxonomy-sidebar' data-module='track-click' role='complementary'>
-    <% items.each_with_index do |item, item_index| %>
+    <% items.sort_by { |item| item[:title] }.each_with_index do |item, item_index| %>
```
carvil: What do you think of moving this logic to `govuk_navigation_helpers` instead? We already do some sorting in there for breadcrumbs, which means the govuk components become dummy.

We could, for example, move the sorting from this line: https://github.com/alphagov/govuk_navigation_helpers/blob/master/lib/govuk_navigation_helpers/content_item.rb#L23 into this line: https://github.com/alphagov/govuk_navigation_helpers/blob/master/lib/govuk_navigation_helpers/content_item.rb#L34

This way, we wouldn't need to sort data here.

mobaig: That makes much more sense actually. Will move this behaviour and its tests over.

carvil:
As discussed, we will sort the topics in `govuk_navigation_helpers`, which means this PR is ready to go.

PR #929, opened 2017-03-06
Change link text in banners to be more accessible

selfthinker:
When reading out of context, for example when using a screen reader, "find out what this means" is not very clear.

### DIFF
--- a/app/views/govuk_component/alpha_label.raw.html.erb
+++ b/app/views/govuk_component/alpha_label.raw.html.erb
@@ -5,7 +5,7 @@
       <% if local_assigns.include?(:message) %>
         <%= raw message %>
       <% else %>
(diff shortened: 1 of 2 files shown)


edwardhorsford:
I'm not convinced this is an improvement.

It's introducing language (alpha) that people are unlikely to know. It makes the link more specific, but also more complex. Would like a content designer to review.

selfthinker:
It doesn't introduce anything new, it already says "alpha" and the word "this" is exactly what is meant there.
Initially the whole sentence read:
> **Alpha** This part of GOV.UK is being built - (link)find out what this means(/link)

nickcolley:
This looks good to me 👍

selfthinker:
I wasn't sure if there will still be feedback from the content team. But as there doesn't seem to be any, I will merge now.

nickcolley:
Thanks @selfthinker :)

PR #935, opened 2017-03-07
Set default navigation page type tracking dimension

alecgibson:
This change closely follows https://github.com/alphagov/static/pull/921

Here, we set a default value of `none` for the `navigation-page-type` custom dimension. This is to ease the processing of Google Analytics results, since fields missing this custom dimension are omitted in some reports.

### Trello

https://trello.com/c/Eui3OmTB/491-track-navigation-page-type-for-non-taxon-pages

### DIFF
--- a/app/assets/javascripts/analytics/static-analytics.js
+++ b/app/assets/javascripts/analytics/static-analytics.js
@@ -93,7 +93,17 @@
       dimensions[key] = value;
     });
 
+    // Set defaults first, so that we get a stable ordering in tests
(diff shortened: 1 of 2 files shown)


suzannehamilton:
This might be more of a question for Tara, but should `none` be set on non-navigation pages? Or should it only be set on non-taxon navigation pages like browse and topic pages?

alecgibson:
@suzannehamilton I already checked with Tara - she wants "full coverage" (including 'thing's)

PR #936, opened 2017-03-08
Fixed some inconsistencies in the core.js script.

1DC:
I simply fixed some inconsistencies in the core.js script.

### DIFF
--- a/app/assets/javascripts/core.js
+++ b/app/assets/javascripts/core.js
@@ -2,22 +2,24 @@ $(document).ready(function() {
   $('.print-link a').attr('target', '_blank');
 
   var $searchFocus = $('.js-search-focus');
-  $searchFocus.each(function(i, el){
-    if($(el).val() !== ''){
+  $searchFocus.each(function(i, el) {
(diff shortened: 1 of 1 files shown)


cbaines:
Thanks for this @1DC . I've triggered the CI process to run, so hopefully that will pass.

Would you consider tweaking the commit message slightly? Putting something like what you have put in the Pull Request description in the commit message, e.g. "Fix stylistic inconsistencies in core.js script" would be much more useful than  "Consistency".

1DC:
@cbaines will do when I get home, can't do that through GitHub's UI.

nickcolley:
Hi :)

Thanks for this, I think we should merge this, we're moving towards linting using http://standardjs.com/ but that's a big ask for you to update this to use that right now so thank you for getting us closer!

nickcolley:
I've updated the commit message for you @1DC over at https://github.com/alphagov/static/pull/997

1DC:
Woops! Forgot about this. Thanks @nickcolley

PR #937, opened 2017-03-08
Add a taxonomy survey notification ID to surveys.js

mobaig:
If we're showing the new taxonomy navigation survey, we don't want to
show other banners.


mobaig:
As suggested by @suzannehamilton, it's probably sensible to introduce our own ID for the new navigation survey. It's clearer wherever we use this in frontend apps, and allows us to avoid various hacks to reverse the styling applied to the other banner IDs.

PR #938, opened 2017-03-08
Taxon names should be underlined in sidebar

carvil:
Trello: https://trello.com/c/koyo5MKj/48-topic-titles-in-sidebar-should-be-underlined

@alextea is this what you expect?

screen shot 2017-03-08 at 17 43 21


alextea:
👍

PR #941, opened 2017-03-09
Workaround for IE8 bug on related link bullet points

mobaig:
Fixes an issue where pseudo elements on an anchor tag receive the same
text decoration as the anchor tag.

## IE 8 (before)
 
screen_shot_2017-03-09_at_15_46_23

## IE 8 (after)

screen shot 2017-03-09 at 13 35 22

## Chrome (no change)

screen shot 2017-03-09 at 13 35 41

### DIFF
--- a/app/assets/stylesheets/govuk-component/_taxonomy-sidebar.scss
+++ b/app/assets/stylesheets/govuk-component/_taxonomy-sidebar.scss
@@ -22,8 +22,14 @@
       list-style-type: none;
       line-height: 1.5;
 
+      // We use a pseudo element to add bullet points to related links. This
+      // makes the bullets clickable and highlights them on link hover. With
+      // this approach, IE8 erroneously applies the anchor tag's underline to
+      // the pseudo element. Specifying separate text-decoration rules for the
+      // anchor tag and a nested span fixes this.
       a {
         display: block;
+        text-decoration: none;
 
         &:before {
           content: "•";
@@ -32,7 +38,10 @@
           margin-left: -15px;
         }
       }
+
+      a span {
+        text-decoration: underline;
+      }
     }
   }
-
 }
--- a/app/views/govuk_component/taxonomy_sidebar.raw.html.erb
+++ b/app/views/govuk_component/taxonomy_sidebar.raw.html.erb
@@ -32,7 +32,6 @@
                 <li>
                   <%=
                     link_to(
-                      related_item[:title],
                       related_item[:link],
                       data: {
                           track_category: 'relatedLinkClicked',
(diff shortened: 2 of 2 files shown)


alextea:
Did you try using the [ie(8)](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/docs/mixins.md#ie) mixin and setting `text-decoration: none` on the pseudo element?

36degrees:
Not convinced adding more markup is the solution here, but don't have the context – is there a reason the bullets being added to the links and not the list items, where you'd usually expect to find them?

36degrees:
Another issue with the current implementation – Voiceover (and possibly other assistive tech) will read the bullet as part of the link. So when tabbing through the links on the page it'll read 'Bullet. National curriculum in England: primary curriculum'.

Using a pseudo element on the list item would prevent that but might still cause similar problems. Ideally though you would just use a normal bulleted list, and no pseudo elements at all?

mobaig:
@36degrees 
>is there a reason the bullets being added to the links and not the list items, where you'd usually expect to find them?

Explained here: https://github.com/alphagov/static/pull/941/files#diff-b2af3eb92cd75ba5d8c0d1ab14ce7424R25. I'm not aware of a way to achieve the same effect with bullets on the list item.

@alextea 
Unfortunately that won't work in this instance. The problem is that IE8 fumbles the CSS spec and ignores `text-decoration` rules on the pseudo element.

mobaig:
@36degrees given these pages are in beta, I'd suggest we proceed with the current implementation. I've made a note of the issue here https://trello.com/c/37jp8eJQ/514-taxonomy-sidebar-related-links-bullet-points-can-we-make-these-better-for-screenreaders. Alex'll take another look at this once he's back from leave the week after next.

36degrees:
It'd be good to document the _reason_ behind '[making] the bullets clickable and [highlighting] them on link hover' because that seems like an odd thing to do… was this something that came out of user research?

fofr:
I agree with @36degrees – making bullet points clickable seems like unnecessary complexity and web inconsistent behaviour.

nickcolley:
We have anchors in lists all over GOV.UK and we don't reimplement bullets.

carvil:
These are all fair points and none of these decisions are final. We have enabled the A/B test today and are keen to keep running experiments on it to see how it can be improved.

We have a [trello](https://trello.com/c/37jp8eJQ/514-taxonomy-sidebar-related-links-bullet-points-can-we-make-these-better-for-screenreaders) card to revisit this, and if people think this isn't a good way forward we can remove it before the A/B test ends. 

Please note that this is only shown inside the new navigation pages to a small percentage of users. The sidebar is a bit different already in terms of its layout, so there are quite a few things to consider before we enable it to 100% of users:

### Related Links

screen shot 2017-03-09 at 17 19 19

### Taxonomy sidebar

screen shot 2017-03-09 at 17 19 31

Please do keep giving your feedback on these PRs, because I think that's important and will help shape things as we go along.

PR #942, opened 2017-03-09
Add README tips for local static development

NeilvB:


nickcolley:
@tijmenb is there a way to disable the cache with an ENV flag?

PR #943, opened 2017-03-10
[WIP] Fix clickable related link bullet styling

alecgibson:
This fix is for a more robust CSS fix to achieve the following effect in the taxonomy sidebar:

- Listed links have clickable bullet points
- The bullet points are styled to match the colour of the link text (even on hover/focus/visited)
- The bullet points are not underlined

The previous fix used a pseudoelement, which is both:
- a bullet point anti-pattern
- broken on screenreaders (which try to read out the pseudoelement)

This has been tested in Chrome, Firefox and IE8/11.

That all said, I'm not 100% convinced that having bullets that follow the behaviour and styling of a link is particularly typical. My preferred fix for this would be to remove this altogether. However, if that's the design we're sticking with for now this Pull Request should at least achieve the design in a more typical and robust way.

### Trello

### DIFF
--- a/app/assets/stylesheets/govuk-component/_taxonomy-sidebar.scss
+++ b/app/assets/stylesheets/govuk-component/_taxonomy-sidebar.scss
@@ -19,28 +19,20 @@
     .related-content {
       margin: 0 0 0 15px;
       padding: 0;
-      list-style-type: none;
-      line-height: 1.5;
 
-      // We use a pseudo element to add bullet points to related links. This
-      // makes the bullets clickable and highlights them on link hover. With
-      // this approach, IE8 erroneously applies the anchor tag's underline to
-      // the pseudo element. Specifying separate text-decoration rules for the
-      // anchor tag and a nested span fixes this.
-      a {
-        display: block;
-        text-decoration: none;
-
-        &:before {
-          content: "•";
-          display: inline-block;
-          width: 15px;
-          margin-left: -15px;
-        }
+      li {
+        font-size: 12px;
       }
 
-      a span {
-        text-decoration: underline;
+      a {
+        display: block;
+        font-size: 16px;
+        line-height: 1.5;
+
+        span {
+          position: relative;
+          top: 1px;
+        }
       }
     }
   }


36degrees:
Whilst browsers are happy to render it, the only elements 'allowed' as direct descendants of lists are list items (`li`) – and trying to validate a reduced test case results in:

`Error: Element a not allowed as child of element ul in this context.`

I _guess_ if browsers render it OK, it's not the end of the world (?) but it seems really odd to go out of our way with pseudo elements or invalid HTML to introduce non conventional behaviour like this.

I'd really like to understand the drive for this change, but at the minute I would definitely agree with your closing paragraph and encourage removal.

alecgibson:
@36degrees good point about valid HTML - this further adds to my suspicion that this is highly atypical behaviour.

@alextea what was the user need driving this change?

alextea:
Can I see a screenshot of how this looks?

alecgibson:
@alextea this has examples of all three link states:

sidebar-clickable-bullets

alecgibson:
If this is behaviour that we definitely want, we should probably do something along the lines of:

```HTML
<ul>
  <li>
    <a href="..." class="no-underline">
      <span class="hidden-from-screenreaders">&#8226;</span>
      <span class="underline">
        My link text
      </span>
    </a>
  <li>
</ul>
```

My normal argument against this would be that adding spans to all the links is really icky, and shouldn't be the concern of the HTML markup. This is largely assuaged by the fact that the list is rendered in Ruby, so you only have to change it in one place.

_However_ if this were to be a pattern we wanted to adopt in other places, you suddenly need to copy across this pseudo-bullet markup elsewhere, too, which isn't very DRY.

(Also I haven't worked through the CSS yet, but something tells me that getting that bullet to position itself correctly is going to be non-trivial/fragile)

nickcolley:
Have you found users trying to click the bullet points in testing?

alecgibson:
@nickcolley that's what I'd like to know

alecgibson:
I've had input from @alextea, so to save confusion I'm closing the PR and opening another one.

PR #944, opened 2017-03-10
Update breadcrumb component to allow optional collapsed display

mobaig:
When viewing the new taxonomy navigation on mobile, we want to show the
parent taxon only rather than the full breadcrumb.

This commit allows a 'collapse_on_mobile' flag to be passed in by any
app rendering the component, changing the applied CSS accordingly.

# Examples

## Government frontend
screen shot 2017-03-10 at 16 37 05

## Collections
screen shot 2017-03-10 at 16 41 30

https://trello.com/c/Y7OirowB/472-mobile-version-of-breadcrumbs-should-only-show-previous-link

### DIFF
--- a/app/assets/stylesheets/govuk-component/_breadcrumbs.scss
+++ b/app/assets/stylesheets/govuk-component/_breadcrumbs.scss
@@ -15,4 +15,32 @@
     color: $secondary-text-colour;
     text-decoration: none;
   }
+
+  @include media-down(mobile) {
+    &.collapse-on-mobile li {
+      display: none;
+
+      &.parent-taxon {
+        background-image: none;
+        display: block;
+        margin-left: 0;
+        padding-left: 14px;
+        position: relative;
+
+        &:before {
+          border-bottom: 5px solid transparent;
+          border-right: 6px solid #0b0c0c;
+          border-top: 5px solid transparent;
+          content: '';
+          display: block;
+          height: 0;
+          left: 0;
+          margin-top: -6px;
+          position: absolute;
+          top: 50%;
+          width: 0;
+        }
+      }
+    }
+  }
 }
--- a/app/views/govuk_component/breadcrumbs.raw.html.erb
+++ b/app/views/govuk_component/breadcrumbs.raw.html.erb
@@ -1,18 +1,21 @@
 <%
   breadcrumbs ||= []
+  collapse_on_mobile ||= false
 %>
-<div class="govuk-breadcrumbs" data-module="track-click">
+
+<div class="govuk-breadcrumbs <%= "collapse-on-mobile" if collapse_on_mobile %>" data-module="track-click">
   <ol>
   <% breadcrumbs.each_with_index do |crumb, index| %>
-    <li>
-      <%
-        is_link = crumb[:url].present? || crumb[:is_current_page]
-        path = crumb[:is_current_page] ? '#content' : crumb[:url]
-        aria_current = crumb[:is_current_page] ? 'page' : 'false'
-        css_class = crumb[:is_current_page] ? 'breadcrumb-for-current-page' : ''
+    <%
+      is_link = crumb[:url].present? || crumb[:is_current_page]
+      is_parent = index == breadcrumbs.length - 2
+      path = crumb[:is_current_page] ? '#content' : crumb[:url]
+      aria_current = crumb[:is_current_page] ? 'page' : 'false'
+      css_class = crumb[:is_current_page] ? 'breadcrumb-for-current-page' : ''
+    %>
 
-        if is_link
-      %>
+    <li class='<%= "parent-taxon" if is_parent %>'>
+      <% if is_link %>
         <%= link_to(
           crumb[:title],
           path,


app/assets/stylesheets/govuk-component/_breadcrumbs.scss
```diff
@@ -15,4 +15,32 @@
...
+
+      &.parent-taxon {
+        background-image: none;
+        display: block;
+        margin-left: 0;
+        padding-left: 14px;
+        position: relative;
+
+        &:before {
+          border-bottom: 5px solid transparent;
```
alecgibson: What's the transparent border for?

app/assets/stylesheets/govuk-component/_breadcrumbs.scss
```diff
@@ -15,4 +15,32 @@
     color: $secondary-text-colour;
     text-decoration: none;
   }
+
+  @include media-down(mobile) {
+    &.collapse-on-mobile li {
+      display: none;
+
+      &.parent-taxon {
```
alecgibson: As a stylistic point, this is quite nested. I personally don't mind too much (especially since components shouldn't be concerned with CSS overriding), but I know some people can have strong opinions on this.

app/assets/stylesheets/govuk-component/_breadcrumbs.scss
```diff
@@ -15,4 +15,32 @@
...
+        padding-left: 14px;
+        position: relative;
+
+        &:before {
+          border-bottom: 5px solid transparent;
+          border-right: 6px solid #0b0c0c;
+          border-top: 5px solid transparent;
+          content: '';
+          display: block;
+          height: 0;
```
alecgibson: Height (and width) of 0 makes me nervous - why is this like this?

app/assets/stylesheets/govuk-component/_breadcrumbs.scss
```diff
@@ -15,4 +15,32 @@
...
+          left: 0;
+          margin-top: -6px;
```
alecgibson: Please check this in Firefox - we saw it misbehaving with the breadcrumb links when I introduced negative margin on those.

app/views/govuk_component/breadcrumbs.raw.html.erb
```diff
@@ -1,18 +1,21 @@
...
   <% breadcrumbs.each_with_index do |crumb, index| %>
-    <li>
-      <%
-        is_link = crumb[:url].present? || crumb[:is_current_page]
-        path = crumb[:is_current_page] ? '#content' : crumb[:url]
-        aria_current = crumb[:is_current_page] ? 'page' : 'false'
-        css_class = crumb[:is_current_page] ? 'breadcrumb-for-current-page' : ''
+    <%
+      is_link = crumb[:url].present? || crumb[:is_current_page]
+      is_parent = index == breadcrumbs.length - 2
```
alecgibson: This won't work on any breadcrumbs which don't have the current page breadcrumb (which is currently optional).

app/assets/stylesheets/govuk-component/_breadcrumbs.scss
mobaig: Required to draw the back arrow in CSS - the prototype lifts the styles from here: http://govuk-elements.herokuapp.com/typography/#typography-links, I've just lifted them from the prototype.

app/assets/stylesheets/govuk-component/_breadcrumbs.scss
mobaig: As above, but doesn't look like these are necessary to render the arrow. Removing...

mobaig:
Working on email, closing this PR until someone else picks this change up.

app/assets/stylesheets/govuk-component/_breadcrumbs.scss
mobaig: Leaving as is unless there's a strong case for flattening.

alecgibson: Fine by me

app/assets/stylesheets/govuk-component/_breadcrumbs.scss
alecgibson: Hm. In that case I'd rather we import the [actual Sass](https://github.com/alphagov/govuk_elements/blob/c9af6fb107c86955fcb105d35ce3c5647fa870ab/public/sass/elements/_elements-typography.scss#L184-L231) rather than copy-paste.

suzannehamilton: I tried adding govuk-elements to the project, but it's not straightforward to include the arrow in the sass because govuk-elements provides the whole back link, including arrow, which doesn't match our styles here. There are also conflicts between the `@media` breakpoints, since we just want this to appear on mobile but the govuk-elements elements applies at all levels.

It might be nice to extract the arrow as a mixin and use it in both places, but I suggest doing it later once we've decided we're going to stick with this pattern. It's also going to take some thought (and discussion with the frontend teams) about how to add it to the library as a  reusable component, because it doesn't quite fit what they have already.

app/assets/stylesheets/govuk-component/_breadcrumbs.scss
suzannehamilton: I've checked this on FF on MacOS and it looks fine. I'll do some cross-browser testing once it's on integration.

app/views/govuk_component/breadcrumbs.raw.html.erb
suzannehamilton: I've updated this to get an `is_page_parent` from the breadcrumb itself. See #948 for my change.

app/assets/stylesheets/govuk-component/_breadcrumbs.scss
alecgibson: @fofr / @nickcolley who would be best to pass this action on to?

PR #945, opened 2017-03-10
Resize bullets

alecgibson:
This Pull Request fixes an earlier attempt to simply resize bullet points. The previous attempt to do this used pseudoelements, which:

- is not a very typical pattern
- does not play nice with screenreaders, which try to read out the bullets

### Screenshot

sidebar-resized-bullets

### Trello

### DIFF
--- a/app/assets/stylesheets/govuk-component/_taxonomy-sidebar.scss
+++ b/app/assets/stylesheets/govuk-component/_taxonomy-sidebar.scss
@@ -19,28 +19,25 @@
     .related-content {
       margin: 0 0 0 15px;
       padding: 0;
-      list-style-type: none;
-      line-height: 1.5;
 
-      // We use a pseudo element to add bullet points to related links. This
-      // makes the bullets clickable and highlights them on link hover. With
-      // this approach, IE8 erroneously applies the anchor tag's underline to
-      // the pseudo element. Specifying separate text-decoration rules for the
-      // anchor tag and a nested span fixes this.
-      a {
-        display: block;
-        text-decoration: none;
-
-        &:before {
-          content: "•";
-          display: inline-block;
-          width: 15px;
-          margin-left: -15px;
-        }
+      li {
+        font-size: 12px;
       }
 
-      a span {
-        text-decoration: underline;
+      a {
+        display: block;
+        font-size: 16px;
+        line-height: 1.5;
+
+        // For design reasons, we want the bullet points to be smaller than normal. This is achieved
+        // by giving the list a different font size to the actual link text.
+        // Since the bullet point has a different font size to the actual link text, the bullet's
+        // height is not vertically aligned correctly with the text. Having the text in this
+        // span element allows us to move the text relative to the bullet.
+        span {
+          position: relative;
+          top: 1px;
+        }
       }
     }
   }


alecgibson:
cc @alextea

mobaig:
Implementation looks fine but PR description is confusing.

> This Pull Request fixes an earlier attempt to simply resize bullet points. 

That doesn't really correlate with what's going on in that diff. Worth explaining the context behind any changes/removals for anyone needing to refer back to this.

app/assets/stylesheets/govuk-component/_taxonomy-sidebar.scss
```diff
@@ -19,29 +19,21 @@
...
-        text-decoration: none;
+        font-size: 16px;
+        line-height: 1.5;
 
-        &:before {
-          content: "•";
-          display: inline-block;
-          width: 15px;
-          margin-left: -15px;
+        span {
```
36degrees: Is the `<span>` still needed? Whats it doing now?

alecgibson: It's needed for vertical alignment of the bullet (which is slightly off because of its font size being different to that of the text) - I thought about it over the weekend and I think I want to slightly tweak this. Will change to WIP and add a comment in the CSS.

alecgibson: Actually @36degrees I tried to do this another way, but it didn't work so I've left it as-is. I've added a comment to the CSS, repeated here for convenience:

> Since the bullet point has a different font size to the actual link text, the bullet's height is not vertically aligned correctly with the text. Having the text in this span element allows us to move the text relative to the bullet.

I realise this isn't a pretty fix, so if you can think of a better way to do this, I'm all ears.

alecgibson:
@mobaig from talking to @alextea I gathered that the pseudoelement was never there to make the bullet clickable/highlighted; it was just to make it smaller than normal (please correct me if I'm wrong?).

But you're right in that that doesn't quite align with this PR. Here is some more context.

Given that we had the end goal of having smaller bullet points, we originally implemented this using pseudoelements. This wasn't a great solution for the following reasons:

- It's an anti-pattern. Lists are already capable of rendering bullets, so adding a pseudoelement is unnecessary and confusing
- Pseudoelements don't play nice with screen readers (which can try to read them out)
- The pseudoelements adopted the underline styling of the link in IE8, which required some of the additional fixes that I've removed in this PR

With this solution, we go back to using the list's built-in rendering, which should be less surprising in terms of implementation.

The one thing I'm not a fan of is having to force the bullets into vertical alignment with the text using `position: relative;`, so if anyone has any bright ideas on this front I'm all ears.

36degrees:
I _think_ removing `display: block` from the `.related-content a` and adding `vertical-align: middle` does the job and allows you to remove the span

alecgibson:
@36degrees as discussed, removing `display: block` will [make a non-clickable area in the middle of the link](https://trello.com/c/xshqCdju/33-it-s-possible-to-click-in-the-middle-of-related-links-and-not-go-anywhere), so annoyingly we can't use that solution.

nickcolley:
Nice work 👍 

>The one thing I'm not a fan of is having to force the bullets into vertical alignment with the text using position: relative;, so if anyone has any bright ideas on this front I'm all ears.

This does feels weird, I don't recall anything like that for our other bullet lists.

alecgibson:
@nickcolley the realignment is because the bullets are a different font size to the link text. Have you seen this done elsewhere?

fofr:
> Given that we had the end goal of having smaller bullet points
Why should they be smaller? Does the answer to that justify the complexity?

alecgibson:
@fofr it was a design decision from @alextea (who's unfortunately away for the week).

alecgibson:
I'm closing this for now until we can clarify with @alextea

alextea:
I think these look great.

@fofr originally we had the links in a list without bullets, using padding to delimit each item. However this used a lot of vertical space, especially when a content item is tagged to multiple topics. To solve that we are using a bulleted list without the padding, however the default bullet size in the font is quite large and visually heavy and was making the sidebar imbalanced. 

Reducing the size of the bullets means the sidebar has a more consistent visual weight.

alecgibson:
@alextea do you want to make a call on whether we bother vertically aligning the bullets or not? Here's an animation of aligned (text dropped) and not-aligned (text raised):

related-link-bullet-alignment

alecgibson:
I've had to reopen this PR separately because of Git force-push reasons: https://github.com/alphagov/static/pull/965

PR #946, opened 2017-03-10
Explicitly include button styles for the email survey submit button

h-lame:
The send button in the email survey UI used the ".button" class to get the
styling defined in `helpers/_buttons`.  Unfortunately this styling is not
included in all layouts that an app might request from static.  For example
it is included in the static layout, but not the header_and_footer_only or
core_layout.  This means that only some frontend rendering apps will style
the button correctly.

Instead we explicitly include the button styling mixin in the css for a
button in the user-satisfaction-survey in `helpers/_header`.  This helper
is included in all layouts so all frontend rendering apps will style the
button correctly.

For: https://trello.com/c/dfSufy85/114-develop-a-new-survey-delivery-method

### DIFF
--- a/app/assets/javascripts/surveys.js
+++ b/app/assets/javascripts/surveys.js
@@ -27,7 +27,7 @@
                             '      <input name="email_survey_signup[survey_source]" type="hidden" value="">' +
                             '      <input name="email_survey_signup[email_address]" type="text" placeholder="Your email address">' +
                             '      <div class="actions">' +
-                            '        <button class="button">Send</button>' +
+                            '        <button type="submit">Send</button>' +
                             '        <p class="button-info">We won\'t store your email address or share it with anyone</span>' +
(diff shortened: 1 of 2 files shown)


nickcolley:
@h-lame can you let me know the best way to test this? E.g. which pages I should check etc, will be helpful when it gets deployed too.

h-lame:
One example is that currently the button for email surveys is green on pages rendered by frontend (e.g. https://www.gov.uk/access-to-elected-office-fund: 

access to elected office fund - gov uk
).

but not on pages rendered by manuals-frontend (e.g. https://www.gov.uk/guidance/the-highway-code:

the highway code - guidance - gov uk

).

After merging the integration versions of the manuals-publisher page should render a green button.

Note that the survey is due to stop at midnight on Monday, so testing will be hard after that (not impossible, just hard).

PR #947, opened 2017-03-13
Add support for displaying a banner across all of GOV.UK

boffbowsh:
Allows for a partial to be deployed based on `app/views/root/_promo_banner_example.html.erb`. This should include the `global-bar` javascript module in order to enable view count tracking. The banner won't display alongside the survey banner or the emergency publishing banner.

`app/views/root/_promo_banner.html.erb` should remain empty if the banner isn't in use.

Example: 

screen shot 2017-03-13 at 16 55 49

### DIFF
--- a/app/assets/javascripts/modules/global-bar.js
+++ b/app/assets/javascripts/modules/global-bar.js
@@ -1,0 +1,57 @@
+/*
+  Global bar
+
+  Manages count of how many times a global bar has been seen
+  using cookies.
+*/
+(function(Modules) {
+  "use strict";
+
+  Modules.GlobalBar = function() {
+    this.start = function($el) {
+      var GLOBAL_BAR_SEEN_COOKIE = "global_bar_seen",
+          count = viewCount();
+
+      $el.on('click', '.dismiss', hide);
+
+      if ($el.is(':visible')) {
+        incrementViewCount(count);
+        track('Viewed');
+      }
+
+      function hide(evt) {
+        $el.hide();
+        GOVUK.setCookie(GLOBAL_BAR_SEEN_COOKIE, 999, {days: 84});
+        track('Manually dismissed');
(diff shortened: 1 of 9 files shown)


app/assets/javascripts/modules/global-bar.js
```diff
@@ -0,0 +1,57 @@
...
+        GOVUK.setCookie(GLOBAL_BAR_SEEN_COOKIE, 999, {days: 84});
+        track('Manually dismissed');
+        evt.preventDefault();
+      }
+
+      function incrementViewCount(count) {
+        count = count + 1;
+        GOVUK.setCookie(GLOBAL_BAR_SEEN_COOKIE, count, {days: 84});
+
+        if (count == 2) {
```
fofr: Should we auto-dismiss after 1 view? Is the banner informative rather than a call to action?

app/assets/javascripts/surveys.js
```diff
@@ -128,6 +128,7 @@
     init: function() {
       var activeSurvey = userSurveys.getActiveSurvey(userSurveys.defaultSurvey, userSurveys.smallSurveys);
       if (userSurveys.isSurveyToBeDisplayed(activeSurvey)) {
+        $('#global-bar').hide(); // Hide global bar if one is showing
```
fofr: We currently have some fairly aggressive surveys running (ie every 2 pageviews). We should avoiding running these types of survey during a banner.

boffbowsh: Yep, we've checked on that 👍

app/assets/javascripts/modules/global-bar.js
boffbowsh: This can be configured based on the need for the banner. Currently people are happy with 3 as a default

PR #948, opened 2017-03-14
Update breadcrumb component to allow optional collapsed display

suzannehamilton:
When viewing the new taxonomy navigation on mobile, we want to show the parent taxon only rather than the full breadcrumb.

This commit allows a 'collapse_on_mobile' flag to be passed in by any app rendering the component, changing the applied CSS accordingly.

https://trello.com/c/Y7OirowB/472-mobile-version-of-breadcrumbs-should-only-show-previous-link

This is a continuation of @mobaig's PR #944 - I can't reopen it because I force-pushed the branch first. TIL.

I've added my changes to Mo's original PR as a separate commit so you can see the diff. I'll squash before merging to master.

This will need a corresponding change in govuk_navigation_helpers to add the flag: https://github.com/alphagov/govuk_navigation_helpers/compare/master...breadcrumb-parents?expand=1

### DIFF
--- a/app/assets/stylesheets/govuk-component/_breadcrumbs.scss
+++ b/app/assets/stylesheets/govuk-component/_breadcrumbs.scss
@@ -15,4 +15,22 @@
     color: $secondary-text-colour;
     text-decoration: none;
   }
+
+  @include media-down(mobile) {
+    &.collapse-on-mobile li {
+      display: none;
+
+      &.parent-breadcrumb {
+        background-image: none;
+        display: block;
+        margin-left: 0;
+        padding-left: 14px;
+        position: relative;
+
+        &:before {
+          @include back-arrow;
+        }
+      }
+    }
+  }
 }
--- a/app/assets/stylesheets/govuk-component/_component.scss
+++ b/app/assets/stylesheets/govuk-component/_component.scss
@@ -4,6 +4,7 @@
 @import "design-patterns/alpha-beta";
 
 // Component mixins
+@import "mixins/back-arrow";
 @import "mixins/margins";
 @import "mixins/touch-friendly-links";
 
--- a/app/assets/stylesheets/govuk-component/mixins/_back-arrow.scss
+++ b/app/assets/stylesheets/govuk-component/mixins/_back-arrow.scss
@@ -1,0 +1,13 @@
+// An arrow to represent a "back" link
+
+@mixin back-arrow {
+  border-bottom: 5px solid transparent;
+  border-right: 6px solid;
+  border-top: 5px solid transparent;
+  content: '';
+  display: block;
+  left: 0;
+  margin-top: -6px;
+  position: absolute;
+  top: 50%;
+}
--- a/app/views/govuk_component/breadcrumbs.raw.html.erb
+++ b/app/views/govuk_component/breadcrumbs.raw.html.erb
@@ -1,18 +1,22 @@
 <%
   breadcrumbs ||= []
+  collapse_on_mobile ||= false
 %>
-<div class="govuk-breadcrumbs" data-module="track-click">
+
+<div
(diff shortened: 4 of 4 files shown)


app/assets/stylesheets/govuk-component/_breadcrumbs.scss
```diff
@@ -15,4 +17,30 @@
     color: $secondary-text-colour;
     text-decoration: none;
   }
+
+  @include media-down(mobile) {
+    &.collapse-on-mobile li {
+      display: none;
```
alecgibson: If no breadcrumbs are marked `is_page_parent` (ie if Navigation Helpers is out-of-date), then breadcrumbs won't show at all on mobile. Is that acceptable?

suzannehamilton: The `collapse-on-mobile` class should only be added if there is a parent breadcrumb:

```
<%= "collapse-on-mobile" if collapse_on_mobile && breadcrumbs.any? { |crumb| crumb[:is_page_parent] } %>"
```

app/assets/stylesheets/govuk-component/_breadcrumbs.scss
```diff
@@ -15,4 +17,30 @@
...
+      &.parent-breadcrumb {
+        background-image: none;
+        display: block;
+        margin-left: 0;
+        padding-left: 14px;
+        position: relative;
+
+        &:before {
+          border-bottom: 5px solid transparent;
+          border-right: 6px solid $back-link-arrow-color;
```
nickcolley: If this is the current text color can we remove this entirely and rely on the browsers `currentColor` behaviour?

https://developer.mozilla.org/en/docs/Web/CSS/color_value#currentcolor_keyword

This would look like `border-right: 6px solid;'

app/assets/stylesheets/govuk-component/_breadcrumbs.scss
```diff
@@ -15,4 +17,30 @@
...
+          border-top: 5px solid transparent;
+          content: '';
+          display: block;
+          left: 0;
+          margin-top: -6px;
+          position: absolute;
```
nickcolley: Feels like this should be possible without absolute positioning. Not sure without poking it but this would allow you to remove the magic positioning numbers.

alecgibson: (As mentioned on [the previous PR](https://github.com/alphagov/static/pull/944#discussion_r105689187), all this styling is from Govuk Elements, and should be commonised at some point)

app/assets/stylesheets/govuk-component/_breadcrumbs.scss
alecgibson: Right! Missed that.

app/assets/stylesheets/govuk-component/_breadcrumbs.scss
suzannehamilton: Good point, it is the same. I've removed the extra colour.

app/assets/stylesheets/govuk-component/_breadcrumbs.scss
suzannehamilton: I've extracted this arrow as a mixin within static as a step towards commonising it with govuk-elements.

suzannehamilton: @nickcolley What's your opinion on the current version of this?

suzannehamilton: I'm going to merge this now - let me know if you have any ideas for merging this with govuk_elements.

gemmaleigh:
GOV.UK elements uses the [breadcrumbs partial](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/stylesheets/design-patterns/_breadcrumbs.scss) from the GOV.UK frontend toolkit. 

If you would like to share any code with GOV.UK elements, the best place to put it is the GOV.UK frontend toolkit.

PR #949, opened 2017-03-14
Add new user research surveys

mybits:
For: https://trello.com/c/hgp4Rrkl

Add user research survey on 13 different pages for evaluations 2 campaigns: (1) Help Users Hold Government to Account and (2) Improve the platform basics (excluding Blogs and Campaigns); all inviting users to join the research panel

This PR also deletes old expired surveys (both url and email ones)

PR #950, opened 2017-03-15
Define surveys in html not js

h-lame:
For: https://trello.com/c/8YSfHXc2/110-reduce-need-to-change-static-js-when-deploying-surveys

Inspiration (if not any code) taken from @NeilvB's WIP here: https://github.com/alphagov/static/tree/survey-template-in-html which in turn was based on a conversation with @nickcolley.

The goal here is to define the surveys outside the JS bundle, so that we don't break the caching benefits for frequent visitors.  The approach here is to define the templates and surveys in `script` tags (using html for the template, and json for the surveys) in the HTML and to read the contents of those in the surveys JS implementation.

All of the dynamic parts of the definition have been moved into the survey implementation so that the survey definitions can be expressed in pure JSON, rather than the Javascript Object form they were in when they lived in the JS file.  This may or may not require more thought to check for security concerns.

This is best reviewed commit-by-commit I think as there's a lot of change involved in taking the dynamic parts out and making them into parameterised strings.  Particularly turning the "do what you want" `activeWhen: function () { ... }` into a simple object of parameters like `activeWhen: { path: ['government/publications'], organisations: ['<D10>'] }`.

### DIFF
--- a/app/assets/javascripts/surveys.js
+++ b/app/assets/javascripts/surveys.js
@@ -244,9 +244,7 @@
 
       $.each(smallSurveys, function(_index, survey) {
         if(userSurveys.currentTime() >= survey.startTime && userSurveys.currentTime() <= survey.endTime) {
-          if(typeof(survey.activeWhen) === 'function') {
-            if(survey.activeWhen()) { activeSurvey = survey; }
-          } else {
+          if (userSurveys.activeWhen(survey)) {
             activeSurvey = survey;
           }
         }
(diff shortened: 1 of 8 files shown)


nickcolley:
Wow, this is awesome! I'll put some time aside to properly review.

How're you testing this locally?

h-lame:
@nickcolley - so far, just running the tests ;).  Now I'm done I'll be running it locally and messing with the survey definitions to make them appear / not appear.  This PR needs a rebase before it can be merged because of #949, so if it's easier we can add more surveys and change the dates in a commit in this PR and then we'll get rid of them before merging.

app/views/surveys/_email_survey_template.html.erb
```diff
@@ -0,0 +1,39 @@
+<script type="text/html+template"
```
h-lame: Is this ok? Is there a better content-type I could use here that would make it more obvious?

app/views/surveys/_survey_definitions.html.erb
```diff
@@ -0,0 +1,54 @@
+<script type="application/json" data-survey-default>
+  {
+    "url": "https://www.surveymonkey.com/s/2MRDLTW?c={{currentPath}}",
+    "identifier": "user_satisfaction_survey",
+    "frequency": 50,
+    "surveyType": "url",
+  }
+</script>
```
h-lame: Quite wary of this, does a chunk of JSON on the page open us up to any potential security or injection problems?

h-lame: I thought about doing it via data attributes (like @NeilvB has in his [WIP version](https://github.com/alphagov/static/commit/bf516ac88a368d5baa1b08b3761f6e3bdedfcbd0#diff-3d28a36beb5925e7c6261c07ad15301bR39), but figured they'd be quite unwieldy with the number of options possible.

app/assets/javascripts/surveys.js
```diff
@@ -148,7 +81,27 @@
       return activeSurvey;
     },
 
+    getDefaultSurvey: function() {
+      try {
+        return JSON.parse($('#user-satisfaction-survey-container [data-survey-default]').text());
```
h-lame: Is `JSON.parse` safe to use across the browsers we expect to serve?  [Can I Use](http://caniuse.com/#feat=json) suggests it's widely supported, but I'm not sure what our browser support matrix is.

nickcolley:
We're tracking this over at: https://trello.com/c/FU2xWknu/7-3-define-surveys-in-html-not-js-950

Will try and get this moving ASAP.

nickcolley:
I've managed to test this before and after, and included the network recordings
[before-and-after-har.zip](https://github.com/alphagov/static/files/941991/before-and-after-har.zip)

Main things I notice are we're adding around 2kb to all pages with this change, from 7kb to 9kb.

I'm not sure this is the best approach since we're adding weight to every page now, I think we might be able to solve this by just splitting the survey code from the main bundle? This would solve the original problem while not adding weight to the original document.

elenatanasoiu:
Following @nickcolley's measurements, it looks like changing the surveys from JS to HTML actually adds 2kb to the page. In which case it's wiser to keep the surveys as they are.

The solution we will try next is to cache the survey JS separately so we will only have to expire and redownload a small part of the JS whenever we update surveys.

This PR will be closed.

PR #952, opened 2017-03-16
Allow the sidebar heading to not be a link

Davidslv:
We want to use this component with the "Elsewhere on GOV.UK" heading,
which doesn't have a link.

Trello:
https://trello.com/c/vM6Kaw9B/505-ias-and-content-designers-can-curate-related-links-for-any-content-type

### DIFF
--- a/app/views/govuk_component/taxonomy_sidebar.raw.html.erb
+++ b/app/views/govuk_component/taxonomy_sidebar.raw.html.erb
@@ -5,7 +5,8 @@
       <div class='sidebar-taxon'>
         <h2>
           <%=
-            link_to(
+            link_to_if(
+              item[:url],
               item[:title],
               item[:url],
               class: 'taxon-link',
--- a/test/govuk_component/taxonomy_sidebar_test.rb
+++ b/test/govuk_component/taxonomy_sidebar_test.rb
@@ -117,6 +117,7 @@ class TaxonomySidebarTestCase < ComponentTestCase
     total_sections = 2
     total_links_in_section_1 = 3
 
+    assert_select 'h2 a', "Item title"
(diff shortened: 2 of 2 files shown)


app/views/govuk_component/taxonomy_sidebar.raw.html.erb
```diff
@@ -4,20 +4,24 @@
...
-              data: {
-                track_category: 'relatedLinkClicked',
-                track_action: "#{item_index + 1}",
-                track_label: item[:url],
-                track_dimension: item[:title],
-                track_dimension_index: 29,
-              },
-            )
-          %>
+          <% if item[:url] %>
```
tijmenb: Maybe you could try http://apidock.com/rails/ActionView/Helpers/UrlHelper/link_to_if here.

carvil:
@Davidslv could you please add a screenshot of how this would look like together with other taxon related links?

Davidslv:
@carvil It will look like this screenshot, but bear in mind that we haven't built the functionality into navigation helpers yet so this isn't a screenshot of the end result.

screen shot 2017-03-16 at 16 10 39

app/views/govuk_component/taxonomy_sidebar.raw.html.erb
Davidslv: It looks less verbose, but does the job. will address.

Davidslv:
@alextea would you like to give us some input on this please?

alextea:
I'd like to understand more about when this would be used, with real examples of content.

alextea:
Spoken to Holly and I think this is good as it is for now. We will continue to look at the data to see how the links are used.

tijmenb:
Rebased and resolved conflicts.

PR #956, opened 2017-03-21
Add missing govspeak components to core layout

mcgoooo:
this adds the missing parts of govspeak to the core layout
used by government frontend, this is part of the work of
moving frontend answers to government frontend.

i have deliberately left out the documentation for the component guide for this.
things such as form-download just did not exist in government-frontend, and the rest if legacy.

We currently have no way of telling whether the legacy stuff is still used, and [we are in the process of defining a developing tool to analyse the (possibly) deprecated parts](https://trello.com/c/4HHInbRp/169-currently-have-no-way-of-properly-deprecating-govspeak-elements) 

this code is directly copied from frontend, and is the current css used to render (minus some deprecated  ie hacks)


nickcolley:
> i have deliberately left out the documentation for the component guide for this.

I think it'd be good to add these fixture like the other govspeak features so everything is complete?

mcgoooo:
@nickcolley i disagree, as this is the documentation that the content team is currently using, i do not wish to add elements that they may not wish to be using

mcgoooo:
a lot of it is essentially cruft, and without knowing what is what, i dont want to go adding things that people may start using, i also do not wish to remove, as things may break until we build tooling around it

nickcolley:
See @36degrees 's work for an example of how they should look: https://github.com/alphagov/static/commit/76ae1a7bdb62a7a0ae9af6f344c6cda3078c2d72

nickcolley:
Can we make these consistent with the other partials?

1. Scope to `.govuk-govspeak` to ensure this cant be reused outside of govspeak
2. Document the support for this feature is in the govspeak gem or whitehall

You can see an example of this in all the partials at the top of the file https://github.com/alphagov/static/commit/76ae1a7bdb62a7a0ae9af6f344c6cda3078c2d72#diff-ff72ff054f14a9ca284a7a17c99d5cfcR1

fofr:
The steps pattern `steps.scss` isn't included in `_govspeak.scss`

fofr:
The summary and place examples are incomplete. They don't demonstrate actual usage or the full styles that can be applied.

mcgoooo:
@fofr this was just updating the styles that are used, a lot of those styles have been deprecated according to the content teams guide, so this is why they are not included, the steps missing is a n oversight

fofr:
@mcgoooo You've included CSS styles for these features without providing a way to see how they render.

They should have complete examples, and the examples should indicate that they are being deprecated.

PR #965, opened 2017-03-27
Resize bullets

alecgibson:
This Pull Request fixes an earlier attempt to simply resize bullet points. The previous attempt to do this used pseudoelements, which:

- is not a very typical pattern
- does not play nice with screenreaders, which try to read out the bullets

Note that this PR replaces https://github.com/alphagov/static/pull/945

### Screenshot

sidebar-resized-bullets

### Trello

### DIFF
--- a/app/assets/stylesheets/govuk-component/_taxonomy-sidebar.scss
+++ b/app/assets/stylesheets/govuk-component/_taxonomy-sidebar.scss
@@ -12,31 +12,25 @@
     }
 
     .taxon-description {
-      font-size: 16px;
+      @include core-16;
       margin: 0;
     }
 
     .related-content {
+      @include core-16;
       margin: 0 0 0 15px;
       padding: 0;
 
       li {
-        font-size: 12px;
-      }
+        // Scale the bullets down in size using this variable, and
+        // then scale the text back up by the reciprocal
+        $bullet-fraction: 0.75;
+        font-size: #{$bullet-fraction}em;
 
-      a {
-        display: block;
-        font-size: 16px;
-        line-height: 1.5;
-
-        // For design reasons, we want the bullet points to be smaller than normal. This is achieved
-        // by giving the list a different font size to the actual link text.
-        // Since the bullet point has a different font size to the actual link text, the bullet's
-        // height is not vertically aligned correctly with the text. Having the text in this
-        // span element allows us to move the text relative to the bullet.
-        span {
-          position: relative;
-          top: 1px;
+        a {
+          display: block;
+          font-size: #{1 / $bullet-fraction}em;
+          line-height: 1.5;
         }
       }
     }
--- a/app/assets/stylesheets/govuk-component/_taxonomy-sidebar.scss
+++ b/app/assets/stylesheets/govuk-component/_taxonomy-sidebar.scss
@@ -19,28 +19,25 @@
     .related-content {
       margin: 0 0 0 15px;
       padding: 0;
-      list-style-type: none;
-      line-height: 1.5;
 
-      // We use a pseudo element to add bullet points to related links. This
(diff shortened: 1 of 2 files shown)


alecgibson:
@alextea do you want to make a call on whether we bother vertically aligning the bullets or not? The alignment adds a bit more complexity to the HTML/CSS, so if you don't think anyone will notice it's better to not bother aligning them. Here's an animation of aligned (text dropped) and not-aligned (text raised):

related-link-bullet-alignment

alecgibson:
(cc @36degrees @nickcolley @fofr @mobaig )

nickcolley:
@alecgibson what does it look like without any resizing shenanigans?

alecgibson:
@nickcolley the difference in bullet size is shown in this gif (large bullets being the default): 

bullet-sizes

(NB: I haven't readjusted vertical alignment in this gif to keep emphasis on bullet sizes)

alextea:
The vertical spacing is nice, but if it's considered too technically complex, I won't lose any sleep.

Great work, thanks @alecgibson 👍

app/assets/stylesheets/govuk-component/_taxonomy-sidebar.scss
```diff
@@ -19,29 +19,26 @@
...
-      // the pseudo element. Specifying separate text-decoration rules for the
-      // anchor tag and a nested span fixes this.
+      li {
+        font-size: 12px;
+      }
+
       a {
         display: block;
-        text-decoration: none;
+        font-size: 16px;
```
alextea: Should we be using responsive type sizes (ie `core-16`)?

alecgibson:
@alextea I've refactored the HTML/CSS to remove the extra `span` I was using for vertical alignment, and I've used `core-19` instead of setting `font-size: 16px`. Does this look alright now?

alextea:
`core-19` would make the text 19px not 16px as desired.

alecgibson:
@alextea right - I got confused because I was testing at mobile breakpoint. I've changed it to `core-16`, but note that this decreases the `font-size` for mobile compared to previously. Are you happy with that?

mobile-bullets-small-font

nickcolley:
@alecgibson have you tested this behaviour cross browser?

alecgibson:
@nickcolley I've checked Chrome, Firefox, Safari and IE8. The fix works across all browsers in that the bullets are smaller than their default, but the size of the bullets is slightly different in each browser. @alextea are you okay with that? The weirdest one is probably Safari on mobile:

screen shot 2017-03-30 at 09 34 11

nickcolley:
I really think we should do the simple thing here - keep the lists are they are across GOV.UK.

Alternatively if the balance of our normal bullet points look odd we have other approaches for example using en-dashes or em-dashes consistent with our contents list.

That said, I will not block this, so will leave it to your judgement. 👍

alextea:
@nickcolley we are changing the size of the bullets because of the position and context they're used in. The default bullets are too large and give the sidebar an unbalanced visual weighting.

app/assets/stylesheets/govuk-component/_taxonomy-sidebar.scss
```diff
@@ -12,36 +12,27 @@
...
 
     .taxon-description {
-      font-size: 16px;
+      @include core-16;
       margin: 0;
     }
 
     .related-content {
+      @include core-16;
       margin: 0 0 0 15px;
```
alextea: Should we use `$gutter-half` here?

alecgibson: For the `margin` ?

alextea: Yeah

PR #967, opened 2017-03-27
Add linked domain tracking to Plan For Britain campaign

boffbowsh:
Adds linked tracking without creating a new tracker, as per
https://developers.google.com/analytics/devguides/collection/analyticsjs/linker. 
As this is a temporary measure, we’re calling `ga` directly
rather than adding the functionality to
`GOVUK.GoogleAnalyticsUniversalTracker` in `govuk_frontend_toolkit`.


boffbowsh:
Relies on #966 being merged

PR #969, opened 2017-03-28
Fix tracking of "More" link in mainstream related links

suzannehamilton:
Google Analytics dimension 28 is used to track the length of a list of links. Update this value for mainstream related links so that it tracks the full length of the list including the "More" links, and update the "More" link itself to track the link list length within a section rather than the number of sections.

This makes dimension 28 consistent with dimension 29 which tracks the position in the list and already includes the "More" link in that list.

https://trello.com/c/JffDJtJ5/500-track-the-number-of-items-in-accordions-breadcrumbs-and-related-links


tijmenb:
@suzannehamilton this needs to be rebased now.

PR #977, opened 2017-03-30
Track the navigation document type for all pages

suzannehamilton:
Add a default value for the navigation document type, so that all pages (which include the JS from static) track the page as 'other' rather than a specific navigation type like 'guidance'.

https://trello.com/c/Tkn0P6Fm/532-track-guidance-navigation-supertype-in-google-analytics

Marked as DO NOT MERGE until we've checked that this dimension is giving us the expected results for pages which explicitly add the `meta` tag for navigation document types.


suzannehamilton:
Vin has confirmed that the 'guidance' tracking is working as expected, so it's fine for us to merge this to add the default value.

PR #978, opened 2017-03-30
WIP: Make metadata scale to content

nickcolley:
WIP: I need to rebase the commits just getting this up in the meantime.

Allows keys to dictate the the offset of the values, removing the need for magic numbers.

I'm in two minds if we should have this markup since if you run it in a validator it'll say it's not right but it's something that will most likely become valid based on the new specs.

Thinking of just using a boring list.


fofr:
Closing for now.

PR #988, opened 2017-04-05
Add a custom GA dimension for content-id meta tag

chriscarter90:
Trello: https://trello.com/c/wCb2VvDw/132-custom-dimension-content-unique-id

> We need to have a canonical key for joining GA data to other sources of data about content. The URI is the only thing currently in GA and is not authoritative.

This PR ensures that the `content-id` meta tag is sent as dimension 4.

### DIFF
--- a/app/assets/javascripts/analytics/static-analytics.js
+++ b/app/assets/javascripts/analytics/static-analytics.js
@@ -103,6 +103,7 @@
     this.setNavigationPageTypeDimension(dimensions['navigation-page-type']);
     this.setUserJourneyStage(dimensions['user-journey-stage']);
     this.setNavigationDocumentTypeDimension(dimensions['navigation-document-type']);
+    this.setContentIdDimension(dimensions['content-id']);
   };
 
   StaticAnalytics.prototype.setDimensionsThatDoNotHaveDefaultValues = function(dimensions) {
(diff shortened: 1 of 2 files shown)


spec/javascripts/analytics/static-analytics-spec.js
```diff
@@ -153,6 +153,12 @@ describe("GOVUK.StaticAnalytics", function() {
           number: 34,
           defaultValue: 'other',
           setupArgumentsIndex: 8
+        },
+        {
+          name: 'content-id',
+          number: 4,
+          defaultValue: 'unknown',
```
boffbowsh: Could this be an empty UUID instead? Just in case we want to later always fit this value into a UUID field in a database. `00000000-0000-0000-0000-000000000000`

app/assets/javascripts/analytics/static-analytics.js
```diff
@@ -168,6 +169,10 @@
 
   StaticAnalytics.prototype.setThemesDimension = function(themes) {
     this.setDimension(3, themes || 'other');
+  }
+
+  StaticAnalytics.prototype.setContentIDDimension = function(contentID) {
```
boffbowsh: I think this should be `setContentIdDimension`. See this SO post on acronyms in camelcase http://stackoverflow.com/questions/15526107/acronyms-in-camelcase

chriscarter90:
Hi @boffbowsh - thanks for your comments, they have been addressed and this should be ready to go.

app/assets/javascripts/analytics/static-analytics.js
```diff
@@ -168,6 +169,10 @@
 
   StaticAnalytics.prototype.setThemesDimension = function(themes) {
     this.setDimension(3, themes || 'other');
+  }
+
+  StaticAnalytics.prototype.setContentIdDimension = function(contentId) {
+    this.setDimension(4, contentId || '00000000-0000-0000-0000-000000000000');
```
fofr: Do we need the empty default? ('00000000-0000-0000-0000-000000000000')

PR #995, opened 2017-04-11
Create rake tasks to add and remove emergency banner values in Redis

leenagupte:
Trello card: https://trello.com/c/zewXbizh
Follows on from PR #990 

## Motivation

The current emergency banner doesn't allow static to be deployed while the banner is up.
PR #990 retrieved values from Redis to determine:

* The banner should be displayed
* What the contents of the banner should be, i.e. the heading and banner colour etc.

This PR sets those values.

## Expected changes

Two rake tasks have been created, one to deploy the emergency banner, and the other to remove it.

### Deploying the Emergency Banner

Run task:
```
emergency_banner:deploy['{campaign_class}','{heading}','{short_description}','{link}']
```
Where
* `campaign_class` is the colour of the banner
* `heading` is the \<H1\> title
* `short_description` is the text that appears under the title
* `link` is the more information link

### Removing the Emergency Banner

Run task:
```
emergency_banner:remove
```


doc/emergency-banner.md
```diff
@@ -0,0 +1,19 @@
...
+See the [opsmanual](https://docs.publishing.service.gov.uk/manual/emergency-publishing.html#adding-emergency-publishing-banners) for information about what the Emergency Banner is and when it should be deployed.
+
+## Running the Emergency Banner locally
+
+### Deploying the Emergency banner
+
+```
+bundle exec rake emergency_banner:deploy['{campaign_class}','{heading}','{short_description}','{link}']
+```
+
```
bevanloon: Since `campaign_class` can only be `red`, `black` or `green` at the moment, perhaps we should mention that?

lib/emergency_banner/display.rb
```diff
@@ -0,0 +1,34 @@
+module EmergencyBanner
+  class Display
+    def initialize
+      @redis = Redis.new
```
bevanloon: If we are doing a refactor, we might as well clean this up a bit?
There's no real reason to expose `@redis` unless we want to inject it in when initialising the class.

If we don't care about that then we could hide it under `private`:

```
private
  def connect
    Redis.new
  end

  def content
    @data ||= connection.hgetall("emergency_banner")
    ...
  end
```

test/unit/tasks/emergency_banner_test.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  should 'pass all supplied parameters to the deploy job' do
+    EmergencyBanner::Deploy.any_instance.expects(:run).with('campaign class', 'heading', 'short description', 'link')
+
+    Rake::Task['emergency_banner:deploy'].invoke('campaign class', 'heading', 'short description', 'link')
+  end
+end
+
+describe 'emergency_banner:remove' do
+  should 'run the process to deploy the emergency banner' do
```
bevanloon: deploy -> remove

test/unit/tasks/emergency_banner_test.rb
```diff
@@ -0,0 +1,47 @@
+require 'test_helper'
+require_relative '../../../lib/emergency_banner/deploy'
+require_relative '../../../lib/emergency_banner/remove'
+
+describe 'emergency_banner:deploy' do
```
bevanloon: is there any benefit to keeping these here rather than in `test/unit/emergency_banner/deploy_test.rb` etc?

test/unit/emergency_banner/deploy_test.rb
```diff
@@ -0,0 +1,50 @@
...
+                                              :campaign_class, "black",
+                                              :heading, "A title",
+                                              :short_description, "A short description of the event",
+                                              :link, "https://www.gov.uk"
+                                              )
+
+      EmergencyBanner::Deploy.new.run("black", "A title", "A short description of the event", "https://www.gov.uk")
+    end
+  end
+end
```
bevanloon: Do we need a specific test for the case of keys already being set and then overwritten? Or is that going too far and we are trusting Redis to do the right thing?

doc/emergency-banner.md
nickcolley: Not sure if it's within the scope of your work but it'd be good to rename the classes so they're based semantic rather than based on their presentation.

e.g.

'red => alert'
'black => fatality'
'green => (not sure what this means)'

fofr: Green is a large scale localised emergency.
Red is a national emergency
Black is death of a notable person.

https://docs.publishing.service.gov.uk/manual/emergency-publishing.html#types-of-banners

lib/emergency_banner/display.rb
leenagupte: I agree. I'll clean this up and fixup the change.

doc/emergency-banner.md
bevanloon: @nickcolley perhaps that would come into this card https://trello.com/c/3pKoTsFM/49-review-the-frontend-code-and-markup ?

nickcolley: @bevanloon looks like it 👍

test/unit/tasks/emergency_banner_test.rb
leenagupte: The structure of the tests in rails should mimic the structure of the code, so really, I should move these tests into `test/tasks`.  I'll do that and fixup. Thanks!

test/unit/emergency_banner/deploy_test.rb
leenagupte: No, I think that that is a valid thing to test.

leenagupte:
@bevanloon I've replied to your comments, and fixed up the changes. Cheers!

PR #996, opened 2017-04-11
Track total number of sections and section links

alecgibson:
At the moment we only record related link and accordion link information
on the click event (ie only if someone uses the link). We have no
information about if links are not used, since they don’t get clicked
and then never send a tracking event.

This change will add tracking for custom dimensions on the page view,
which will report information on:

- the total number of “sections” on the page
- the total number of “section links” on the page

These values will have different meanings depending on the type of page
we are on.

On a content item, the number of sections will refer to the number of
sections of related links in the sidebar. The total number of links
is the total number of related links in the sidebar (not including
the heading in a taxon sidebar, but including the “More” link).

On a leaf page in the new taxon navigation, the number of sections is
not tracked, and the total number of links is the total number of
leaf links.

On an accordion page in the new taxon navigation, the number of
sections will refer to the number of accordion sections, and the total
number of links will refer to the total number of content items
under all the accordion sections.

On a grid page in the new taxon navigation, the number of sections is
the number of grid links, and the total number of links will refer to
the total number of grid links plus the leaf links.

### Dependencies:

- [x] https://github.com/alphagov/collections/pull/326

### Trello

https://trello.com/c/VS6HcoBS/556-track-the-total-number-of-links-and-sections-for-accordions-and-related-links

### DIFF
--- a/app/assets/javascripts/analytics/static-analytics.js
+++ b/app/assets/javascripts/analytics/static-analytics.js
@@ -108,6 +108,8 @@
     this.setTaxonIdDimension(dimensions['taxon-id']);
     this.setTaxonSlugsDimension(dimensions['taxon-slugs']);
     this.setTaxonIdsDimension(dimensions['taxon-ids']);
+    this.setTotalNumberOfSections();
+    this.setTotalNumberOfSectionLinks();
   };
 
   StaticAnalytics.prototype.setDimensionsThatDoNotHaveDefaultValues = function(dimensions) {
@@ -173,7 +175,7 @@
 
   StaticAnalytics.prototype.setThemesDimension = function(themes) {
     this.setDimension(3, themes || 'other');
-  }
+  };
 
   StaticAnalytics.prototype.setContentIdDimension = function(contentId) {
     this.setDimension(4, contentId || '00000000-0000-0000-0000-000000000000');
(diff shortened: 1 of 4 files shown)


alecgibson:
I would love thoughts on this approach. I realise that tracking using CSS selectors can be a little fragile (especially because the tests are based on mocked HTML). However, the only alternative I could think of was to try to put the responsibility of tracking into eg the `taxonomy_sidebar.raw.html.erb` template, and then inject tracking `meta` tags into the `<head>` element eg using `<% content_for :head do %>`.

However, the problems with this approach were:
- In `government-frontend` I couldn't get the `content_for :head` to work (but `content_for :extra_head_content` works?)
- Editing the `head` from a magic component feels icky

app/assets/javascripts/analytics/static-analytics.js
```diff
@@ -207,6 +209,26 @@
     this.setDimension(21, position);
   };
 
+  StaticAnalytics.prototype.setTotalNumberOfSections = function() {
+    var sidebarSections = $('.govuk-related-items h2').length;
```
nickcolley: Agreed these selectors know too much about the internals of the component, and will break without tests.

Not thought it through but maybe we can put some data attributes in the component?

alecgibson: What do you mean by data attributes?

alecgibson: Do you mean do something similar to here: https://github.com/alphagov/static/pull/996/files#diff-a15d624c56e33899934804a2828f0d7fR222 ?

nickcolley: Yeah, this would allow us to keep those attributes consistent without blocking changes to presentation (CSS classes) or semantics (h2)

alecgibson: @nickcolley I've changed this to use special data attributes.

nickcolley: Small nit: having `a[]` seems like that may be overqualified for a selector? Do we need the anchors tags in the selector at all?

alecgibson: If you feel strongly I can remove them, but given that we're actively looking for links, I think the `a` is justified and descriptive.

alecgibson:
@nickcolley are you happy with these changes? Can you please approve if so?

app/views/govuk_component/related_items.raw.html.erb
```diff
@@ -1,7 +1,7 @@
 <% if local_assigns[:sections] && !sections.blank? %>
   <aside class="govuk-related-items" data-module="track-click" role="complementary">
     <% sections.each_with_index do |section, section_index| %>
-      <h2
+      <h2 data-track-count="sidebarRelatedItemSection"
```
nickcolley: Maybe you've considered this and disregarded it so sorry, but instead of the JavaScript checking the amount of elements on the page could we potentially do something like.

```erb
<aside
  class="govuk-related-items"
  data-module="track-click"
  data-track-total-sections="<%= sections.length %>" 
  role="complementary"
>
```

One con of this is it could drop out of sync with what users see but interested to hear your thoughts since it would make your logic simpler I think.

nickcolley:
@alecgibson approved with a non blocking comment.

app/views/govuk_component/related_items.raw.html.erb
alecgibson: @nickcolley I'd originally thought of doing something like this, but I was put off by exactly your observation: it could drift out of sync with what users see, which is exactly what we're trying to track here and therefore is of the utmost importance.

For example, in an early draft of this change, I overlooked the following line buried in the taxonomy sidebar: https://github.com/alphagov/static/blob/master/app/views/govuk_component/taxonomy_sidebar.raw.html.erb#L31

This sort of logic is already in our codebase, creating differences between `count` and what the user sees, and I think it's much safer to select for elements that we know we want to track, rather than assume a count from the content item.

Given that Vin is quite keen to start tracking this all as soon as possible, I'm going to go ahead and merge, but if you do disagree with this approach, please let me know and we can make any necessary changes.

PR #997, opened 2017-04-11
Fix stylistic inconsistencies in core.js script

nickcolley:
See: https://github.com/alphagov/static/pull/936

New branch since I can't push to @1DC 's branch, will keep @1DC 's contribution as they're still the author 😄


nickcolley:
Okay nevermind, looks like @fofr got the commit in as is, no worries 👍

PR #1000, opened 2017-04-11
Change how the Emergency Banner is deployed

leenagupte:
Trello cards:
* https://trello.com/c/2D0KFQZP
* https://trello.com/c/EPTEOgHJ
* https://trello.com/c/EPTEOgHJ
* https://trello.com/c/3eq0AeHA

Related PRs:
* https://github.com/alphagov/govuk-puppet/pull/5734
* https://github.com/alphagov/govuk-puppet/pull/5760
* https://github.com/alphagov/frontend/pull/1208

## Context

The emergency banner implementation has been re-written so that instead of using a fabric script to add content to some partials in Frontend and Static, we will instead populate a hash in Redis with the values needed for the emergency banner.

This will mean that should we need to deploy an emergency banner, Static and Frontend deploys are no longer blocked.

We have chosen to store the values in Redis rather than another datastore because all applications can access the same Redis environment, so Frontend and Static don't need to keep separate stores for the Emergency Banner values, they can share the same ones.

## The Implementation

* Add the Redis gem to both Frontend and Static
* Static retrieves a `emergency_banner` hash from Redis and uses it to determine whether to display the emergency banner
* Create two emergency banner partial templates, one for the homepage and the other for all other pages. Populate them with the contents of the `emergency_banner` hash from Redis.
* Create two tasks, one to set values in the `emergency_banner` hash in Redis and the other to delete the hash

## Future work

Future work will remove the current implementation of the emergency banner code which relies on empty partials that are populated by a fabric task.


app/views/root/_base.html.erb
```diff
@@ -30,11 +32,21 @@
   <% if @banner_notification.present? %>
     <%= @banner_notification %>
   <% end %>
+
+  <% if @emergency_banner %>
```
nickcolley: What's the difference between the homepage banner and the default banner?

leenagupte: @nickcolley Just the styling (or the lack of it on the default banner). If the emergency is ever restyled so that it looks identical on all pages, including the homepage, this `is_homepage` check can be removed.

nickcolley: 👍  good to know thanks @leenagupte

lib/emergency_banner/display.rb
```diff
@@ -0,0 +1,34 @@
...
+    end
+
+  private
+
+    def content
+      @data ||= connect.hgetall("emergency_banner")
+      @data.symbolize_keys if @data
+    end
+
+    def connect
```
boffbowsh: For timeouts, I'd suggest something like 0.1. Redis in production looks like it responds to most queries within 1ms but we should give a bit of leeway for network latency.

app/views/notifications/_emergency_banner.erb
```diff
@@ -0,0 +1,9 @@
+<section id="emergency-banner-notification" class="<%= banner.campaign_class %>">
```
leenagupte: Thanks. I've renamed the partials and fixed up.

lib/emergency_banner/display.rb
leenagupte: Thanks. I'll give this ago and fix up the commit.

lib/emergency_banner/display.rb
```diff
@@ -0,0 +1,34 @@
...
+    end
+
+    def link
+      content[:link] if content[:link].present?
+    end
+
+  private
+
+    def content
+      @data ||= connect.hgetall("emergency_banner")
```
leenagupte: @boffbowsh I'm not sure if we need this. `campaign_class` and `heading` must both be defined for the banner to show. So if one fails, and the other succeeds, the user still won't see the banner.

It's also possible and ok for the `emergency_banner` key to not exist (return nil) in Redis. The `emergency_banner` key will be nil in an environment until the first time it is set from the rake task. As the banner is only going to be "up" in rare occasions I think it's ok not to notify Errbit that the emergency banner content could not be found.

boffbowsh: Yeah so if the key doesn't exist and it returns nil, you need to memoize that `nil` value so that it doesn't try and request it four times per page view. If `@content` ends up being `nil` due to `hgetall` returning `nil`, the `||=` will mean it tried to fetch it again on the next call. Using `if defined?` means that it checks if the `@content` variable has been defined at all, regardless of whether it's `nil` or not.

leenagupte: Thanks. That makes a lot of sense. Will apply and fix up.

lib/emergency_banner/display.rb
```diff
@@ -0,0 +1,45 @@
+module EmergencyBanner
+  class Display
+    class << self
+      def client
+        @client ||= Redis.new(timeout: 0.1)
```
mcgoooo: what happens if redis is unavailable?

mcgoooo:
Can i ask what alternative approaches to this where thought of for this?

lib/emergency_banner/display.rb
boffbowsh: It will fallback to not displaying anything.

boffbowsh:
We looked at using ESI in Fastly but it doesn't support including from outside our infrastructure, which would have been the main benefit. There will be a writeup after the blitz of the other options we discussed.

PR #1001, opened 2017-04-18
[Do not merge] Redesign emergency banner

alextea:
Trello card: https://trello.com/c/eSfTdFEx

This pull request consolidates the emergency banners into a single partial, improves the accessibility of the markup and improves the design of the banners.

It also uses semantic classnames rather than colours (‘red’, ‘green’ and ‘black’).

The changes include:

* single partial with same markup for both homepage and content variants
* use `role="banner"` for improved accessibility
* use heading
* move link to left hand side to improve readability and layout
* use 2/3 column width for text
* improve type size and spacing to make banners legible
* use same colour for local and national emergency, to simplify design

You can view [a prototype of how the banners will look](https://govuk-emergency-banners.herokuapp.com/).

There's probably other code that will need changing before this can be merged (especially the tests). I haven't checked this running in a VM.

### Homepage
**Desktop:**
home-local-desktop

**Tablet:**
home-national-tablet

**Mobile:**
home-death-mobile

### Content page

**Desktop:**
content-national-desktop

**Tablet:**
content-death-tablet

**Mobile:**
content-local-mobile


app/assets/stylesheets/helpers/_emergency-banner-notifications.scss
```diff
@@ -1,113 +1,67 @@
...
-    padding: 0;
-    margin: 0;
+    @include core-19;
+    color: $white;
+    margin-top: 0;
+    margin-bottom: $gutter-two-thirds;
 
     @include media(tablet) {
-      padding: 0 30px 0 30px;
+      width: percentage(2/3);
```
nickcolley: Rather than using percentages widths under breakpoints you could consider using `max-width` with `em`s.

Since here you're trying to limit line length for legibility if you did something like the following it had a direct relationship with the amount of characters:

We did something similar on short-text descriptions https://github.com/alphagov/government-frontend/pull/323

app/views/notifications/_emergency_banner.html.erb
```diff
@@ -1,9 +1,9 @@
-<section id="emergency-banner-notification" class="<%= banner.campaign_class %>">
+<section id="emergency-banner-notification" class="<%= banner.campaign_class %>" role="banner">
```
nickcolley: I think we could take this opportunity to remove the use of the `id` and only use classes, this makes it more consistent with [GOV.UK Publishing Components](http://govuk-component-guide.herokuapp.com/) which I'd think we probably would want to make a 'banner' component out of this work at some point. (Non-blocking comment for this though)

Related: https://csswizardry.com/2011/09/when-using-ids-can-be-a-pain-in-the-class/

nickcolley:
These are looking great, I added a few comments 👍

app/views/notifications/_emergency_banner.html.erb
```diff
@@ -1,9 +1,9 @@
...
   <div>
-    <p><%= banner.heading %><br>
-    <%= banner.short_description %></p>
+    <h1><%= banner.heading %></h1>
+    <p><%= banner.short_description %></p>
     <% if banner.link %>
-      <a href="<%= banner.link %>" class="more-information">More information</a>
+      <a href="<%= banner.link %>">More information</a>
```
alextea: In order to [enable custom link text](https://trello.com/c/ZPkk7hnk/51-add-new-field-for-link-text), some more work needs to be done. As the blitz is over, they're out of scope for this PR.

app/assets/stylesheets/helpers/_emergency-banner-notifications.scss
alextea: I'd like to keep the percentage for now as it means the text aligns with other elements in the grid.

bevanloon:
I think currently, on non-hompage banners, the three quarter length blue bar below the black top banner moves below the emergency banner(see the screenshot below). Are we intentionally moving this above the banner? In which case, I think we might need to edit the code in `_base.html.erb`

screen_shot_2017-04-05_at_15 30 54

app/views/notifications/_emergency_banner.html.erb
bevanloon: @nickcolley still keen to sit down and discuss this once we get back into the office.

app/views/notifications/_emergency_banner.html.erb
```diff
@@ -1,9 +1,9 @@
-<section id="emergency-banner-notification" class="<%= banner.campaign_class %>">
+<section id="emergency-banner-notification" class="<%= banner.campaign_class %>" role="banner">
   <div>
-    <p><%= banner.heading %><br>
-    <%= banner.short_description %></p>
+    <h1><%= banner.heading %></h1>
```
selfthinker: Ideally there should only be one h1 per page, and that should reflect what the page is about.
I recently tried to find proper evidence of the why and how but also had trouble finding anything substantial. (I'm currently investigating something similar.) I know some screen reader users would jump to the first h1 and not assume there is another one.

Semantically I would say the emergency banner heading doesn't make sense to be an h1, especially since it will be the same across multiple pages. What was the reason for it? I'd suggest using an h2.
(Although it's also controversial to use any heading before the h1. But I think giving structure is more important than being strict about the hierarchy.)

Sorry, that's not a very clear answer. The main problem is that HTML doesn't have a concept of structure and heading hierarchy outside of the main content. The W3C tried solving that with the infamous document outline but failed. So, although there might be a better answer, I don't think there will be a 100% answer any time soon.

app/views/notifications/_emergency_banner.html.erb
selfthinker: Why not just fix the empty p? (`unless short_description.blank? ...`)

app/views/notifications/_emergency_banner.html.erb
alextea: Do I need to make a change here @nickcolley?

app/views/notifications/_emergency_banner.html.erb
alextea: The reason is that I was following the markup in the original template. I will change it to an h2. I actually thought the HTML outline still stood and that a section would have it's own hierarchy.

alextea:
@bevanloon I think it looks better with the blue bar below the emergency banner. I will update the prototype to reflect those changes.

app/views/notifications/_emergency_banner.html.erb
nickcolley: cc @selfthinker thoughts?

alextea:
@nickcolley @selfthinker I have made some changes based on your feedback.

alextea:
Updated screenshots too

app/views/notifications/_emergency_banner.html.erb
selfthinker: The outline is still in the HTML5 spec but it's actually not interpreted by any browser. That's why there is a warning in the spec to say so: http://w3c.github.io/html/sections.html#creating-an-outline

app/views/notifications/_emergency_banner.html.erb
selfthinker: I think a section would make sense here, but am not sure myself. Regarding the banner role, I just found
> Within any document or application, the author SHOULD mark no more than one element with the banner role.
on https://www.w3.org/TR/wai-aria/roles#banner
But I don't have enough experience to know for sure what the effect would be in assistive technology.

nickcolley: For now, it would seem the best bet would be to remove the banner role if we already have one.

selfthinker: If we remove it, what do you think of bringing the section back? Or is there another role we could use instead? Maybe `region`?

selfthinker: FYI, `<section>` has the implicit aria role of `region`, so it should either be one or the other... or both if we want to cater for older screen readers, as we do with other elements, e.g. `header`.

app/assets/stylesheets/helpers/_govuk-emergency-banner.scss
```diff
@@ -62,6 +61,6 @@
   }
 
   &.local-emergency {
-    background-color: $red;
+    background-color: $turquoise;
```
leenagupte: I'd either split this out into it's own commit, or squash into to the first one as it doesn't fit the commit description of making the banner more like other GOV.UK components.

app/views/notifications/_emergency_banner.html.erb
```diff
@@ -1,7 +1,9 @@
 <div class="govuk-emergency-banner <%= banner.campaign_class %>" role="banner">
   <div>
     <h2><%= banner.heading %></h2>
-    <p><%= banner.short_description %></p>
+    <% unless banner.short_description.blank? %>
```
leenagupte: Would `if banner.short_description` be better? Also consistent with the check for `banner.link`.

test/integration/notifications_test.rb
```diff
@@ -31,7 +31,7 @@ class NotificationsTest < ActionDispatch::IntegrationTest
 
       should "render the short description and more information link" do
         EmergencyBanner::Display.any_instance.stubs(:heading).returns("Alas poor Yorick")
-        EmergencyBanner::Display.any_instance.stubs(:campaign_class).returns("black")
+        EmergencyBanner::Display.any_instance.stubs(:campaign_class).returns("notable-death")
```
leenagupte: Wasn't this updated in the 2nd commit?

app/assets/stylesheets/helpers/_govuk-emergency-banner.scss
alextea: 😿

fofr:
Should this be closed in favour of #1049?

bevanloon:
Closing in favour of #1049

PR #1002, opened 2017-04-19
Style address blocks with margins

fofr:
Addresses are used in the $A markdown pattern:
https://github.com/alphagov/govspeak/blob/a04ef6b54f090ec03911d0daaed4c74741578fea/lib/govspeak.rb#L249-L251

This pattern is used in specialist documents that include contacts:
https://www.gov.uk/drug-device-alerts/transwarmer-infant-transport-mattress-novaplus-transwarmer-infant-heat-therapy-mattress-with-warmgel-infant-heel-warmer-risk-of-serious-burn

| Before | After |
|--|--|
|screen shot 2017-04-19 at 14 40 08|screen shot 2017-04-19 at 14 45 42|

## Example

screen shot 2017-04-19 at 15 01 18

### DIFF
--- a/app/assets/stylesheets/govuk-component/govspeak/_contact.scss
+++ b/app/assets/stylesheets/govuk-component/govspeak/_contact.scss
@@ -6,18 +6,17 @@
 // - alphagov/govspeak: ✔︎
 
 .govuk-govspeak {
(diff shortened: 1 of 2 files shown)


nickcolley:
Makes sense, thanks for the catch and quick fix 👍

PR #1003, opened 2017-04-21
No longer display the Article 50 banner

Rosa-Fox:


fofr:
Any reason we didn't remove the JS and styles for the banner?

PR #1005, opened 2017-04-21
Add missing step css from govspeak import

mcgoooo:
this was an oversight from a previous commit


mcgoooo:
How steps will look
screen shot 2017-04-24 at 15 02 54
To note, the below screenshot is how they look on live, the alignment is slightly out, but that wil be tackled in a further pull.


screen shot 2017-04-24 at 15 02 59

PR #1006, opened 2017-04-21
Add schema name analytics meta tag

nickcolley:
Part of https://trello.com/c/0xPPAdDd/286-track-schema-name-in-google-analytics

Paired with @michaelihejirika 

Also tested locally with government-frontend and can confirm calls to GA with dimension17 are set correctly.


fofr:
Has dimension 17 been set up in our analytics profile(s)?

michaelihejirika:
@fofr yes it has, Paul Cronk will turn it on as soon as we merge and deploy this.

mcgoooo:
i don't think this change is quite correct, i think we should keep all this code centralised
https://github.com/alphagov/slimmer/blob/master/lib/slimmer/processors/metadata_inserter.rb#L15

PR #1007, opened 2017-04-24
Enable GOV.UK analytics tracker configuration

boffbowsh:
The configuration variable is misnamed and will be changed once this is
proven to work. The endpoint is now served entirely by Fastly, and
returns an empty 200 response. We only call the endpoint using
`$.get()` and ignore the response so its `Content-Type` doesn’t matter.

### DIFF
--- a/app/assets/javascripts/analytics/init.js.erb
+++ b/app/assets/javascripts/analytics/init.js.erb
@@ -15,6 +15,8 @@
     universalId: universalId,
     cookieDomain: cookieDomain,
     allowLinker: true,
+    // This is served by Fastly in production, and returns an empty 200 response
+    govukTrackerGifUrl: '<%= asset_path "/static/a" %>',
   });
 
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/analytics/init.js.erb
```diff
@@ -15,6 +15,7 @@
     universalId: universalId,
     cookieDomain: cookieDomain,
     allowLinker: true,
+    govukTrackerGifUrl: '<%= asset_path "/static/a" %>',
```
cbaines: As this is served from Fastly, it would be good to put that in a comment here, as well as the details about the content and status code.

boffbowsh: Good point, done.

cbaines: :+1:

PR #1008, opened 2017-04-24
[DO NOT MERGE] Adds SRI to js and css files

elenatanasoiu:
[Trello card](https://trello.com/c/oFhtO0Gm/146-enable-subresource-integrity-sri-on-government-frontend-l)

This will add security checks to verify that the content being served on GOV.UK hasn't been modified by third parties:

- upgrades `sprockets-rails` to version 3.2.0 so that we can use the SRI check feature
- adds `integrity: true` to `javascript_include_tag` and `stylesheet_link_tag`

Explicitly setting `config.assets.digest` to `false`, as the default value is set to `true` since sprockets-rails 3.0: rails/sprockets-rails@3cb84ea

Depends on: 
- government-frontend: https://github.com/alphagov/government-frontend/pull/331
- govuk_template: https://github.com/alphagov/govuk_template/pull/294

In order to use SRI, the browser expects CORS to be supported. This is why we need to add `crossorigin: true` alongside `integrity: true`. See more [here](https://hacks.mozilla.org/2015/09/subresource-integrity-in-firefox-43/)


fofr:
Why does this PR depend on https://github.com/alphagov/government-frontend/pull/331?

elenatanasoiu:
@fofr The changes to SRI are added together in government-frontend, static and govuk_template.

government-frontend uses static to receive some assets + layout
static receives some assets from govuk_template that it passes on to government-frontend. 

Therefore these three projects have to go together and be tested together.

Gemfile
```diff
@@ -1,20 +1,20 @@
 source 'https://rubygems.org'
 
-gem 'rails', '4.2.7.1'
```
h-lame: Can you leave the version number in here; it's an easy shorthand to work out which version of rails we think this app is running on.  Intentionally or not by removing it you've bumped the app to rails 4.2.8, which isn't a problem but you didn't mention it in the commit note so perhaps you didn't intend it.

elenatanasoiu: I've but the rails version back.

elenatanasoiu:
@h-lame I've moved the changed from [PR 1041](https://github.com/alphagov/static/pull/1041) here

h-lame:
The CI failures here are because we need the fix in https://github.com/alphagov/govuk_template/pull/305 to be released (see: https://github.com/alphagov/govuk_template/pull/307)

h-lame:
Note that this is on hold while we investigate a firefox issue with SRI (see https://github.com/alphagov/government-frontend/pull/368)

h-lame:
See https://github.com/alphagov/govuk_template/pull/308 and https://github.com/alphagov/govuk_template/pull/301 for discussion on impact.  We're holding off on SRI until there are fewer users using the version of Firefox that has an issue with calculating SRI.

fofr:
Should we close this for now?

fofr:
Closing while SRI is on hold.

PR #1013, opened 2017-05-02
Pass custom dimensions directly to page tracking

alecgibson:
We’re seeing some weird behaviour on our Google Analytics tracking:
- Some pages appear to have a very high (10-100) repeat visit rate
  within a single (half hour) session
- Dimensions that rely on the DOM having loaded also are reporting 0
  count sometimes.

Noteably, when the count *is* reported correctly, the repeat visit
rate also appears to drop back to “normal levels” (~2).

This could be because of setting custom dimensions at the “session”
level, rather than passing them directly into the page view tracking,
but this is hard to test because we cannot reproduce the bug on our
machines.

In order to debug, we’re attempting to pass custom dimensions
directly to the page view tracking.

### Trello

### DIFF
--- a/app/assets/javascripts/analytics/static-analytics.js
+++ b/app/assets/javascripts/analytics/static-analytics.js
@@ -1,24 +1,18 @@
-(function() {
+(function () {
   "use strict";
   window.GOVUK = window.GOVUK || {};
 
-  var StaticAnalytics = function(config) {
+  var StaticAnalytics = function (config) {
 
     // Create universal tracker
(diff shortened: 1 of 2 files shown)


fofr:
Are you able to separate the whitespace/style changes from the feature change? It's making the diff harder to read.

carvil:
@fofr GitHub has a nice option for that, try this view instead: https://github.com/alphagov/static/pull/1013/files?w=1

It's a bit clearer than the normal diff I think.

PR #1017, opened 2017-05-03
Track sidebar navigation sections and links server-side

alecgibson:
This change reverts the previous attempts at tracking navigation sections
and links, and instead counts them in the component, and then reported
in `meta` tags. (These changes are two separate commits).

**Importantly** these tags assume that the `analytics_meta_tags`
component is being rendered in order to add extra content to this
component.

### Trello


fofr:
Interesting solution

app/assets/javascripts/analytics/static-analytics.js
```diff
@@ -176,7 +176,7 @@
 
   StaticAnalytics.prototype.setThemesDimension = function(themes) {
     this.setDimension(3, themes || 'other');
-  };
```
fofr: 🤔

carvil: This was a revert of a change that included this fix. We will add the fix again independently.

PR #1021, opened 2017-05-03
Fix a couple of IE7 errors

36degrees:
See individual commits for details.


nickcolley:
We should really have tests for this...

PR #1022, opened 2017-05-03
Fix the linting problems

mcgoooo:
a lot of this was done using the automated rubocop -a option


test/govuk_component/taxonomy_sidebar_test.rb
```diff
@@ -26,9 +26,6 @@ def component_name
         },
       ]
     )
-
```
tijmenb: Interesting finds here! @carvil looks like some tests in this file aren't actually testing anything?

carvil: Yup, it looks like we can delete this particular test!

test/govuk_component/taxonomy_sidebar_test.rb
```diff
@@ -26,9 +26,6 @@ def component_name
...
-    taxon_titles = css_select(".sidebar-taxon h2").map { |taxon_title| taxon_title.text.strip }
-    taxon_titles = ["Item 1 title", "Item 2 title"]
```
carvil: Rather than deleting both lines, could you please delete the last line and add the following:

```
assert_equal ["Item 1 title", "Item 2 title"], taxon_titles
```

Otherwise the entire test case still doesn't make sense.

mcgoooo: :+1:

mcgoooo:
@carvil fixed up, just need a passing review and i'll merge!

PR #1026, opened 2017-05-05
Revert navigation link tracking

alecgibson:
This commit reverts f22ab3ff8a1111e35c30ddb4f038cb557760e7dc

We’ve seen a bug in our Google Analytics (GA) tracking, where the number
of links and sections we report is sometimes incorrectly reported as
zero (‘0’) to GA, even despite performing the count server-side and
presenting it in a `meta` tag.

In this case, the only reasons for GA to incorrectly track the counts
should be:

- We are incorrectly counting the links server-side (but this should
  presumably break **all** cases, not just some)
- The `meta` tag is not present on the page (again, this should break
  **all** pages for a given rendering application)
- We’re doing something strange to our own tracking values in our
  JavaScript
- Something’s going wrong on transmission to GA

The bug is extremely hard to track down, and so far no one has been able
to reproduce it on their own machines. I’ve tried also using both
Smokey and Wraith to attempt a high volume of traffic on Production, but
all the counts were correct. The symptoms are listed here for posterity:

- For some pages, we see two values tracked for the number of sidebar
  sections and links. One value is the correct count. The other value is
  zero (‘0’)
- `/student-finance-calculator` is a consistent, high-volume example of
  such a page
- Inconsistent counts are seen independent of:
  - Navigation A/B test bucket
  - Rendering application
  - Browser
  - Operating System
- Where we see an incorrect count, the incorrect count has a very high
  ratio of (unique page views/page views) (~100+). This would mean each
  of these users was refreshing/revisiting the page over 100 times in a
  single half-hour session

### Trello


alecgibson:
We've solved the bug! 🎉

The causes were twofold:
- On `smart-answers`, our tracking is set up to aggregate pageviews underneath a given path (which is different to other pages on GOV.UK). Since the pages after the start page don't have a sidebar, these pages were _correctly_ reporting zero links and sections, but these results were then presented at the "root" of the `smart-answer` because of aggregation
- We still saw problems on publications, where such aggregation doesn't occur. Since we couldn't reproduce the bug, we couldn't track down exactly what the problem was, but moving to a server-side solution rather than JavaScript appears to have fixed the problem. The only conclusion we can draw is that we should avoid using complex JavaScript DOM operations in our tracking.

nickcolley:
Thanks for the update, nice work team figuring that out :)

fofr:
Our support ticket has been moved to done too:
https://trello.com/c/1cdu7atD/65-incorrect-counts-of-related-links-being-sent-to-analytics

PR #1028, opened 2017-05-05
Go back to js navigation link tracking

alecgibson:
This reverts commit dfcf9e8a8ef9ac51454671e62c759ae3a8128e6a, reversing
changes made to d3dd5f9554be1ff43dad43701efd2816ab84c20b.

We’ve seen a bug in our Google Analytics (GA) tracking, where the number
of links and sections we report is sometimes incorrectly reported as
zero (‘0’) to GA, even despite performing the count server-side and
presenting it in a `meta` tag.

We realised that part of the answer was because we aggregate
`smart-answers` paths, so pages after the start page (which legitimately
don't have a sidebar) will report a `0`, which is aggregated into the
`smart-answers` "root" path.

However, we also found a number of guidance content items which still
exhibit the bug, but without any of the path aggregation.

Since the bug is still affecting us even with server-side counting, we
will revert to using a JavaScript count, because a JavaScript count:

- will more accurately track what a user actually sees
- makes it easier to count things like accordions, which are split
  across several Ruby partials

The bug is extremely hard to track down, and so far no one has been able
to reproduce it on their own machines. I’ve tried also using both
Smokey and Wraith to attempt a high volume of traffic on Production, but
all the counts were correct. The symptoms are listed here for posterity:

- For some pages, we see two values tracked for the number of sidebar
  sections and links. One value is the correct count. The other value is
  zero (‘0’)
- Inconsistent counts are seen independent of:
  - Navigation A/B test bucket
  - Rendering application
  - Browser
  - Operating System
- Where we see an incorrect count, the incorrect count has a very high
  ratio of (unique page views/page views) (~100+). This would mean each
  of these users was refreshing/revisiting the page over 100 times in a
  single half-hour session

### Trello


fofr:
Is this data worth all this hassle?

alecgibson:
@fofr apparently so. We've checked that it's okay that the data might be a little off - it's correct for the majority of cases.

PR #1029, opened 2017-05-05
Add data attributes to track history link events

steventux:
https://trello.com/c/AmrwFMJU/22-implement-ga-event-tracking-on-content-history-see-all-updates-and-full-page-history-links

We'd like to understand how users interact with the page history
and updates links on a page, so add the data attributes which
enable tracking and provide the appropriate tracking options

cc @fofr

### DIFF
--- a/app/views/govuk_component/document_footer.raw.html.erb
+++ b/app/views/govuk_component/document_footer.raw.html.erb
@@ -40,7 +40,12 @@
     <% if history.any? %>
       <div class="change-notes" data-module="toggle">
         <p>
-          <a href="#full-history" data-controls="full-history" data-expanded="false">
+          <a href="#full-history" data-controls="full-history"
+                                  data-expanded="false"
+                                  data-module="track-click"
+                                  data-track-category="content-history"
+                                  data-track-action="full-page-history-toggled"
+                                  data-track-label="full-history">
             + <%= t("govuk_component.document_footer.full_page_history", default: "full page history") %>
           </a>
         </p>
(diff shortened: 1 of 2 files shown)


nickcolley:
Should these have aria-expanded for assistive technology?

app/views/govuk_component/document_footer.raw.html.erb
```diff
@@ -40,7 +40,11 @@
     <% if history.any? %>
       <div class="change-notes" data-module="toggle">
         <p>
-          <a href="#full-history" data-controls="full-history" data-expanded="false">
+          <a href="#full-history" data-controls="full-history"
+                                  data-expanded="false"
+                                  data-track-category="Content History"
+                                  data-track-action="fullPageHistoryOpened"
+                                  data-track-label="full-history">
```
fofr: Three different styles here:
* Words with spaces for category
* CamelCase for action
* Hyphenated for label

Could they all be hyphenated?

fofr: Does this also need `data-module="track-click"` to work?

steventux:
thanks @fofr updated

app/views/govuk_component/document_footer.raw.html.erb
```diff
@@ -40,7 +40,12 @@
     <% if history.any? %>
       <div class="change-notes" data-module="toggle">
         <p>
-          <a href="#full-history" data-controls="full-history" data-expanded="false">
+          <a href="#full-history" data-controls="full-history"
+                                  data-expanded="false"
+                                  data-module="track-click"
+                                  data-track-category="content-history"
+                                  data-track-action="full-page-history-opened"
```
fofr: This will track on open and close, might be worth saying: "full-page-history-toggled"

fofr:
One minor comment, otherwise 👍

PR #1030, opened 2017-05-08
Enable setting options for the next tracked pageview

alecgibson:
We currently have some generic methods for calling arbitrary methods
within the Analytics object on the next page view. This is actually only
ever used to set a custom dimension.

In order to close down our API, this change removes this generic,
reflective and overly open method, and instead exposes a method to set
an options object for the next tracked pageview, which is really what
we’re trying to achieve.

This will enable us to also remove the externally-exposed, specialised
dimension-setting methods.

A number of `TODO`s have been raised in this PR - these will be resolved
after we have merged https://github.com/alphagov/frontend/pull/1221

### Trello

https://trello.com/c/5dIbwHvU/120-tidy-up-tracking-changes

### DIFF
--- a/app/assets/javascripts/analytics/static-analytics.js
+++ b/app/assets/javascripts/analytics/static-analytics.js
@@ -10,9 +10,10 @@
     this.analytics = new GOVUK.Analytics(config);
 
     this.callMethodRequestedByPreviousPage();
+    var trackingOptions = getOptionsFromCookie();
 
     // Track initial pageview
-    this.trackPageview();
+    this.trackPageview(null, null, trackingOptions);
 
     // Begin error and print tracking
     GOVUK.analyticsPlugins.error({filenameMustMatch: /gov\.uk/});
@@ -24,6 +25,7 @@
     });
   };
 
+  // TODO: Remove once we're using setOptionsForNextPageview instead
   StaticAnalytics.prototype.callOnNextPage = function (method, params) {
     params = params || [];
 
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/analytics/static-analytics.js
```diff
@@ -117,6 +121,19 @@
...
+    if (!GOVUK.cookie || typeof options !== 'object') {
+      return;
+    }
+
+    var cookieOptions = getOptionsFromCookie();
+    $.extend(cookieOptions, options);
+
+    // Singly-stringified JSON sometimes gets escaped when put into a cookie, but inconsistently. The command-line
+    // tests will escape, but browser tests will not. Double-stringify in order to get consistently-escaped strings.
+    GOVUK.cookie('analytics_next_page_options', JSON.stringify(JSON.stringify(cookieOptions)));
```
carvil: We would probably need to update `/help/cookies` before we can use this live. Is there an alternative to using cookies?

alecgibson: Can I cheat by reusing the old cookie? The alternative may be using [`sessionStorage`](https://developer.mozilla.org/en/docs/Web/API/Window/sessionStorage)?

carvil: I think it's fine to re-use the old cookie (`analytics_next_page_call`) in this case.

alecgibson: Yup. Done this. As discussed, this change should be robust against which version of `frontend` we use:

- the old version will attempt to use `callOnNextPage`, which will store a function in the cookie, and then delete the cookie (so the new method doesn't find it)
- the new version will store a JSON object in the cookie, which won't resolve to a function when checked by `callMethodRequestedByPreviousPage`, so it will be ignored by the old code and consumed by the new code

PR #1035, opened 2017-05-10
Allow bundling with local govuk_frontend_toolkit

nickcolley:
Due to the way the toolkit is distributed we need to do some tomfoolery to get the local version in static, this makes it easy.

### DIFF
--- a/startup.sh
+++ b/startup.sh
@@ -11,6 +11,7 @@ if [[ $1 == "--test-govuk-frontend-toolkit" ]] ; then
   rm -rf ${temporary_location}
 
   # Copy the old assets aside
+  # Using sudo here since the installed location has elevated permissions
   sudo cp -r ${installed_location} ${temporary_location}
 
   # Remove current submoduled assets
--- a/startup.sh
+++ b/startup.sh
@@ -1,4 +1,29 @@
-#!/bin/bash 
+#!/bin/bash
 
 bundle install
+
+if [[ $1 == "--test-govuk-frontend-toolkit" ]] ; then
+  # Find out where it the gem is installed
+  installed_location="$(bundle show govuk_frontend_toolkit)"
+  temporary_location="./tmp/govuk_frontend_toolkit_gem_dev"
+
+  # Remove any existing tmp file
+  rm -rf ${temporary_location}
+
+  # Copy the old assets aside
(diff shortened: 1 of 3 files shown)


startup.sh
```diff
@@ -1,4 +1,29 @@
...
 
 bundle install
+
+if [[ $1 == "--test-govuk-frontend-toolkit" ]] ; then
+  # Find out where it the gem is installed
+  installed_location="$(bundle show govuk_frontend_toolkit)"
+  temporary_location="./tmp/govuk_frontend_toolkit_gem_dev"
+
+  # Remove any existing tmp file
+  rm -rf ${temporary_location}
```
fofr: This feels _mildly_ dangerous

startup.sh
```diff
@@ -1,4 +1,29 @@
...
+
+  # Copy the old assets aside
+  sudo cp -r ${installed_location} ${temporary_location}
```
fofr: Why is `sudo` needed here?

fofr:
I ran this script against my local version of toolkit `6.0.1` (it's two versions behind 6.0.3) using this branch.

When I ran `./startup.sh` again afterwards I expected everything to revert back to normal but found this in my diff (because we don't pin minor version):

Gemfile.lock:
```diff
-    govuk_frontend_toolkit (6.0.1)
+    govuk_frontend_toolkit (6.0.3)
```

Not sure if that might catch people out.

Other than a couple of minor comments, seems good :+1:

startup.sh
fofr: It'd be good for the comment to include _why_ they are being copied

nickcolley: Mainly because I didn't want to `sudo rm -rf` I think

fofr: Can you update the comment so it’s clear?

nickcolley:
@fofr updated the branch which addresses your out of date gem worry

vanitabarrett:
Just tried to run this and got the following error: 
`ln: failed to create symbolic link ‘./tmp/govuk_frontend_toolkit_gem_dev/app/assets’: No such file or directory`

Fixed by running `bundle show govuk_frontend_toolkit` manually and using this value to populate the installed_location variable in the startup file.

nickcolley:
Closing this since it's a bit hacky and doesnt seem to work everytime

PR #1038, opened 2017-05-15
Extract surveys.js as a separate JS bundle

h-lame:
For: https://trello.com/c/ZEJRuWiU/151-reduce-need-to-change-static-js-when-deploying-surveys-part-2

The contents of surveys.js can change frequently as we introduce new
surveys and retire old ones.  Every time we deploy these changes the
application.js bundle gets a new digest hash and so frequent visitors
have to re-download the whole JS bundle just to get these updates to
surveys.  To avoid this we separate the surveys.js so it is required
separately and changes to it will not change the digest of the main
application.js.

The assumption here is that applications using the JS provided by
static only do so by using the layouts via slimmer and not by requesting
the application.js or header-footer-only.js files directly.  If that
is the case they will also need to be updated to also request the
surveys.js bundle.

### DIFF
--- a/app/assets/javascripts/surveys.js
+++ b/app/assets/javascripts/surveys.js
@@ -1,9 +1,8 @@
-(function() {
-  "use strict";
+//= require_self
 
-  var root = this,
-      $ = root.jQuery;
-  if(typeof root.GOVUK === 'undefined') { root.GOVUK = {}; }
+(function($) {
+  "use strict";
+  window.GOVUK = window.GOVUK || {};
 
   var URL_SURVEY_TEMPLATE = '<section id="user-satisfaction-survey" class="visible" aria-hidden="false">' +
                             '  <div class="wrapper">' +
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/surveys.js
```diff
@@ -443,5 +445,9 @@
   };
 
   root.GOVUK.userSurveys = userSurveys;
+
+  $(document).ready(function() {
+    GOVUK.userSurveys.init();
```
nickcolley: Should we do a check to see if `GOVUK.userSurveys` exists?

nickcolley:
v noice 👍  Will review

app/assets/javascripts/surveys.js
h-lame: Sure.  The other document ready functions don't but there's no reason we shouldn't pave the way.  Will do.

h-lame: Done in a rebase.  Also made it use window.GOVUK instead of root.GOVUK so it's more like other JS files in this project.

fofr:
Has this been tested on integration, end to end with a couple of frontend apps?

h-lame:
@fofr tested on integration on a few pages and it seems to all work.  I also tested the "doesn't invalidate the main js bundle" part of it by deploying this, then deploying a version with a change to surveys.js.  The application.js asset didn't get a new digest, but the surveys.js did.  Are you happy for me to merge?

PR #1041, opened 2017-05-16
Bump govuk_template gem to make use of SRI changes

elenatanasoiu:
See https://github.com/alphagov/static/pull/1008


Gemfile.lock
```diff
@@ -281,4 +281,4 @@ DEPENDENCIES
   webmock
 
 BUNDLED WITH
-   1.14.5
+   1.14.6
```
h-lame: None of our rubies are pinned to bundler 1.14.6 yet, so this shouldn't be changed here.

config/routes.rb
```diff
@@ -10,12 +10,10 @@
 
   # Icon redirects
   get "/favicon.ico", to: "icon_redirects#show"
-  get "/apple-touch-icon.png", to: "icon_redirects#apple_60_60"
+  get "/apple-touch-icon.png", to: "icon_redirects#apple_touch_icon"
+  get "/apple-touch-icon-180x180.png", to: "icon_redirects#show"
+  get "/apple-touch-icon-167x167.png", to: "icon_redirects#show"
   get "/apple-touch-icon-152x152.png", to: "icon_redirects#show"
```
h-lame: Smarter folk than me might want to weigh in here with thoughts about us suddenly 404-ing all these old icons.  Should we leave the routes in and redirect to the closest current size that we do have an asset for?

app/controllers/icon_redirects_controller.rb
```diff
@@ -3,7 +3,7 @@ def show
     redirect_to view_context.asset_path(request.path.to_s[1..-1]), status: 301
   end
 
-  def apple_60_60
-    redirect_to view_context.asset_path('apple-touch-icon-60x60.png'), status: 301
+  def apple_touch_icon
+    redirect_to view_context.asset_path('apple-touch-icon.png'), status: 301
```
h-lame: The behaviour of this is now the same as the behaviour of `show` (e.g. `/apple-touch-icon.png` becomes a redirect to `asset/path/to/apple-touch-icon.png`) maybe we should get rid of it?

h-lame:
I'm also interested in how we make it more obvious that changing icons in govuk_template requires changes to static.  Are there docs anywhere about this coupling?

fofr:
For context, this is what happened last time app icon sizes were updated: https://github.com/alphagov/static/pull/453

test/integration/icon_redirects_test.rb
```diff
@@ -29,14 +28,13 @@ class IconRedirectsTest < ActionDispatch::IntegrationTest
   end
 
   [
-    'apple-touch-icon.png',
-    'apple-touch-icon-precomposed.png',
+    'apple-touch-icon.png'
   ].each do |file|
```
fofr: We can probably drop the `.each` here now the array is just one item.

elenatanasoiu:
@h-lame I've not been able to find documents on this issue, so I've created a PR [here](https://github.com/alphagov/govuk_template/pull/306)

h-lame:
Should this PR be a commit added on top of #1008 ?  They're related after all, and #1008 isn't merged yet.  #1008 also exhibits the same issue in Gemfile.lock that we've already fixed here (see https://github.com/alphagov/static/pull/1041#discussion_r116762719) so maybe it's worth squashing these down into one PR.

elenatanasoiu:
This PR has been merged with https://github.com/alphagov/static/pull/1008/commits/cab86e5dccb6a07e814a6cd3f90d7fa29e53f0f9

All changes have been moved there

PR #1044, opened 2017-05-17
Metadata component expand/collapse for long lists

andysellick:
* erb file now splits links into two chunks if longer than 5 (+2) items, including toggle link
* Applied to From, Part of and Other fields
* Tests added for long list with truncation required, short list with no truncation required
* Toggle JS updated to include text change on toggle option, tests updated

screen shot 2017-05-17 at 16 01 05


app/views/govuk_component/metadata.raw.html.erb
```diff
@@ -42,14 +46,16 @@
...
+      <% other.each_with_index do |(title, definition), index| %>
         <%
           definition ||= []
           definition = Array(definition)
         %>
         <% if definition.any? %>
           <dt><%= title %>:</dt>
-          <dd><%= definition.to_sentence.html_safe %></dd>
+          <dd>
+            <%= render file: 'govuk_component/metadata_sentence.raw', locals: { items: definition, toggle_id: index } %>
```
nickcolley: Is this going to end up with a lot of HTTP Requests from slimmer?

app/views/govuk_component/docs/metadata.yml
```diff
@@ -17,6 +17,15 @@ fixtures:
     from:
```
nickcolley: Should these lines be deleted?

andysellick: Sorry, not sure which lines you meant. If I understand the file structure correctly, lines 17 and 20 start separate examples in the component guide. Lines 17 to 19 was an existing one, I added lines 20 - 28 as a test of the new expand/collapse when applied to the 'from' field.

app/views/govuk_component/metadata.raw.html.erb
andysellick: Excellent question, I have no idea! If it does I'm not sure what alternative to use. I guess if all of the sections were handled together in a loop (like we originally discussed) then we wouldn't need a partial and there'd be no extra requests. Happy to continue to look into that if you'd like.

app/views/govuk_component/docs/metadata.yml
fofr: @andysellick Looking at the full diff shows the problem:

```
extensive:
    from:
    - <a href='/government/organisations/ministry-of-defence'>Ministry of Defence</a>
    - <a href='/government/organisations/cabinet-office'>Cabinet Office</a>
    from:
    - <a href='/government/organisations/ministry-of-defence'>Ministry of Defence</a>
    - <a href='/government/organisations/cabinet-office'>Cabinet Office</a>
```

app/views/govuk_component/metadata_sentence.raw.html.erb
```diff
@@ -0,0 +1,16 @@
...
+  remaining = []
+  if items.length > (list_limit + 2)
+    remaining = items[list_limit..items.length]
+    items = items[0...list_limit]
+  end
+%>
+<% if remaining.any? %>
+  <%= items.to_sentence(last_word_connector: ', ').html_safe %><span id="toggle-<%= toggle_id %>" class="js-hidden">, <%= remaining.to_sentence.html_safe %></span>
+  <div class="toggle-wrap">
+    <a href="#" data-controls="toggle-<%= toggle_id %>" data-expanded="false" data-toggled-text="- show fewer">+ <%= remaining.length %> more</a>
```
fofr: The text strings here _may_ need to be in the locale file for translations

test/govuk_component/metadata_test.rb
```diff
@@ -59,4 +59,48 @@ def component_name
     assert_definition('Part of:', 'this and that')
     assert_definition('Related topics:', 'a, b, and c')
   end
+
+  test "rendering of long metadata into wrapping for Javascript interaction" do
```
fofr: This test name reads a little oddly.
Maybe: "long lists of metadata are hidden behind toggle"

test/govuk_component/metadata_test.rb
```diff
@@ -59,4 +59,48 @@ def component_name
...
+      "<a href=\"/business-finance-support?industries%5B%5D=travel-and-leisure\">Travel and leisure</a>",
+      "<a href=\"/business-finance-support?industries%5B%5D=utilities-providers\">Utilities providers</a>",
+      "<a href=\"/business-finance-support?industries%5B%5D=wholesale-and-retail\">Wholesale and retail</a>"
+    ]
+    render_component(other: {
+        "Industry": links
+    })
+    assert_select "span", count: 1
+    assert_select "dd > a", count: limit
+    assert_select "dd span a", count: links.length - limit
```
fofr: Should we test `part_of` and `from` too?

app/views/govuk_component/docs/metadata.yml
andysellick: Oh I see. You mean the first example is redundant? I left it in so there was an example of both a non-truncated one and a truncated one. Happy to remove if it's not necessary.

app/views/govuk_component/metadata_sentence.raw.html.erb
andysellick: Happy to do if someone can point me in the right direction.

test/govuk_component/metadata_test.rb
andysellick: Amended.

test/govuk_component/metadata_test.rb
andysellick: Tests expanded to include part_of and from.

app/views/govuk_component/docs/metadata.yml
andysellick: Have removed the first one - there was already an example of a non-truncated 'from' in the page.

config/locales/en.yml
```diff
@@ -25,6 +25,8 @@ en:
       published: "Published"
       last_updated: "Last updated"
       see_all_updates: "see all updates"
+      toggle_less: "Show fewer"
+      toggle_more: "more"
```
fofr: There's some Rails magic you can do to include the number here as well. Eg I'm EN it's 11 more, but in other languages the number may be at the end or in the middle.

app/views/govuk_component/metadata.raw.html.erb
fofr: @nickcolley Can you add your findings?

nickcolley: We had a look in Kibana and it seemed to have an expected amount of requests, we want to do a little bit more investigation when deploying but are otherwise happy.

PR #1045, opened 2017-05-18
Linting to unpollute metadata pull request

andysellick:
* Changes in metadata expand/collapse PR had included linting, making reading the changes difficult. This commit contains all of the linting and none of the changes, so should make that PR more readable once merged

### DIFF
--- a/app/assets/javascripts/start-modules.js
+++ b/app/assets/javascripts/start-modules.js
@@ -1,9 +1,9 @@
-//= require govuk/modules
-//= require modules/global-bar
-//= require modules/sticky-element-container
-//= require modules/toggle
-//= require modules/track-click
+// = require govuk/modules
+// = require modules/global-bar
+// = require modules/sticky-element-container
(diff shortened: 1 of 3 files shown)


mcgoooo:
@andysellick looks fine

app/assets/javascripts/start-modules.js
```diff
@@ -1,9 +1,9 @@
-//= require govuk/modules
-//= require modules/global-bar
-//= require modules/sticky-element-container
-//= require modules/toggle
-//= require modules/track-click
+// = require govuk/modules
```
fofr: Presumably this doesn't break the way Rails asset pipeline includes these files?

andysellick: I don't know about all of them, but toggle definitely still works in the front end.

mcgoooo: it shouldn't do, it should include everything in modules

PR #1046, opened 2017-05-18
Next and previous buttons icons updated

andysellick:
* Added new SVGs for next and previous
* SVGs not supported from IE8 and below, so no icon appears, no fallback implemented (shout if there should be one)
* Would have used CSS transform and only one SVG but IE9 doesn't support rotating an SVG
* Modified pagination slightly to not output pagination label element if no label content
screen shot 2017-05-18 at 14 26 47

https://trello.com/c/gzo59MFU/355-1-deployment-previous-and-next-arrows


app/assets/images/govuk-component/arrow-right.svg
```diff
@@ -0,0 +1,86 @@
...
+     inkscape:cy="6.6840729"
+     inkscape:window-x="0"
+     inkscape:window-y="23"
+     inkscape:window-maximized="0"
+     inkscape:current-layer="svg2" />
+  <!-- Generator: Sketch 42 (36781) - http://www.bohemiancoding.com/sketch -->
+  <title
+     id="title4">Group</title>
+  <desc
+     id="desc6">Created with Sketch.</desc>
```
mcgoooo: can you run your svg's through an optimiser please

nickcolley:
When we tweak the arrows for the wonkyness we should also optimise them to reduce the size.

https://jakearchibald.github.io/svgomg/ works well for this

@andysellick can you link the trello card so it's easy for people to check in review? 👍

PR #1047, opened 2017-05-18
Make steps not images and accessible via css

mcgoooo:
So the Initial aim was just to add some css so i could access the numbers through css
i then dicovered a few issues with them. The first was the excessive amount
of css it generated due to having two images. The next one was that the
content was in an image, so didnt scale properly or accessible in the dom.

due to the concerns of adding extra html in, i decided to use before and after
which seems to be reasonably well supported. It's an improvement over what
was there, but maybe not perfect. it seems to be supported by a lot of assistive
technology, but not maybe it all, which is an improvement over none.

### old
screen shot 2017-05-18 at 14 23 05

### new
screen shot 2017-05-18 at 14 26 02

### DIFF
--- a/app/assets/stylesheets/govuk-component/govspeak/_steps.scss
+++ b/app/assets/stylesheets/govuk-component/govspeak/_steps.scss
@@ -9,15 +9,27 @@
     list-style-type: decimal;
     margin-left: 0;
     padding: 0.75em 0 0.75em 2.2em;
+    position: relative;
+
+    &:before {
+      position: absolute;
+      left: 0;
+      margin-top: -1px;
+      height: 18px;
+      width: 18px;
+      padding: 4px;
+      border-radius: 50%;
+      background-color: $black;
+      color: $white;
+      @include core-16();
+      font-weight: bold;
+      text-align: center;
+    }
 
     @for $i from 1 through 14 {
-      &:nth-child(#{$i}) {
-        background-image: image-url("icon-steps/icon-step-#{$i}.png");
-
-        @include device-pixel-ratio() {
-          background-image: image-url("icon-steps/icon-step-#{$i}-2x.png");
-          background-size: 24px 24px;
-        }
+      &:nth-child(#{$i}):before,
+      .steps-step#{$i}:before {
+        content: "#{$i}";
       }
     }
   }


app/assets/stylesheets/govuk-component/govspeak/_steps.scss
```diff
@@ -1,3 +1,6 @@
+$black: #000;
+$white:#fff;
```
fofr: Why hasn't our linter picked up on the whitespace here ^

fofr: Black and white are already defined here:
https://github.com/alphagov/govuk_frontend_toolkit/blob/89168934d1455754f455c398ba5c350a484d9a2a/stylesheets/colours/_palette.scss

app/assets/stylesheets/govuk-component/govspeak/_steps.scss
```diff
@@ -9,15 +12,27 @@
...
-      &:nth-child(#{$i}) {
-        background-image: image-url("icon-steps/icon-step-#{$i}.png");
+    &:before {
+      position: absolute;
+      left: 0;
+      margin-top: -1px;
+      height: 18px;
+      width: 18px;
+      padding: 4px;
+      border-radius: 18px;
```
fofr: Could use `border-radius: 50%`

app/assets/stylesheets/govuk-component/govspeak/_steps.scss
mcgoooo: knew they would be hidden somewhere! cheers

app/assets/stylesheets/govuk-component/govspeak/_steps.scss
```diff
@@ -9,15 +12,27 @@
...
+      @include core-16();
+      font-weight: bold;
+      text-align: center;
+    }
 
-        @include device-pixel-ratio() {
-          background-image: image-url("icon-steps/icon-step-#{$i}-2x.png");
-          background-size: 24px 24px;
-        }
+    @for $i from 1 through 14 {
```
36degrees: Have you explored [CSS counters](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Lists_and_Counters/Using_CSS_counters) for this? Would need some testing but I _think_ the browser support is a lot better than you might expect…

mcgoooo: yeah we had a little discussion about that, did you see the extra line of css i added? i need to be able to get at each one explcitity with `step-step1`

fofr:
Should we do this upstream in Elements and govuk_frontend_toolkit too?

app/assets/stylesheets/govuk-component/govspeak/_steps.scss
36degrees: But I think with counters you wouldn't need any `nth-child` rules at all? Just increment a counter each `.step`.

mcgoooo:
@fofr probably once we leave this one to settle in, i think we should maybe just have one implementation of it.

fofr:
The way breadcrumbs was done is a good example.

app/assets/stylesheets/govuk-component/govspeak/_steps.scss
mcgoooo: i need one starting at two https://www.gov.uk/child-benefit-tax-calculator/main

mcgoooo: i need one starting at 2..

fofr:
Have found that this has already been done in Elements:
https://govuk-elements.herokuapp.com/icons-images/example-icons/

Using: `circle circle-step`

mcgoooo:
we should look at replacing out the one in static with this, but i think this one is still valid, the element one includes extra markup, this one doesn't.

fofr:
@mcgoooo What's stopping this getting merged?

fofr:
Closing this as we are taking a different approach.

PR #1048, opened 2017-05-18
Remove unused tracking code

alecgibson:
As part of a refactor to reduce the number of public functions we expose on the `StaticAnalytics` object, we have removed now-unused public methods:

- `callOnNextPage`
- `callMethodRequestedByPreviousPage`
- `setSectionDimension`
- `setResultCountDimension`
- `setSearchPositionDimension`

Notably, we have _not_ removed `setDimension`, which is still in use by a range of applications across GDS.

### Trello

https://trello.com/c/5dIbwHvU/120-tidy-up-tracking-changes

### DIFF
--- a/app/assets/javascripts/analytics/static-analytics.js
+++ b/app/assets/javascripts/analytics/static-analytics.js
@@ -28,23 +28,6 @@
     GOVUK.Analytics.load();
   };
 
-  // TODO: Remove this, and its corresponding call in collections
-  StaticAnalytics.prototype.setSectionDimension = function () {
-  };
-
-  // TODO: We're setting this at a session level, because it's called in frontend's live-search.js to update
-  // the search count. We should make this consistent with the other dimensions and pass the dimension
-  // directly into the pageview arguments.
-  StaticAnalytics.prototype.setResultCountDimension = function (count) {
-    this.setDimension(5, count);
-  };
-
-  // TODO: We're setting this at a session level, because it's used by search through callOnNextPage. We should
-  // make this consistent with the other dimensions and pass the dimension directly into the pageview arguments.
-  StaticAnalytics.prototype.setSearchPositionDimension = function (position) {
-    this.setDimension(21, position);
-  };
-
   StaticAnalytics.prototype.trackPageview = function (path, title, options) {
     var trackingOptions = this.getAndExtendDefaultTrackingOptions(options);
     this.analytics.trackPageview(path, title, trackingOptions);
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/modules/logger.js
```diff
@@ -0,0 +1,30 @@
...
+      if (typeof console[key] === 'function') {
+        fakeConsole[key] = function() {};
+      }
+    });
+
+    return fakeConsole;
+  }
+
+  GOVUK.Logger = Logger;
+})();
```
mcgoooo: i am not convinced this is a good idea, what material difference does it make on production? and also, the way static is deployed currently, unless someone is actively working on it, this is less likely to get picked up easily.

the console is not user facing from an average users POV

alecgibson: I'm fine to remove this - I was unsure of what our policy was on abusing `console`. Do we prefer having developer messages visible in the production console, then?

mcgoooo: it's not really abusing console imo, we have no release mechanism for static, so this is about the best you can do, i don't thunk there is a policy. Also can i ask why we are deprecating it? as in, it seems as if it could hang about?

alecgibson: It could probably hang about, and if people think there's a real use case for it, then I can remove the deprecation warning altogether. The reason for the deprecation warning was because the code is now promoting explicitly passing custom dimensions to tracking code so you know exactly what you're getting; page-level things should ideally be set in `meta` tags.

mcgoooo: yep, i just see 56 mentions of it in code, just unsure as to when we may get all these bits deprecated. i just didnt know if it was causing problems, or jsut tryign to put people down a certain path

alecgibson: Definitely the latter, but I'm not too fussed I guess; the code is now written this way to be far more explicit about:

- which methods are _actually_ used by external JavaScript (we previously exposed **everything** and figuring out what was actually being used externally was pretty hard to see)
- make it more explicit what we're trying to track when setting custom dimensions by passing them directly into the tracking event/pageview they're intended for

In terms of how people actually want to use this, though, we could just leave it open?

mcgoooo: yeah, for patterns like this, i am not too sure how it fits in, but the revealing module pattern is the way to go i find, anyway you have a general approval (not too fussed about the logger tbh)

alecgibson: Fine. I was on the fence about the logger anyway, so I've removed it.

PR #1055, opened 2017-05-23
Remove unused global bar

fofr:
We're running unnecessary JS at the top of every page, and including JS and styles we don't need.

* Remove global bar JS, styles, templates and tests
* Global bar used for EU referendum was brought back for Article 50 notification – #947
* It was partially removed in #1003

@timblair @nickcolley @Rosa-Fox


fofr:
Double checked with @boffbowsh that we're ok to remove all of it.

nickcolley:
image

💯 💥

PR #1056, opened 2017-05-23
Metadata rewritten to remove partial

andysellick:
* Now loops through all attributes and outputs using a single view within the same template

https://trello.com/c/GToVab74/386-improve-metadata-expand-collapse-code


fofr:
I think this makes a strong case for @carvil's RFC: https://github.com/alphagov/govuk-rfcs/pull/71. Ideally we'd move a lot of this logic out into helpers with their own tests, and allow code re-use between components.

app/views/govuk_component/metadata.raw.html.erb
```diff
@@ -9,55 +9,102 @@
...
-      <dt><%= t("govuk_component.metadata.history", default: "History") %>:</dt>
-      <dd><%= history %></dd>
-    <% end %>
-    <% if local_assigns.include?(:first_published) && first_published %>
-      <dt><%= t("govuk_component.metadata.published", default: "Published") %>:</dt>
-      <dd><%= first_published %></dd>
-    <% end %>
-    <% if local_assigns.include?(:last_updated) && last_updated %>
-      <dt><%= t("govuk_component.metadata.last_updated", default: "Last updated") %>:</dt>
+    <% metadata_items.each_with_index do |(title, items, extratext), index| %>
```
fofr: `extratext` is more readable as `extra_text` and as use is quite specific, `see_all_updates_text` might be better still

fofr: I'd be tempted to pass in a chunk of HTML that gets appended, allowing any type of metadata to append something if it needs to.

nickcolley:
@fofr coming from the React world where component specific behaviour is co-located within components themselves, this does not feel weird to me.

I don't agree with moving code away from a component in the off chance it'll be reused, or that testing data manipulation code specific to a component also should be tested external to how the component renders.

That said I agree if we started duplicating similar logic between components that'd be a +1 for what Carlos' RFC describes, which would end up happening eventually.

app/views/govuk_component/metadata.raw.html.erb
andysellick: I wrote it that way initially by @nickcolley suggested the markup should be in the view. We could always amend it to that in future if needed?

fofr:
@nickcolley We may want the same toggle logic for long lists in the document_footer component. How should that logic be shared?

carvil:
Right now its tricky to share code between components unfortunately. I don't necessarily think having too much logic in the components is helpful even from a readability perspective, but I do know the existing limitations of static.

We will be doing a spike soon in order to see if we can prove that returning rendered HTML is a viable option. If that works, then we will be able to turn static into an easier to read/use application where we can split the code between there expected places.

nickcolley:
@fofr agreed, we would definitely want to share the code then 👍

nickcolley:
I'm going to close this out for now pending @carvil investigations, then we can revisit this.

PR #1057, opened 2017-05-23
Indent report a problem toggle and form correctly

### LINKED ISSUE #998: Layout for feedback slightly broken
The layout for 'Is there something wrong with this page' is slightly off - it's indented too far.

Screenshot:
screen shot 2017-04-11 at 12 06 31

screen shot 2017-04-11 at 12 03 39

Publishing Frontend Support: https://trello.com/c/VVGQ6axB/61-layout-for-feedback-slightly-broken

fofr:
When on mobile, in some layouts the report a problem link and form get indented twice – once because of a 15px margin applied to #wrapper, and once again on report a problem itself.

* If the link is nested within wrapper, don’t indent again
* Keeps indentation if no wrapper is present

The homepage is a special case, it uses #wrapper but removes the margin

* Include site-width-container again with higher specificity for the homepage. This prevents the `0` margin resets from taking effect.

## Before
screen shot 2017-05-23 at 13 38 21

## After
screen shot 2017-05-23 at 13 38 07

Previous attempt to fix: https://github.com/alphagov/static/pull/660 reverted in https://github.com/alphagov/static/pull/662

Fixes #998

### DIFF
--- a/app/assets/stylesheets/helpers/_report-a-problem.scss
+++ b/app/assets/stylesheets/helpers/_report-a-problem.scss
@@ -2,10 +2,29 @@
   display: none;
 }
 
-.report-a-problem-container{
+.report-a-problem-container,
+.report-a-problem-toggle-wrapper {
   @extend %site-width-container;
   clear: both;
 
+  // TODO: Refactor to avoid having to extend site-width-container
+  //       The pattern should be flexible in its placement
+  // #wrapper extends `site-width-container` too, which leads to
+  // double margins on mobile viewports, reset margins when nested
+  // within #wrapper
+  #wrapper & {
+    margin-left: 0;
+    margin-right: 0;
+  }
+
+  // The homepage removes margins from wrapper
+  // Be specific to override the above margin resets
+  .homepage #wrapper & {
+    @extend %site-width-container;
+  }
+}
+
+.report-a-problem-container {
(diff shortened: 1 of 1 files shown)


fofr:
## Homepage

screen shot 2017-05-23 at 13 45 48
screen shot 2017-05-23 at 13 45 57

fofr:
Tested by deploying to integration and looking at Whitehall, Homepage, Frontend and government-frontend layouts. Also searched for other `#wrapper` resets but found none.

nickcolley:
The problem here is that the component should not define a width or horizontal margin, it should sit inside the parent wrapper.

I'd prefer to leave this as is and figure out what the work is to have this component sit within the pages' wrapper since this seems like it could break

fofr:
@nickcolley Using the site on mobile, this bug continues to bother me. A "done properly" fix is complex and requires upgrading to the wrapper layout on our older pages, which I think will happen organically but over a long time.

A more complex fix is also more likely to break, or take more time testing/developing.

nickcolley:
I'd be okay merging if there was a TODO with a basic description of why this hack exists 👍

fofr:
Added explanatory comment and a `TODO` about preferred styling.

benilovj:
🎆 🎉

PR #1058, opened 2017-05-23
Add an analytics dimension for 'govuk:content-has-history'

thomasleese:
This will be set in the various frontends.

[Trello Card](https://trello.com/c/uHiHx5RQ/56-custom-dimension-to-identify-session-where-content-history-is-available)

### DIFF
--- a/app/views/govuk_component/analytics_meta_tags.raw.html.erb
+++ b/app/views/govuk_component/analytics_meta_tags.raw.html.erb
@@ -77,6 +77,15 @@
   meta_tags["govuk:taxon-ids"] = taxon_ids.join(',') unless taxon_ids.empty?
   meta_tags["govuk:taxon-slug"] = taxon_slugs_without_theme.first unless taxon_slugs_without_theme.empty?
   meta_tags["govuk:taxon-slugs"] = taxon_slugs_without_theme.join(',') unless taxon_slugs_without_theme.empty?
+
+  def has_content_history(content_item)
+    details = content_item.fetch(:details, {})
+
+    (content_item[:public_updated_at] && details[:first_public_at] && content_item[:public_updated_at] != details[:first_public_at]) ||
+      (details[:change_history] && details[:change_history].size > 1)
+  end
+
+  meta_tags["govuk:content-has-history"] = "true" if has_content_history(content_item_hash)
 %>
 <% meta_tags.each do |name, content| %>
   <meta name="<%= name %>" content="<%= content %>">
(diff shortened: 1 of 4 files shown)


kevindew:
I'm not too familiar with static. Code seems to make sense but I'm a bit unsure about the means to determine if something has history.

How confident are you with: `content_item[:public_updated_at] != content_item[:first_public_at]` ?

It was my understanding that first_public_at appeared in details and was just whitehall specific :-/ but even so, I'd like to understand more. My understanding is that we want to know when change notes are visible and an option so that we can determine number of change notes visible vs times they're clicked. Is that correct? 

PS. Trello card link is to board rather than card.

thomasleese:
@kevindew Hmm, yes, I based this code on https://github.com/alphagov/government-frontend/blob/master/app/presenters/updatable.rb#L34 which is used to determine if the `history` array is empty, and if it is not empty, then the history is displayed by https://github.com/alphagov/static/blob/master/app/views/govuk_component/document_footer.raw.html.erb .

My first attempt on `government-frontend` only: https://github.com/alphagov/government-frontend/pull/372

kevindew:
Ahh sure, I see they're accessing first_public_at in details there. I don't know how that'd work for say specialist_publisher though :-/ 

Would we also need to check that there are items in the change_history array?

thomasleese:
Oh yes, I've fixed it to look at details hash and updated it to include a check on `change_history`.

app/views/govuk_component/analytics_meta_tags.raw.html.erb
```diff
@@ -77,6 +77,15 @@
   meta_tags["govuk:taxon-ids"] = taxon_ids.join(',') unless taxon_ids.empty?
   meta_tags["govuk:taxon-slug"] = taxon_slugs_without_theme.first unless taxon_slugs_without_theme.empty?
   meta_tags["govuk:taxon-slugs"] = taxon_slugs_without_theme.join(',') unless taxon_slugs_without_theme.empty?
+
+  def has_content_history(content_item)
```
steventux: Presumably we have to embed methods and conditional logic here because this is a `raw.html.erb` template? :cry:

app/views/govuk_component/analytics_meta_tags.raw.html.erb
```diff
@@ -77,6 +77,15 @@
...
+    details = content_item.fetch(:details, {})
+
+    (content_item[:public_updated_at] && details[:first_public_at] && content_item[:public_updated_at] != details[:first_public_at]) ||
```
steventux: Are we sure that `:first_public_at` is the correct key? My understanding of this is that [it's an alias for `:first_published_at`](https://github.com/alphagov/whitehall/blob/master/app/presenters/publishing_api/payload_builder/first_public_at.rb#L5) and [not used consistently even in Whitehall where it originates](https://github.com/alphagov/whitehall/search?p=1&q=first_public_at&type=&utf8=%E2%9C%93). 

That said I've little knowledge of how government-frontend deals with these fields so I'm just raising this as a concern, if you are satisfied it's the right field then :+1:

PR #1059, opened 2017-05-23
Adjusting header search box focus style

### LINKED ISSUE #848: Changes to template header search will need minor change to static
Changes in [this GOV.UK Template PR](https://github.com/alphagov/govuk_template/pull/259) will mean a minor fix is needed in static.

That PR moves the focus outline inside of the input, but this leaves a white border to the right of the search input. This is caused by the search in put and button being separated and their container having a white background.

Rather oddly this white border is 1px in Chrome, but 2px in Safari.

Don't know what the fix would be, as the white background is needed to make the placeholder work.

The button also has a 1px left border of #222 ideally this would be changed to our standard black.

Screenshot of the issue:
screen shot 2016-11-04 at 14 29 48

andysellick:
* Safari rendered this with a 1px gap between the input and the button, all other non-IE browsers happy
* Fixes  #848 

**Safari before**
screen shot 2017-05-24 at 08 15 47

**Safari after**
screen shot 2017-05-24 at 08 15 57

**Chrome before**
screen shot 2017-05-24 at 08 14 24

**Chrome after**
screen shot 2017-05-24 at 08 14 38

Worth noting that the negative outline offset isn't supported in IE (even newer versions) so this is generally broken/different in IE anyway. An alternative could be to change the border colour rather than use an outline but I assume there's a good reason for this so didn't want to go off the reservation.

### DIFF
--- a/app/assets/stylesheets/helpers/_header.scss
+++ b/app/assets/stylesheets/helpers/_header.scss
@@ -109,6 +109,7 @@
       &.focus,
       &:focus {
         background: #fff;
+        outline-offset: -3px;
       }
 
       @include ie-lte(7){
@@ -116,7 +117,7 @@
       }
       .js-enabled & {
         width: 86%;
-        @include calc(width, "100% - 37px");
+        @include calc(width, "100% - 35.1px"); /* 35.1 is for Safari, which renders this input too narrow */
       }
     }
 
@@ -157,6 +158,11 @@
 
       &:hover {
         background-color: darken($mainstream-brand, 5%);
+      }
+
+      &.focus,
+      &:focus {
+        outline-offset: -3px;
       }
     }
 
--- a/app/assets/stylesheets/helpers/_header.scss
+++ b/app/assets/stylesheets/helpers/_header.scss
@@ -90,6 +90,9 @@
       border: 0;
       min-height: 36px;
       padding: 6px 0;
+      &::-moz-focus-inner {
+        border: 0;
+      }
       @include ie-lte(8) {
         min-height: 24px;
       }


fofr:
Screenshots please.

fofr:
Is the change around the search button focus intentional?

## Safari before
screen shot 2017-06-07 at 16 30 47

## Safari after
screen shot 2017-06-07 at 16 30 52

app/assets/stylesheets/helpers/_header.scss
```diff
@@ -109,14 +109,15 @@
...
+        outline-offset: -3px;
       }
 
       @include ie-lte(7){
         padding-left: 0;
       }
       .js-enabled & {
         width: 86%;
-        @include calc(width, "100% - 37px");
+        @include calc(width, "100% - 35.1px"); /* 35.1 is for Safari, which renders this input too narrow */
```
fofr: Is it specifically the `.1` that's for Safari?

fofr:
Firefox on Windows 10 has a black dotted outline within the search button that looks _stranger_ with this change:

## Old
screen shot 2017-06-07 at 16 37 45

## New
screen shot 2017-06-07 at 16 38 33

app/assets/stylesheets/helpers/_header.scss
andysellick: It is, yes. Every other browser seems to understand an integer width correctly, but Safari for some reason drops a pixel. Setting it to 35.1px makes Safari over-compensate while other browsers round it down. It's weird, I know.

andysellick:
Firefox second outline fixed (although I feel like this particular fix should be more generally applied, ideally to all inputs and buttons in Elements, but that's probably out of scope for this PR).

Outline change noted in Safari is intentional, yes - search input had inset focus already, so I made the button consistent with that.

PR #1060, opened 2017-05-23
Remove CSS previously used by wrapper layout

fofr:
* Remove travel advice fixes
* Remove old business support styles
* Remove print-link styles
* Remove programme styles
* Remove JS and styles for collapsed multi page navigation (still used here https://www.gov.uk/temporary-events-notice/south-tyneside/apply but is more consistent and accessible without the JS)

Each of these formats/patterns is now rendered using government-frontend. They have been removed from alphagov/frontend.


fofr:
I'd love to get a review on this.

nickcolley:
Having trouble checking this locally, need to get licensing running to try it...

fofr:
@nickcolley Didn't think Licensify used Static. Can you link to where it pulls it in?

nickcolley:
frontend uses licensify (without declaring it as a dependency) to populate those pages, so fails if I try to load without that endpoint setup...

```
Failed to open TCP connection to licensify.www.gov.uk:443 (getaddrinfo: Name or service not known)
```

PR #1063, opened 2017-06-01
Better algorithm for picking surveys

h-lame:
For: https://trello.com/c/X88IDU6Q/117-permit-concurrent-short-surveys-for-the-same-page

The previous algorithm was to choose the default survey and then ask every
other survey defined in `smallSurveys` in turn if it was active. If the
survey says it is then we record it as the active survey and move on to
the next survey to ask if it is active. Once we have a survey then we
check if the survey has already been completed, and roll the dice to check
against the frequency.  Only then do we show it.

There are a couple of problems with this algorithm.  First; if we have two
or more active surveys that could run on a given page we always pick the
one defined last in smallSurveys.  This means that on that page we'll
never show the ones defined earlier.  Specifically this means that we'll
never show the default survey on a page that another survey is running on.
This is not about showing multiple surveys on a page at one time, just that
across all page views we will only ever show the last defined one.  Second;
by choosing the active survey before we check if it can be shown we run the
risk of picking a survey that we can't actually show and thus showing no
survey at all when another survey we could have picked would be showable.

To solve this we select all possible small surveys for the page looking at
their start and end times, and at the `activeWhen` function.  We add the
default survey to this list and then filter out any that can't be shown
because they have already been completed, or their frequency doesn't
match.  If after doing this we have more than one survey we roll the dice
to pick one of them.  It should be safe enough to do this as we have no
other way of choosing which should run, and we've already checked their
frequencies.  This means that a high frequency survey that runs on a page
won't always be shown in favour of a low frequency survey that also runs
on that page, and also that the last defined survey won't always be the
winner either.

Note: the above taken from the commit message on the `Improve mechanism for selecting which survey to run` commit, you'll want to check the other commit too.

### DIFF
--- a/spec/javascripts/surveys-spec.js
+++ b/spec/javascripts/surveys-spec.js
@@ -33,6 +33,7 @@ describe('Surveys', function () {
     $block = $('<div class="govuk-emergency-banner" style="display: none"></div>' +
                '<div id="global-cookie-message" style="display: none"></div>' +
                '<div id="global-browser-prompt" style="display: none"></div>' +
+               '<div id="taxonomy-survey" style="display: none"></div>' +
                '<div id="user-satisfaction-survey-container"></div>')
 
     $('body').append($block)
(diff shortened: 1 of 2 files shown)


h-lame:
This is gonna clash with #1066 because there's a bunch of linting in that PR and many changes.  As that PR is more important it makes sense to let it land first and we'll deal with the conflicts in here.

h-lame:
Rebased and reworked now that #1066 has been merged.

spec/javascripts/surveys-spec.js
```diff
@@ -478,56 +478,145 @@ describe('Surveys', function () {
...
-      var activeSurvey = surveys.getActiveSurvey(defaultSurvey, smallSurveys)
-      expect(activeSurvey).toBe(defaultSurvey)
+      var activeSurveys = surveys.getActiveSurveys([testSurvey])
+      expect(activeSurveys).not.toContain(testSurvey)
     })
 
-    it('returns the small survey when a smallSurvey is active', function () {
-      var smallSurveys = [smallSurvey]
-      spyOn(surveys, 'currentTime').and.returnValue(new Date('July 9, 2016 10:00:00').getTime())
+    it("includes the a survey when we are between it's start and end times", function () {
```
elenatanasoiu: "includes the* survey when we are between its* start and end times"

spec/javascripts/surveys-spec.js
```diff
@@ -478,56 +478,145 @@ describe('Surveys', function () {
...
+        startTime: new Date("July 5, 2016").getTime(),
+        endTime: new Date("July 10, 2016 23:50:00").getTime(),
+        url: 'example.com/small-survey'
+      }
+
+      var displayableSurveys = surveys.getDisplayableSurveys([testSurvey])
+      expect(displayableSurveys).not.toContain(testSurvey)
+    })
+
+    it("includes a survey if it should not be displayed", function () {
```
elenatanasoiu: "if it should be displayed"

spec/javascripts/surveys-spec.js
h-lame: Ooops.  Fixed in rebase.

spec/javascripts/surveys-spec.js
h-lame: Ooops.  Fixed in rebase.

PR #1064, opened 2017-06-01
Check for curated taxonomy sidebars

alecgibson:
In order to allow for the display of curated taxonomy sidebars, we need
to display related links for an arbitrary number of sections. This
change will ignore the check that hides links beyond the second section,
if the sidebar has been curated.

### Trello

https://trello.com/c/J8ofEnrt/167-display-curated-related-links-in-the-sidebar

### DIFF
--- a/app/views/govuk_component/taxonomy_sidebar.raw.html.erb
+++ b/app/views/govuk_component/taxonomy_sidebar.raw.html.erb
@@ -28,7 +28,7 @@
         </p>
 
         <% item[:related_content] ||= [] %>
-        <% if item[:related_content].any? && item_index < 2 %>
+        <% if item[:related_content].any? && (item_index < 2 || item[:is_curated]) %>
           <nav role='navigation'>
             <ul class='related-content'>
               <% item[:related_content].each_with_index do |related_item, related_content_index| %>


app/views/govuk_component/taxonomy_sidebar.raw.html.erb
```diff
@@ -28,7 +28,7 @@
         </p>
 
         <% item[:related_content] ||= [] %>
-        <% if item[:related_content].any? && item_index < 2 %>
+        <% if item[:related_content].any? && (item_index < 2 || item[:is_curated]) %>
```
carvil: I think this logic should be moved to the navigation helpers. I've attempted to do that in the past, but the PR wasn't merged because we weren't ready to do this. Here is how it looked:

- https://github.com/alphagov/govuk_navigation_helpers/pull/68/commits/a02334378abd7003aa76e4fd307c504bcf5de263

What do you think?

alecgibson: Yeah that's a much cleaner change - I'll raise a couple of separate PRs.

alecgibson:
We're going to move this logic into `govuk_navigation_helpers`.

PR #1067, opened 2017-06-05
Fix for sticky contents link overlap issue

andysellick:
* At foot of page was possible to get 'Contents' fixed position link to briefly overlap content below it
* Problem was due to height of parent element including height of Contents link, then changing after it was set to position fixed
* Fixed by defaulting Contents link to position absolute so it never adds any height to the parent

Relates to https://github.com/alphagov/government-frontend/pull/367

### DIFF
--- a/app/assets/stylesheets/modules/_sticky-element-container.scss
+++ b/app/assets/stylesheets/modules/_sticky-element-container.scss
@@ -1,8 +1,8 @@
-.sticky-element {
+.govuk-sticky-element {
   position: absolute;
   bottom: 0;
 
-  &.govuk-sticky-element {
+  &.enabled {
     @include transition (opacity, .3s, ease);
     opacity: 1;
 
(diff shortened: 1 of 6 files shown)


app/assets/stylesheets/modules/_sticky-element-container.scss
```diff
@@ -1,17 +1,6 @@
...
-    @include ie-lte(8) {
-      // XXX: Would be nice to `@include .visuallyhidden`.
-      position: absolute;
-      left: -9999em;
-    }
-  }
+  position: absolute;
+  bottom: 0;
 
   &--stuck-to-parent {
```
fofr: Is this block used now? Can it be removed? I can't find any references to `stuck-to-parent`.

fofr:
When the JS fails on mobile it looks broken:

## Old
screen shot 2017-06-07 at 17 08 50

## New
screen shot 2017-06-07 at 17 07 17

andysellick:
Have removed unnecessary CSS and fixed when JS disabled.

fofr:
Link behaves nicer now, top work :100: 👍

PR #1068, opened 2017-06-08
Add ecommerce search results plugin

dwhenry:
We want to start using the Ecommerce tooling to track search results that are not clicked on a page. 

In order to minimise the number of data requests that we send to GA we need to setup the ecommerce data prior to performing the pageview event.

https://trello.com/c/WlBZo8Sl/118-move-ecommerce-ga-to-use-govukfrontendtoolkit

### DIFF
--- a/app/assets/javascripts/analytics/ecommerce.js
+++ b/app/assets/javascripts/analytics/ecommerce.js
@@ -41,7 +41,9 @@
         });
 
         ga('ec:setAction', 'click', {list: 'Site search results'});
-        ga('send', 'event', 'UX', 'click', 'Results');
+        ga('send', 'event', 'UX', 'click', 'Results',
+          GOVUK.CustomDimensions.getAndExtendDefaultTrackingOptions({})
+        );
       })
     }
   }
@@ -49,7 +51,6 @@
   Ecommerce.ecLoaded = false;
   Ecommerce.start = function (element) {
     element = element || $('[data-analytics-ecommerce]');
-
     if(element.length > 0) {
       if(!Ecommerce.ecLoaded) {
         ga('require', 'ec');
--- a/app/assets/javascripts/analytics/ecommerce.js
+++ b/app/assets/javascripts/analytics/ecommerce.js
@@ -1,0 +1,65 @@
+(function () {
+  "use strict";
+  window.GOVUK = window.GOVUK || {};
+
+  var Ecommerce = function (config) {
+    this.init = function (element) {
+      this.forEachEcommerceRow(element, this._sendImpression);
+      this.forEachEcommerceRow(element, this._trackProductClick);
+    }
+
+    this.forEachEcommerceRow = function (element, fct) {
+      var ecommerceRowCss = '[data-ecommerce-row]';
+      var ecommerceRows = element.find(ecommerceRowCss);
+      var startPosition = parseInt(element.data('ecommerce-start-index'));
+
+      ecommerceRows.each(function(index, ecommerceRow) {
+        var $ecommerceRow = $(ecommerceRow);
+
+        var contentId = $ecommerceRow.attr('data-ecommerce-content-id'),
+          path = $ecommerceRow.attr('data-ecommerce-path');
+
+        fct($ecommerceRow, contentId, path, index + startPosition);
+      });
+    }
+
+    this._sendImpression = function (_row, contentId, path, position) {
+      // We only send the id to GA as additional product data is linked when it is uploaded.
+      // This approach is taken to avoid the GA data packet exceeding the 8k limit
+      ga('ec:addImpression', {
+        id: contentId || path,
+        position: position,
+        list: 'Site search results'
+      });
+    }
+
+    this._trackProductClick = function (row, contentId, path, position) {
+      row.click(function(event) {
+        ga('ec:addProduct', {
+          id: contentId || path,
+          position: position
+        });
+
+        ga('ec:setAction', 'click', {list: 'Site search results'});
+        ga('send', 'event', 'UX', 'click', 'Results');
+      })
+    }
+  }
+
+  Ecommerce.ecLoaded = false;
+  Ecommerce.start = function (element) {
+    element = element || $('[data-analytics-ecommerce]');
+
+    if(element.length > 0) {
+      if(!Ecommerce.ecLoaded) {
+        ga('require', 'ec');
+        Ecommerce.ecLoaded = true;
+      }
+      var ecommerce = new Ecommerce();
+      ecommerce.init(element);
+    }
+  }
+
+  window.GOVUK.Ecommerce = Ecommerce;
+  window.GOVUK.StaticAnalytics.beforeTrackPage = Ecommerce.start;
+})()
--- a/app/assets/javascripts/analytics/ecommerce.js
+++ b/app/assets/javascripts/analytics/ecommerce.js
@@ -1,17 +1,15 @@
+// https://developers.google.com/analytics/devguides/collection/analyticsjs/enhanced-ecommerce
 (function () {
   "use strict";
   window.GOVUK = window.GOVUK || {};
 
   var Ecommerce = function (config) {
     this.init = function (element) {
-      this.forEachEcommerceRow(element, this._sendImpression);
-      this.forEachEcommerceRow(element, this._trackProductClick);
-    }
-
-    this.forEachEcommerceRow = function (element, fct) {
-      var ecommerceRowCss = '[data-ecommerce-row]';
-      var ecommerceRows = element.find(ecommerceRowCss);
-      var startPosition = parseInt(element.data('ecommerce-start-index'));
+      // Limiting to 100 characters to avoid noise from extra longs search queries
+      // and to stop the size of the payload going over 8k limit.
+      var searchQuery   = element.attr('data-search-query').substring(0, 100).toLowerCase();
+      var ecommerceRows = element.find('[data-ecommerce-row]');
+      var startPosition = parseInt(element.data('ecommerce-start-index'), 10);
 
       ecommerceRows.each(function(index, ecommerceRow) {
         var $ecommerceRow = $(ecommerceRow);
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/analytics/ecommerce.js
```diff
@@ -0,0 +1,66 @@
...
+
+    this.forEachEcommerceRow = function (element, fct) {
+      var ecommerceRowCss = '[data-ecommerce-row]';
+      var ecommerceRows = element.find(ecommerceRowCss);
+      var startPosition = parseInt(element.data('ecommerce-start-index'));
+
+      ecommerceRows.each(function(index, ecommerceRow) {
+        var $ecommerceRow = $(ecommerceRow);
+
+        var content_id = $ecommerceRow.attr('data-ecommerce-content-id'),
```
MatMoore: Not sure why this isn't caught by a linter, but shouldn't `content_id` be `contentId` throughout?

app/assets/javascripts/analytics/ecommerce.js
```diff
@@ -0,0 +1,66 @@
...
+        });
+
+        ga('ec:setAction', 'click', {list: 'SITE_SEARCH_RESULTS'});
+        ga('send', 'event', 'UX', 'click', 'Results',
+          GOVUK.CustomDimensions.getAndExtendDefaultTrackingOptions({}));
+      })
+    }
+  }
+
+  Ecommerce.ec_loaded = false;
```
MatMoore: Another snake case var here.

app/assets/javascripts/analytics/ecommerce.js
```diff
@@ -0,0 +1,66 @@
...
+        ga('require', 'ec');
+        Ecommerce.ec_loaded = true;
+      }
+      var ecommerce = new Ecommerce();
+      ecommerce.init(element);
+    }
+  }
+
+  window.GOVUK.Ecommerce = Ecommerce;
+  window.GOVUK.StaticAnalytics.beforeTrackPage = Ecommerce.start;
```
MatMoore: Is there an assumption here about the order these modules are initialised in?

If so is there a better way of wiring these up, like using data-attributes? (I'm not very familiar with the frontend architecture)

dwhenry: Yes, using the existing `module` code would be nicer. 

Unfortunately, it didn't work as the modules are loaded after the pageview code executes.

The load order is explicit set in `app/assets/javascripts/analytics.js` unfortunately we can't really test this without a heap of overhead - dummy rails app with slimmer to stimulate the actual page load.

I think what we have is good enough for now as the implementation is all or nothing, i.e. if it breaks we shouldn't get any data in GA.

app/assets/javascripts/analytics/ecommerce.js
dwhenry: good shout..

app/assets/javascripts/analytics/ecommerce.js
```diff
@@ -0,0 +1,66 @@
...
+  "use strict";
+  window.GOVUK = window.GOVUK || {};
+
+  var Ecommerce = function (config) {
+    this.init = function (element) {
+      this.forEachEcommerceRow(element, this._sendImpression);
+      this.forEachEcommerceRow(element, this._trackProductClick);
+    }
+
+    this.forEachEcommerceRow = function (element, fct) {
```
nickcolley: nit: fct -> full name e.g. callback

fofr:
cc'ing @alecgibson and @carvil as both have worked on analytics recently, and had many issues surrounding timing of custom dimensions tracking.

app/assets/javascripts/analytics/ecommerce.js
```diff
@@ -0,0 +1,62 @@
...
+
+        var contentId = $ecommerceRow.attr('data-ecommerce-content-id'),
+          path = $ecommerceRow.attr('data-ecommerce-path');
+
+        self._sendImpression($ecommerceRow, contentId, path, index + startPosition);
+        self._trackProductClick($ecommerceRow, contentId, path, index + startPosition);
+      });
+    }
+
+    this._sendImpression = function (_row, contentId, path, position) {
```
alecgibson: Are these functions (this and `trackProductClick`) used outside of this module? If not, then do they need to be publicly available?

app/assets/javascripts/analytics/ecommerce.js
```diff
@@ -0,0 +1,62 @@
...
+      // We only send the id to GA as additional product data is linked when it is uploaded.
+      // This approach is taken to avoid the GA data packet exceeding the 8k limit
+      ga('ec:addImpression', {
```
alecgibson: Is this using the globally-available `ga` object? I don't think that's how other tracking is done; I think it all passes through [`GOVUK.Analytics`](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/javascripts/govuk/analytics/analytics.js), which then forwards tracking events both to GA, and on to our own S3-hosted tracking.

app/assets/javascripts/analytics/ecommerce.js
```diff
@@ -0,0 +1,62 @@
...
+        id: contentId || path,
+        position: position,
+        list: 'Site search results'
+      });
+    }
+
+    this._trackProductClick = function (row, contentId, path, position) {
```
alecgibson: (`trackProductOnClick`? Otherwise it sounds like you're actually doing the tracking when you call this method)

app/assets/javascripts/analytics/ecommerce.js
```diff
@@ -0,0 +1,62 @@
...
+      row.click(function(event) {
+        ga('ec:addProduct', {
+          id: contentId || path,
+          position: position
+        });
+
+        ga('ec:setAction', 'click', {list: 'Site search results'});
+        ga('send', 'event', 'UX', 'click', 'Results',
+          GOVUK.CustomDimensions.getAndExtendDefaultTrackingOptions({})
```
alecgibson: (It might be nice to add a method `getDefaultTrackingOptions`?)

app/assets/javascripts/analytics/ecommerce.js
alecgibson: Instead of doing this, can't we just directly invoke `Ecommerce.start` from `static-analytics.js`? Then we wouldn't need to introduce this slightly obscure `beforeTrackPage` mechanism, and the dependency is explicit at the desired point of invocation.

app/assets/javascripts/analytics/ecommerce.js
dwhenry: I spoke with @nickcolley  when we started on this work and the decision was made to not have this code in the `frontend toolkit` until we could prove that it worked/we would be keeping it. It's for this reason that we are accessing `ga` directly here. If we are happy with the code we can move this whole file into the frontend toolkit which will then mirror the remainder of the codebase.

alecgibson: Okay, fine by me so long as you're tracking that somewhere.

app/assets/javascripts/analytics/ecommerce.js
dwhenry: Not really as the code needs to run each time the pageview code is triggered, and as this can happen as a result of an ajax response we need to use this sort of mechanism.

alecgibson: I'm not sure I follow. Why can't you just rewrite `StaticAnalytics.trackPageView` to directly call `Ecommerce.start` instead of invoking `beforeTrackPage`? Are there any instances where you wouldn't want to start Ecommerce when tracking the pageview?

app/assets/javascripts/analytics/ecommerce.js
```diff
@@ -0,0 +1,62 @@
+(function () {
```
fofr: Is it worth including some links to the GA docs at the top?

app/assets/javascripts/analytics/ecommerce.js
```diff
@@ -0,0 +1,62 @@
...
+  var Ecommerce = function (config) {
+    this.init = function (element) {
+      var self = this;
+      var ecommerceRows = element.find('[data-ecommerce-row]');
+      var startPosition = parseInt(element.data('ecommerce-start-index'));
+
+      ecommerceRows.each(function(index, ecommerceRow) {
+        var $ecommerceRow = $(ecommerceRow);
+
+        var contentId = $ecommerceRow.attr('data-ecommerce-content-id'),
```
fofr: Minor: Is there a preference for `$ecommerceRow.attr('data-ecommerce-content-id')` over `$ecommerceRow.data('ecommerce-content-id')`? If the value of this data was altered the former wouldn't pick up the latest value.

app/assets/javascripts/analytics/ecommerce.js
```diff
@@ -0,0 +1,62 @@
+(function () {
+  "use strict";
+  window.GOVUK = window.GOVUK || {};
+
+  var Ecommerce = function (config) {
+    this.init = function (element) {
+      var self = this;
+      var ecommerceRows = element.find('[data-ecommerce-row]');
+      var startPosition = parseInt(element.data('ecommerce-start-index'));
```
fofr: Doesn't calling `.data` coerce the value into an integer? (`parseInt` should also have a base `10` passed in as the 2nd argument – eg if the HTML has the value "09" then this would be parsed as an incorrect number on old browsers.

spec/javascripts/analytics/ecommerce.spec.js
```diff
@@ -0,0 +1,167 @@
...
+      dimension27: '0',
+    })
+  });
+
+  it('will only require the ec library once', function() {
+    GOVUK.Ecommerce.ecLoaded = false;
+    GOVUK.Ecommerce.start($('<div></div>'));
+    GOVUK.Ecommerce.start($('<div></div>'));
+
+    expect(ga).toHaveBeenCalledWith('require', 'ec');
```
fofr: :+1:

spec/javascripts/analytics/static-analytics-spec.js
```diff
@@ -49,6 +49,43 @@ describe("GOVUK.StaticAnalytics", function() {
...
+             <div \
+               data-ecommerce-row\
+               data-ecommerce-path="/path/to/page"\
+               data-ecommerce-content-id="static-analytics-test"\
+             </div>\
+           </div>\
+          </div>\
+        ')
+        analytics = new GOVUK.StaticAnalytics({universalId: 'universal-id'});
+        pageViewObject = getPageViewObject();
```
fofr: Is this `pageViewObject` used?

app/assets/javascripts/analytics/ecommerce.js
fofr: Is `content-id` the content ID we use for documents? Or is it an ecommerce specific content ID? I've assumed the former.

alecgibson: (`.data` will attempt coerce the value into an appropriate JavaScript type, whereas `.attr` won't - this is important when storing integers that get used in custom dimensions, because custom dimensions can only accept `string` types. That said, I'm not sure that's important here?)

app/assets/javascripts/analytics/ecommerce.js
```diff
@@ -0,0 +1,71 @@
...
+        });
+
+        ga('ec:setAction', 'click', {list: 'Site search results'});
+        ga('send', 'event', 'UX', 'click', 'Results',
+          GOVUK.CustomDimensions.getAndExtendDefaultTrackingOptions({dimension71: searchQuery})
+        );
+      })
+    }
+
+    this._truncateQuery = function (searchQuery) {
```
dwhenry: as discuss this can just be inlined

app/assets/javascripts/analytics/ecommerce.js
```diff
@@ -0,0 +1,71 @@
...
+      // We only add the id to GA as additional product data is linked when it is uploaded.
+      // This approach is taken to avoid the GA data packet exceeding the 8k limit
+      ga('ec:addImpression', {
+        id: contentId || path,
+        position: position,
+        list: 'Site search results'
+      });
+    }
+
+    this._trackProductOnClick = function (row, contentId, path, position, searchQuery) {
```
dwhenry: change to ```function trackProductOnClick``` so the method is not public (same for `addImpression`)

fofr:
Updates to my comments look good :+1:. Will leave for @alecgibson to give the OK.

PR #1074, opened 2017-06-27
Document Enhanced E-commerce plugin

MatMoore:
Add some documentation to explain how we've implemented Enhanced E-commerce. 
Trello: https://trello.com/c/jIJOJwXB/142-documentation-for-enhanced-e-commerce

Rendered: https://github.com/alphagov/static/blob/f82a2e82c93906eb9d970f29de793d196584242c/doc/architecture-decisions/0002-add-enhanced-ecommerce.md

I've put this into [ADR](http://thinkrelevance.com/blog/2011/11/15/documenting-architecture-decisions) format so we can keep this close to the code. I'm open to suggestions if people would rather move it somewhere else, like the developer docs.

### DIFF
--- a/doc/architecture-decisions/0002-add-enhanced-ecommerce.md
+++ b/doc/architecture-decisions/0002-add-enhanced-ecommerce.md
@@ -1,0 +1,49 @@
+# 2. Add Enhanced E-commerce plugin
+
+Date: 2017-06-27
+
+## Status
+
+Accepted
+
+## Context
+
+Rendering apps track pageviews and events to help us measure the performance
+of GOV.UK.
+
+Static uses an [abstraction layer](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/docs/analytics.md) in GOV.UK Frontend Toolkit, which reports the data to Google Analytics (but the underlying provider could be changed).
+
+For analysing search performance, we want to be able to reliably track clickthrough rates for search result links.
+
+[Enhanced E-Commerce](https://developers.google.com/analytics/devguides/collection/analyticsjs/enhanced-ecommerce) is a plugin for google analytics that can track impressions as well as clicks. We model each search result page as a "product list", and each content item as a "product".
+
+## Decision
+
+We added a plugin for E-Commerce to Static, so we could try using it
+to measure search performance.
+
+We didn't add an abstraction layer to GOV.UK Frontend Toolkit, because we don't
+know if there is a wider need for it, and we wanted to validate our use case with GOV.UK data. The code could be moved to the Frontend Toolkit at a later date if necessary.
+
+We're sending the search query as a custom dimension. This is a workaround: without the custom dimension, Enhanced-Ecommerce
+can infer it from the path, but only for impression data. We couldn't get it to set the query for the click data.
+
+For each "product" we send the id, position, and list ('site search result'). To avoid unnecessary data transfer when users are interacting with the page, we use Query Time Imports to add
+metadata to products later.
+
+We use publishing platform `content_id` as the unique id where available, and use the path as a fallback.
+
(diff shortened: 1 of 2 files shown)


doc/architecture-decisions/0002-add-enhanced-ecommerce.md
```diff
@@ -0,0 +1,46 @@
...
+For analysing search performance, we want to be able to reliably track clickthrough rates on links in search results.
+
+[Enhanced E-Commerce](https://developers.google.com/analytics/devguides/collection/analyticsjs/enhanced-ecommerce) is a plugin for google analytics that can track impressions as well as clicks. We model each search result page as a "product list", and each content item as a "product".
+
+## Decision
+
+We added a plugin for E-Commerce to Static, so we could try using it
+to measure search performance.
+
+We didn't add an abstraction layer to GOV.UK Frontend Toolkit, because we don't
```
suzannehamilton: IIRC, the Frontend team would have preferred us to implement it in Frontend Toolkit. And the reason we didn't do it like that (as well as your reasons below) was that it made it technically challenging to ensure the GA events all fired in the correct order. Is that right, @brenetic?

nickcolley: The reason you've listed is correct but we also would prefer it was iterated on in static and try to keep govuk_frontend_toolkit as lightweight as possible

nickcolley:
We could include the risks associated with plugin injection here too

doc/architecture-decisions/0002-add-enhanced-ecommerce.md
```diff
@@ -0,0 +1,46 @@
...
+know if there is a wider need for it, and we wanted to validate our use case with GOV.UK data.
+
+We send the search query as a custom dimension, because although Enhanced-Ecommerce
+can pick it up from the path ....it didn't work .... ?????
```
dwhenry: We send the search query as a custom dimension, because although Enhanced-Ecommerce can pick it up from the path for the impression data, we were not able to get it to populate the click data. We do not know/understand why it doesn't work for the click data.

doc/architecture-decisions/0002-add-enhanced-ecommerce.md
```diff
@@ -0,0 +1,46 @@
...
+
+For each "product" we send the content id and path, so we can uniquely identify
```
dwhenry: we only send the id, position, list ('site search result') and query term not path.

doc/architecture-decisions/0002-add-enhanced-ecommerce.md
```diff
@@ -0,0 +1,46 @@
...
+each page. To avoid unnecessary data transfer when users are interacting with the page, we use Query Time Imports to add
+metadata to products later.
```
dwhenry: Mention that GA has a hard limit of 8k and if the data packet is larger than this it will fail to send any page view data.  Further, we could mention that increases in the size of this data packet can affect page view data upload and data usage limit on people with slower connections. This is most relevant to people using a mobile device which overseas.

doc/architecture-decisions/0002-add-enhanced-ecommerce.md
dwhenry: we use content-id for the id field, with a fallback to the url when we don't have one (external links, etc)

doc/architecture-decisions/0002-add-enhanced-ecommerce.md
dwhenry: The code was written so it could be ported to Frontend Toolkit with minimal changes.

MatMoore:
Have added a note about the plugin injection risk.

PR #1075, opened 2017-06-27
Add limiting to how many times a user can be shown a survey

h-lame:
For: https://trello.com/c/WaguHidO/184-adding-a-display-limit-to-the-govuk-survey-banner

We want to increase the frequency that some surveys appear, but we don't want to annoy frequent visitors to GOV.UK who show no interest in the survey.  To this end we:

1. introduce a surveySeen cookie for each survey that tracks how many times it has been presented to the user
2. don't show the survey to a user if it has been shown already to that user twice
3. allow a survey to configure the limit for how many times it considers too many (default is 2)
4. allow a survey to configure the number of days to persist the cookie that counts how many times it has been seen (default is to use a session cookie)

We also reduce the number of days for the "I've taken this survey don't show it again" cookie from 120 days (~4 months) to 90 days (~3 months).

### DIFF
--- a/app/assets/javascripts/surveys.js
+++ b/app/assets/javascripts/surveys.js
@@ -265,7 +265,7 @@
     isSurveyToBeDisplayed: function (survey) {
       if (GOVUK.cookie(userSurveys.surveyTakenCookieName(survey)) === 'true') {
         return false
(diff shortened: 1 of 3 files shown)


h-lame:
Removed DNM as we've now clarified with the survey team.

PR #1077, opened 2017-07-04
Metadata font size increase

andysellick:
* Increase from 14px to 16px
* Rewritten slightly to use mobile first
* Width of dt and dd set at 30% and 70%, max width applied to prevent large gap between the two when in a wide space
* Fixed bug where elements collapsed to single column then back to two column before ending up as single column as screen width reduced

**Before, desktop:**

screen shot 2017-07-04 at 16 22 51

**After, desktop:**

screen shot 2017-07-04 at 16 23 46

**Before, mobile:**

screen shot 2017-07-04 at 16 24 33

**After, mobile:**

screen shot 2017-07-04 at 16 25 37

https://trello.com/c/2pik3I0j/109-1-increase-font-size-in-metadata-component-in-static

### DIFF
--- a/app/assets/stylesheets/govuk-component/_metadata.scss
+++ b/app/assets/stylesheets/govuk-component/_metadata.scss
@@ -1,5 +1,5 @@
 .govuk-metadata {
-  @include core-14;
+  @include core-16;
   @include responsive-bottom-margin;
   @extend %contain-floats;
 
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/govuk-component/_metadata.scss
```diff
@@ -9,13 +9,16 @@
...
-    clear: left;
-    width: auto;
-    min-width: 120px;
+    margin-top: 0.5em;
+    line-height: normal;
 
     @include media(tablet) {
+      float: left;
+      clear: left;
+      width: calc(30% - 10px);
```
nickcolley: Do we need to do this if we use `box-sizing: border-box`?

andysellick: No, we could use border-box. That would just mean two lines of CSS instead of one.

andysellick: You're right - better to use border-box. Have changed.

PR #1079, opened 2017-07-07
Adding search component

andysellick:
* Intended initially to replace search box on homepage only, but aim is to expand options on component so it can replace other, similar search boxes on GOV.UK
* Renders by default to appear on a white background, other options detailed in component guide
* See https://github.com/alphagov/frontend/pull/1260 for this component being used

https://trello.com/c/QUGWKwTe/431-unify-search-boxes-as-component

screen shot 2017-07-17 at 12 26 26

Unusued JS as a toggle-class module in this gist for reference: https://gist.github.com/andysellick/90d68e33e8cf5056ad9ce5d86b0b3269


app/assets/javascripts/govuk-component/search.js
```diff
@@ -0,0 +1,28 @@
...
+    });
+  }
+
+  GOVUK.SearchBox = SearchBox;
+
+  $('.govuk-search').each(function(){
+    GOVUK.SearchBox($(this));
+  });
+
+})(jQuery);
```
fofr: Should this JS be written as a module? https://github.com/alphagov/govuk_frontend_toolkit/blob/master/docs/javascript.md#modules

fofr: The component would be wrapped with `<div data-module="search-box"></div>` for instance, and the instantiation can be removed from the .js file.

app/assets/javascripts/govuk-component/search.js
```diff
@@ -0,0 +1,28 @@
+(function ($) {
+  "use strict";
+  window.GOVUK = window.GOVUK || {};
+
+  function SearchBox($el){
+    var $searchFocus = $el.find('.js-search-input-focus');
```
fofr: This JS feels like a generic "add focus class"  behaviour rather than something directly associated with search.

app/assets/javascripts/govuk-component/search.js
andysellick: Have now converted the JS to a module.

app/assets/javascripts/govuk-component/search.js
andysellick: Have now converted the JS to a module.

app/assets/javascripts/modules/toggle-class.js
```diff
@@ -0,0 +1,72 @@
+/*
```
fofr: As discussed offline, I think this would be simpler re-producing just the module behaviour that's needed eg `toggle-class-on-focus`. It doesn't need to be this generic yet, if/when more event actions are needed then this would be useful. For the purpose of this PR I think it obscures the module's purpose.

fofr: There was some misunderstanding around "making a module generic".

* making a needed behaviour easily re-usable elsewhere
* making a configurable module that would meet potential future needs (needs that may never arise), eg other event types

spec/javascripts/modules/toggle-class-on-focus.spec.js
```diff
@@ -0,0 +1,72 @@
+/* eslint-env jasmine */
+/* eslint-disable no-multi-str */
+
+describe('A search box module', function () {
```
fofr: I'd change the description.

spec/javascripts/modules/toggle-class-on-focus.spec.js
```diff
@@ -0,0 +1,72 @@
...
+  var element, toggle;
+
+  afterEach(function(){
+    element.remove();
+  });
+
+  describe('when the search box is interacted with', function () {
+    beforeEach(function () {
+      element = $('\
+        <form class="govuk-search" action="/search" method="get" role="search" data-module="toggle-class-on-focus">\
```
fofr: You could probably make these test fixtures more minimal – ie the minimum markup needed to make this work:

```html
<form data-module="toggle-class-on-focus">
  <input class="js-class-toggle">
</form>
```

app/views/govuk_component/search.raw.html.erb
```diff
@@ -0,0 +1,21 @@
+<%
+  appearance_class ||= "on-light-background"
```
nickcolley: This exposes the internal implementation to applications that are using it, so if we wanted to rename this class we couldn't easily.

We could have a 'inverse' flag or something else and then flip the class internally.

app/views/govuk_component/search.raw.html.erb
```diff
@@ -0,0 +1,21 @@
...
+  layout_class ||= ""
```
nickcolley: Same as above, we could have a 'size' input?

app/views/govuk_component/search.raw.html.erb
```diff
@@ -0,0 +1,21 @@
...
+  search_term ||= ""
```
nickcolley: Should we mirror HTML where possible so this would be 'search_value' or 'value'?

spec/javascripts/modules/toggle-class-on-focus.spec.js
andysellick: Good point.

app/assets/stylesheets/govuk-component/_search.scss
```diff
@@ -0,0 +1,184 @@
...
+    .search-element {
+      position: relative;
+      display: table-cell;
+      vertical-align: top;
+    }
+
+    .search-input {
+      margin: 0;
+
+      // the .focus class is added by JS and ensures that the input remains above the label once clicked/filled in
```
fofr: Minor: I think `hide-search-label` could be a better class name. Or `hide-placeholder-label`.
It's not strictly always on focus.

nickcolley: Any internal changes shouldn't block this PR unless we can't iterate later easily.

app/views/govuk_component/search.raw.html.erb
andysellick: Good point, altered.

app/views/govuk_component/search.raw.html.erb
andysellick: We only have one size variant at the moment (large) so I've implemented that as a flag. If we have more than one in the future we'd perhaps need another mechanism but there currently are no other sizes I'm aware of on gov.uk.

app/views/govuk_component/search.raw.html.erb
```diff
@@ -0,0 +1,25 @@
+<%
+  className = "on-light-background"
```
nickcolley: Just realised, I think this shouldnt camelcase 🤔

andysellick: Fixed.

app/views/govuk_component/docs/search.yml
```diff
@@ -0,0 +1,19 @@
+name: "Search"
+description: "Search box"
+fixtures:
+  default: {}
+  for_use_on_a_dark_but_not_black_background:
```
fofr: Would this be better as `on_govuk_blue`? That's the manuals use case?
The dark background parameter might suggest it is ok to use on black.

app/views/govuk_component/docs/search.yml
```diff
@@ -0,0 +1,19 @@
...
+  stop_the_label_appearing_inline:
+    inline_label: false
+  apply_custom_styling_to_the_label:
+    inline_label: false
+    label_text: "<span class=\"fixture-title\">Search this website</span>"
+  set_id_for_search_input:
+    id: "my_unique_id"
+  set_search_input_value:
+    value: "This is what I searched for"
+  large_version:
```
fofr: An example with many of the parameters set would be useful, possibly at the start.

app/views/govuk_component/docs/search.yml
```diff
@@ -0,0 +1,19 @@
+name: "Search"
+description: "Search box"
```
fofr: Other things that should be indicated upfront:
* Component must be included within a form
* Search input uses the name `q`, which isn't customisable (that's ok for now I think)

app/views/govuk_component/docs/search.yml
```diff
@@ -0,0 +1,19 @@
...
+  default: {}
+  for_use_on_a_dark_but_not_black_background:
+    on_dark_background: true
+  change_label_text:
+    label_text: "Search all the things"
+  stop_the_label_appearing_inline:
+    inline_label: false
+  apply_custom_styling_to_the_label:
+    inline_label: false
+    label_text: "<span class=\"fixture-title\">Search this website</span>"
```
fofr: An example based on real world search results HTML would be useful:

```html
<h1>
Search results for <span class="visuallyhidden">“thing”</span>
</h1>
```

spec/javascripts/modules/toggle-class-on-focus.spec.js
```diff
@@ -0,0 +1,60 @@
+/* eslint-env jasmine */
```
fofr: What are these eslint vars doing?

app/views/govuk_component/search.raw.html.erb
```diff
@@ -0,0 +1,25 @@
...
+    <%= label_text %>
+  </label>
+  <div class="search-wrapper">
+    <input type="search" value="<%= value %>" id="<%= id %>" name="q" title="Search" class="search-element search-input js-class-toggle">
+    <div class="search-element search-submit-wrapper">
+      <button type="submit" class="search-submit">Search</button>
+    </div>
+  </div>
+</div>
+
```
fofr: Stray newline.

app/views/govuk_component/docs/search.yml
fofr: Perhaps a set of fixtures that represent usage in different contexts:

* on_homepage_on_govuk_blue
* in_header
* on_search_results
* on_manuals

app/assets/stylesheets/govuk-component/_search.scss
```diff
@@ -0,0 +1,184 @@
...
+      }
+    }
+  }
+
+  &.search-separate-label {
+    .search-label {
+      position: relative;
+      left: auto;
+
+      h1 {
```
fofr: There's no fixture that demonstrates this styling.

app/views/govuk_component/docs/search.yml
andysellick: This example is based on a real world use case where I intend the search component to eventually be used (https://www.gov.uk/search?q=test) where the page has a H1 is placed inside a label, but that's invalid, so I changed the example given from an included H1 to a span with the same classes, to simply demonstrate that markup can be included in the label if required. Example updated to be closer to a real world example.

andysellick: Re: fixtures, the component currently is only built for use on the homepage (although I have included options to make it suitable for use elsewhere), but yes, we should expand it to include further fixture options later.

app/views/govuk_component/docs/search.yml
andysellick: Have merged two of the examples and put it at the top.

spec/javascripts/modules/toggle-class-on-focus.spec.js
andysellick: I copied this from another test, turns out it's not needed so have removed.

app/views/govuk_component/search.raw.html.erb
andysellick: Removed.

app/assets/stylesheets/govuk-component/_search.scss
andysellick: There was previously (see discussion of H1 above) but now there isn't, good spot, removed.

app/views/govuk_component/docs/search.yml
andysellick: Didn't know there was a body: option, that's really useful. Have added.

andysellick: Had left the search input name as it is because I assumed the backend was expecting that.

fofr: Yep, that's fine. Need to document that decision in the component.

app/views/govuk_component/docs/search.yml
```diff
@@ -1,19 +1,22 @@
 name: "Search"
 description: "Search box"
+body: |
+  Component built primarily for the GOV.UK homepage (where it uses the on_dark_background option) but intended for use throughout the site.
+
+  Component provides only the search box itself and must be wrapped in a form.
```
fofr: Here's what I think the body should say:

> A search box with label and  attached submit button. The component must be used within an HTML form.
> The search input has a `name="q"` attribute and a customisable ID.
>
> It can be used on white or on `govuk-blue` using the `on_dark_background` option.

app/views/govuk_component/docs/search.yml
```diff
@@ -1,19 +1,22 @@
...
   default: {}
-  for_use_on_a_dark_but_not_black_background:
-    on_dark_background: true
-  change_label_text:
-    label_text: "Search all the things"
-  stop_the_label_appearing_inline:
+  apply_custom_styling_to_the_label_and_set_search_value:
     inline_label: false
-  apply_custom_styling_to_the_label:
+    label_text: "<span class=\"fixture-title\">Search results for <span class=\"visuallyhidden\">\"driving licence\"</span></span>"
```
fofr: Still using the `fixture-title` class which will not give an accurate representation of the component (as the guide styles this class).

`apply_custom_styling_to_the_label` – this component is not isolated.

I expected a fixture like the following, with the h1 styles still in the component:
```yml
search_results_with_h1_in_label:
  label_text: "Search results for <h1 class=\"visuallyhidden\">\"driving licence\"</span></h1>"
  value: "driving licence"
```

app/views/govuk_component/docs/search.yml
```diff
@@ -1,19 +1,22 @@
 name: "Search"
 description: "Search box"
+body: |
+  Component built primarily for the GOV.UK homepage (where it uses the on_dark_background option) but intended for use throughout the site.
```
fofr: Should be `on_govuk_blue`

app/views/govuk_component/search.raw.html.erb
```diff
@@ -1,6 +1,6 @@
 <%
   class_name = "on-light-background"
-  class_name = "on-dark-background" if local_assigns.include?(:on_dark_background)
+  class_name = "on-dark-background" if local_assigns.include?(:on_govuk_blue)
```
fofr: I guess the class name could change too? ¯\_(ツ)_/¯

spec/javascripts/modules/toggle-class-on-focus.spec.js
```diff
@@ -1,4 +1,3 @@
-/* eslint-env jasmine */
 /* eslint-disable no-multi-str */
```
fofr: Should the second one also be removed?

app/views/govuk_component/docs/search.yml
andysellick: Updated, thanks.

app/views/govuk_component/search.raw.html.erb
andysellick: Renamed class.

spec/javascripts/modules/toggle-class-on-focus.spec.js
andysellick: Yep, removed.

fofr:
:100: 👍  Nice work.

PR #1081, opened 2017-07-11
Fix print view of numbered lists

andysellick:
* Problem in print view when ordered lists have more than 9 items in them, only enough space on the left to display the last digit, so the numbering appears to reach 9 and then reset to 0
* Padding left increased to counter this problem, should now be good up to 999 items
* Issue documented here https://trello.com/c/CuYEun0D/110-content-list-print-view-content-list-numbering-10-issues

**Old**
screen shot 2017-07-11 at 13 27 31

**New**
screen shot 2017-07-11 at 13 27 44

### DIFF
--- a/app/assets/stylesheets/helpers/_print-base.scss
+++ b/app/assets/stylesheets/helpers/_print-base.scss
@@ -53,7 +53,7 @@ p {
 ul,
 ol {
   @include core-14;
-  padding-left: 13pt;
+  padding-left: 26pt;
 }
 
 ul {


app/assets/stylesheets/govuk-component/_govspeak-html-publication-print.scss
```diff
@@ -1,5 +1,3 @@
-.govuk-govspeak-html-publication {
-  .govuk-sticky-element {
-    display: none;
-  }
+.govuk-sticky-element {
```
nickcolley: This feels dangerous because now it's in the govspeak-html-publication-print.scss file but is a global selector. And it's not guaranteed we'll want to always hide anything using the sticky element feature.

andysellick: Yeah, problem is that the .govuk-govspeak-html-publication class doesn't exist anymore (I think it must have been removed during the visual rework recently).

Feel like this is turning into a wider issue so I'm going to remove this change entirely and create a new card to look at it separately.

nickcolley: Sweet, sounds good

fofr: I think the contents link in government-frontend needs a `dont-print` class added to. But I agree – it's separate to this PR.

PR #1084, opened 2017-07-14
Fixing misaligned warning sign on mobile

### LINKED ISSUE #1061: The warning sign on mobile on this start page is broken
image

https://www.gov.uk/book-driving-test

The bottom is cut off, and I'm not sure the alignment is correct.

vanitabarrett:
Fixes #1061 . Previously, the warning sign on transaction pages was misaligned on mobile due to the default margin being overridden by a separate style. It looks like this separate style was being wrongly applied to multiple elements when it should have just been applied to the first paragraph of any transaction page, to reduce the distance between the first heading and p.

This fix makes the CSS selector more specific to make sure this selector only applies as described above, and no longer effects warning signs.

## **Before:**
**Icon looks partially hidden. Looking more closely, we can see a margin-top: 0 is causing this.**

screen shot 2017-07-14 at 11 41 13
screen shot 2017-07-14 at 11 42 35

## **After:**
**The icon is now not hidden as the margin-top: 0 is no longer being applied.**

screen shot 2017-07-14 at 11 45 05

**We can also see that the margin-top: 0 is still being correctly applied to the first line of the page directly beneath the heading:**

screen shot 2017-07-14 at 11 46 13

### DIFF
--- a/app/assets/stylesheets/helpers/_text.scss
+++ b/app/assets/stylesheets/helpers/_text.scss
@@ -303,7 +303,7 @@ article {
     margin-bottom: 1.5em;
     padding-bottom: 1em;
 
-    p:first-child {
+    .get-started-intro > p:first-child {
       margin-top: 0;
     }
 


vanitabarrett:
@fofr Not sure if this CSS selector is too specific... although I haven't found anywhere where this seems to break anything, you might have a better idea of where this .get-started-intro class is used and how (if at all) outside transaction pages.

app/assets/stylesheets/helpers/_text.scss
```diff
@@ -303,10 +303,10 @@ article {
     margin-bottom: 1.5em;
     padding-bottom: 1em;
 
-    p:first-child {
+    .get-started-intro > p:first-child {
       margin-top: 0;
     }
-
+
```
fofr: Some rogue whitespace has been introduced here.
Is your editor configured to strip out trailing whitespace and to show invisibles?

vanitabarrett: @fofr It is now!

fofr:
Small change to fix whitespace, but otherwise :+1:. I think this CSS is specific because it needs to be.

PR #1089, opened 2017-07-19
Upgrade frontend toolkit to 7.0.1

vanitabarrett:
This version has a fix for overlapping volume controls in the media player (see PR https://github.com/alphagov/govuk_frontend_toolkit/pull/429).
Upgrade to 7.x.x also includes removal of multivariate test javascript (see PR https://github.com/alphagov/govuk_frontend_toolkit/pull/420)

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -85,8 +85,8 @@ GEM
     govuk-lint (0.6.1)
       rubocop (~> 0.35.0)
       scss_lint (~> 0.44.0)
-    govuk_frontend_toolkit (6.0.3)
-      rails (>= 3.1.0)
+    govuk_frontend_toolkit (7.0.1)
+      railties (>= 3.1.0)
       sass (>= 3.2.0)
     govuk_template (0.22.2)
       rails (>= 3.1)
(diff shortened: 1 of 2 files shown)


Gemfile.lock
```diff
@@ -281,4 +281,4 @@ DEPENDENCIES
   webmock
 
 BUNDLED WITH
-   1.14.5
+   1.15.1
```
vanitabarrett: @nickcolley Guessing I should change this back to the original?

fofr: My rule of thumb is keep the change if the new bundled version is higher.

fofr:
Could the 3 commits be squashed into a single commit?

Gemfile.lock
nickcolley: Asked in Slack:

Carlos:
"We have recently upgraded bundler in all applications to 1.15.1: github.com/alphagov/govuk-puppet/pull/6103/files

so when people run puppet on the local VM we will all be on the same version, so I think in this case it’s fine to leave the update in Gemfile.lock"

So seems good 👍

fofr:
(The links to the toolkit PRs in the description are pointing to the wrong place)

nickcolley:
Have we confirmed that the multivariate JavaScript is no longer being used?

vanitabarrett:
I'm pretty sure it's not being used

fofr:
Only known use of it was removed from government-frontend a while back: https://github.com/alphagov/government-frontend/pull/387

PR #1097, opened 2017-07-24
[DO NOT MERGE] Add custom dimensions with extra analytics

elenatanasoiu:
This will send along the first_published_at date with each pageview, in custom dimension 90.

### DIFF
--- a/app/assets/javascripts/analytics/custom-dimensions.js
+++ b/app/assets/javascripts/analytics/custom-dimensions.js
@@ -55,7 +55,8 @@
       'taxon-id': {dimension: 57, defaultValue: 'other'},
       'taxon-slugs': {dimension: 58, defaultValue: 'other'},
       'taxon-ids': {dimension: 59, defaultValue: 'other'},
-      'content-has-history': {dimension: 39, defaultValue: 'false'}
+      'content-has-history': {dimension: 39, defaultValue: 'false'},
+      'first-public-at': {dimension: 90, defaultValue: 'other'}
     };
 
     var $metas = $('meta[name^="govuk:"]');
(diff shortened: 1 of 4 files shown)


fofr:
Why is this marked as do not merge? Is it ready for review?

app/assets/javascripts/analytics/custom-dimensions.js
```diff
@@ -55,7 +55,8 @@
       'taxon-id': {dimension: 57, defaultValue: 'other'},
       'taxon-slugs': {dimension: 58, defaultValue: 'other'},
       'taxon-ids': {dimension: 59, defaultValue: 'other'},
-      'content-has-history': {dimension: 39, defaultValue: 'false'}
+      'content-has-history': {dimension: 39, defaultValue: 'false'},
+      'first-published-at': {dimension: 90, defaultValue: 'other'},
```
fofr: There's a stray javascript comma here. This will break in old IEs.

elenatanasoiu:
Yes, it's ready for review, but while I was testing on integration I noticed the `first-published-at` date is always: `2016-02-29T09:24:10.000+00:00` so I was investigating that. It might be because some of the pages I'm looking at have all been published on that date, but that sounds a bit unlikely.

fofr:
Thinking about this some more, there's risk associated with this because the content_item first published data isn't clean or consistent.

Consider:
https://trello.com/c/xCJ3RN6W/3-replace-firstpublicat-with-firstpublished-at
https://github.com/alphagov/government-frontend/blob/master/app/presenters/updatable.rb#L42-L44
https://github.com/alphagov/government-frontend/blob/master/app/presenters/specialist_document_presenter.rb#L164-L177

elenatanasoiu:
@fofr Closing this until the data being offered via `first_published_at` is cleaned up. More information in the Trello card: https://trello.com/c/7HwMeMfX/177-custom-dimension-first-published-date

PR #1099, opened 2017-07-25
Calculate how many links are on a finder

carvil:
This commit correctly sets dimension 27 with the number of links present
in a finder page such as `/government/policies`.

Trello: https://trello.com/c/vRtpDbnk/56-investigate-getting-nav-beta-analytics-tracking-onto-finder-template

Example on integration:

screen shot 2017-07-25 at 14 27 57


carvil:
That's a good point @whoojemaflip I will document this a bit better in a different PR and try to come up with a better way to write that logic, thanks for pointing it out.

PR #1100, opened 2017-07-26
[DO NOT MERGE] Send state and language to GA

elenatanasoiu:
For:
https://trello.com/c/zBBaVUGk/178-custom-dimension-publication-status
https://trello.com/c/cbycNx3y/201-custom-dimension-languages

## What this does

Sends the publishing state (withdrawn/not withdrawn) of the document and the language it's written in to Google Analytics. It will do this on every pageview event for publications, if it finds the state and/or the current language.

## Why 

The purpose is to gather more data about which documents are viewed the most when filtered by publishing state and language.

### DIFF
--- a/app/assets/javascripts/analytics/custom-dimensions.js
+++ b/app/assets/javascripts/analytics/custom-dimensions.js
@@ -46,6 +46,7 @@
       'political-status': {dimension: 7},
       'analytics:organisations': {dimension: 9},
       'analytics:world-locations': {dimension: 10},
+      'withdrawn': {dimension: 12, defaultValue: 'current'},
       'schema-name': {dimension: 17},
       'rendering-application': {dimension: 20},
       'navigation-page-type': {dimension: 32, defaultValue: 'none'},
@@ -78,6 +79,8 @@
         customDimensions['dimension' + dimension.dimension] = value;
       }
     });
+
+    customDimensions['dimension23'] = $('main[id="content"]').attr('lang') || 'unknown';
 
     return customDimensions;
   }
--- a/app/views/govuk_component/analytics_meta_tags.raw.html.erb
+++ b/app/views/govuk_component/analytics_meta_tags.raw.html.erb
@@ -9,6 +9,7 @@
 
   meta_tags["govuk:format"] = content_item_hash[:document_type] if content_item_hash[:document_type]
   meta_tags["govuk:schema-name"] = content_item_hash[:schema_name] if content_item_hash[:schema_name]
+  meta_tags["govuk:withdrawn"] = "withdrawn" if content_item_hash[:withdrawn_notice].present?
 
   organisations = []
   organisations += links_hash[:organisations] || []
--- a/spec/javascripts/analytics/ecommerce.spec.js
+++ b/spec/javascripts/analytics/ecommerce.spec.js
@@ -150,16 +150,18 @@ describe('Ecommerce reporter for results pages', function() {
       dimension11: '1',
       dimension3: 'other',
       dimension4: '00000000-0000-0000-0000-000000000000',
+      dimension12: 'current',
+      dimension23: 'unknown',
+      dimension26: '0',
+      dimension27: '0',
       dimension32: 'none',
       dimension33: 'thing',
       dimension34: 'other',
+      dimension39: 'false',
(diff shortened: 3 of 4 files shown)


app/views/govuk_component/analytics_meta_tags.raw.html.erb
```diff
@@ -20,6 +22,7 @@
   world_locations_content = world_locations.map { |link| "<#{link[:analytics_identifier]}>" }.join
   meta_tags["govuk:analytics:world-locations"] = world_locations_content if world_locations.any?
 
+
```
fofr: Some rogue whitespace here.

app/views/govuk_component/analytics_meta_tags.raw.html.erb
```diff
@@ -9,6 +9,8 @@
 
   meta_tags["govuk:format"] = content_item_hash[:document_type] if content_item_hash[:document_type]
   meta_tags["govuk:schema-name"] = content_item_hash[:schema_name] if content_item_hash[:schema_name]
+  meta_tags["govuk:publishing-state"] = details_hash[:publishing_state] if details_hash[:publishing_state]
+  meta_tags["govuk:current-language"] = content_item_hash[:locale] if content_item_hash[:locale]
```
fofr: Locale is already in the HTML, `<html lang="en">`. Can the javascript look for it from there instead of adding extra HTML to the page?

elenatanasoiu: I'm not sure how that is being populated. If I search for it here: https://github.com/alphagov/static/search?utf8=%E2%9C%93&q=lang%3D&type= 

I don't get any satisfying results, except for the print version. 

It might be that the content_item_hash["locale"] is more accurate than the <html lang="en">, unless there's another place I haven't noticed where the <html lang="en"> is being set.

app/views/govuk_component/analytics_meta_tags.raw.html.erb
elenatanasoiu: Thanks

app/views/govuk_component/analytics_meta_tags.raw.html.erb
fofr: Sorry, I pointed you at the wrong element. The HTML lang is usually `en` as parts of the page are always in English. However the `main` tag gets a lang attribute.

We set the lang on the `main` tag in government-frontend here:
https://github.com/alphagov/government-frontend/blob/master/app/views/layouts/application.html.erb#L25

Example:
https://www.gov.uk/government/news/foreign-secretary-statement-on-the-death-of-nelson-mandela.ru

elenatanasoiu: @fofr Taking language from `<main lang="">`. I've gone through all the documents this will be used for and they all seem to have the language provided so we can use this.

app/assets/javascripts/analytics/custom-dimensions.js
```diff
@@ -46,6 +46,7 @@
       'political-status': {dimension: 7},
       'analytics:organisations': {dimension: 9},
       'analytics:world-locations': {dimension: 10},
+      'publication-status': {dimension: 12, defaultValue: 'other'},
```
fofr: Should publication-status be "withdrawn".
Do we need the default value?

fofr: I believe publication-status is a separate thing to withdrawn.

Publication status would be:
* published
* force-published
* draft
* unpublished

Any of those could also be withdrawn/not-withdrawn.

elenatanasoiu: Ok in that case the publication-status would need to be sent separately. I've renamed the meta tag to `withdrawn`.

Regarding the default value, the original request was to be able to filter out the `withdrawn` documents. I've been told by the analysis team that they prefer to have a default value so that they can easily filter not only "withdrawn" docs, but the opposite as well. This is because GA has a habit of not showing the data at all if the custom dimension value is not set.

fofr: The downside to that is that we're sending more data with every GA request on every page for every user.

PR #1103, opened 2017-07-27
Refactor how we calculate section and link counts

carvil:
This commit moves the logic of determining how many sections and links
are on a given page to a new module, which has more granular functions
to extract all the necessary bits from a page to make a decision.

This refactoring aims at making it clearer what the steps are. Ideally,
each rendering application would set these counts in a meta tag and this
code would only read from it, but it's a bit more complicated than that,
especially for JS-driven finders.

What do you think?

### DIFF
--- a/app/assets/javascripts/analytics/page-content.js
+++ b/app/assets/javascripts/analytics/page-content.js
@@ -1,0 +1,120 @@
+(function () {
+  "use strict";
+  window.GOVUK = window.GOVUK || {};
+  var PageContent = function () { }
+
+  PageContent.getNumberOfSections = function () {
+    switch(true) {
+      case isNavigationGridPage():
+        return $('a[data-track-category="navGridLinkClicked"]').length;
+      case isNavigationAccordionPage():
+        return $('[data-track-count="accordionSection"]').length;
+      case isDocumentCollectionPage():
+        return $('.document-collection .group-title').length;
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/analytics/page-content.js
```diff
@@ -0,0 +1,120 @@
...
+        var sidebarSections = $('[data-track-count="sidebarRelatedItemSection"]').length;
+        var sidebarTaxons = $('[data-track-count="sidebarTaxonSection"]').length;
+
+        return sidebarSections || sidebarTaxons;
+    }
+  };
+
+  PageContent.getNumberOfLinks = function () {
+    switch(true) {
+      case isNavigationGridPage():
```
whoojemaflip: Is there any reason for this specific order that might not be obvious to the next person to touch this?

carvil: Good question. Before, the order was important because of document collections. They are both navigation pages and content pages, so the return statement needed a specific order. But with this `switch` statement, we look for navigation pages first (and they are findable in a deterministic way now since https://github.com/alphagov/whitehall/pull/3358), and default to a content page at the bottom. That means we don't rely on ordering anymore.

whoojemaflip: :+1:

PR #1104, opened 2017-07-28
Add button component

nickcolley:
Adds a button component to be used across applications

Note: We do not consume GOV.UK Elements directly due to the naming conventions being leaky, in time this component will be a wrapper for the [GOV.UK Frontend](https://github.com/alphagov/govuk-frontend) project's button component.
 
 Global button styles seen at `app/assets/stylesheets/helpers/_buttons.scss`
 should be considered deprecated with this component replacing any instances.
 
 govuk-* namespace replaced with pub-c-* otherwise we will conflict with https://github.com/alphagov/govuk-frontend/blob/master/docs/components.md

government-frontend PR using this component: https://github.com/alphagov/government-frontend/pull/418

| Before   |
|---|
| Big |
|   big |
| Not so big |
| lessbig |
| Pretty small |
| kindasmall |
| Small |
| small |

| After |
|---|
| Big |
| big |
| Medium |
| medium |
| Small |
| small |
| iPhone 6 iOS9 |
| iphone6 ios9 |
| Samsung Galaxy S7 |
| samsunggalaxy7 |
| Safari 6 |
| safari6 |
| Chrome 57 |
| chrome57 |
| Firefox 56 |
| firefox56 |
| Edge 15 |
| edge15 |
| IE11 |
| ie11 |
| IE10 |
| ie10 |
| IE9 |
| ie9 |
| IE8 |
| ie8 |
| IE7 |
| ie7 |
| IE6 |
| ie6 |

Also extends component for use in the govspeak component

Component Guide

screen shot 2017-08-01 at 13 15 21

Hardcoded button example

www.gov.uk/access-to-work/apply

screen shot 2017-08-01 at 13 14 47

### DIFF
--- a/app/assets/stylesheets/govuk-component/_button.scss
+++ b/app/assets/stylesheets/govuk-component/_button.scss
@@ -15,7 +15,7 @@
   background-repeat: no-repeat;
 
   @include media-down(mobile) {
-    background-position: 110% 50%;
+    background-position: center right -.35em;
   }
 
   @include ie(7) {
--- a/app/assets/stylesheets/govuk-component/_button.scss
+++ b/app/assets/stylesheets/govuk-component/_button.scss
@@ -1,5 +1,11 @@
 .pub-c-button {
   @include button;
+
+  @media(max-width: 425px) {
+    box-sizing: border-box;
+    width: 100%;
+    text-align: center;
+  }
 }
 
 .pub-c-button--start {
--- a/app/assets/stylesheets/govuk-component/_button.scss
+++ b/app/assets/stylesheets/govuk-component/_button.scss
@@ -1,3 +1,4 @@
+%pub-c-button,
 .pub-c-button {
   @include button;
 
@@ -8,6 +9,7 @@
   }
 }
 
+%pub-c-button--start,
 .pub-c-button--start {
   @include bold-24($line-height: (24 / 20));
   display: inline-block;
--- a/app/assets/stylesheets/govuk-component/_button.scss
+++ b/app/assets/stylesheets/govuk-component/_button.scss
@@ -11,11 +11,13 @@
 .pub-c-button--start {
   @include bold-24($line-height: (24 / 20));
   display: inline-block;
-  padding: 0.60em 1.7em 0.45em 0.67em;
+  padding: 0.6em 1.7em 0.45em 0.67em;
+
   @include media(tablet) {
     padding-top: 0.3em;
     padding-bottom: 0.15em;
   }
+
   background-image: image-url("icon-pointer.png");
   background-position: 100% 50%;
   background-repeat: no-repeat;
@@ -30,8 +32,10 @@
   }
 }
 
+// scss-lint:disable SelectorFormat
 .pub-c-button__info-text {
   display: block;
   margin-top: .5em;
   max-width: 14em;
 }
+// scss-lint:enable SelectorFormat
--- a/app/assets/stylesheets/govuk-component/_button.scss
+++ b/app/assets/stylesheets/govuk-component/_button.scss
@@ -3,9 +3,13 @@
 }
 
 .pub-c-button--start {
-  @include bold-24($line-height: (16 / 24));
+  @include bold-24($line-height: (24 / 20));
   display: inline-block;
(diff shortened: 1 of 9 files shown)


fofr:
@nickcolley Sass linter errors – need to bump gem and update some other rules.

app/assets/stylesheets/govuk-component/_button.scss
```diff
@@ -0,0 +1,41 @@
+.pub-c-button {
```
fofr: Do we need to document what `pub-c` means? As it stands this is different to the other static components.

app/views/govuk_component/docs/button.yml
```diff
@@ -0,0 +1,46 @@
...
+body: |
+  Button text should be short and describe the action the button performs.
+
+  GOV.UK Elements has more information on how buttons should be used https://govuk-elements.herokuapp.com/buttons/
+
+  Note: We do not consume GOV.UK Elements directly due to the naming conventions being leaky,
+  in time this component will be a wrapper for the [GOV.UK Frontend](https://github.com/alphagov/govuk-frontend) project's button component.
+
+  Global button styles seen at `app/assets/stylesheets/helpers/_buttons.scss`
+  should be considered deprecated with this component replacing any instances.
```
fofr: Good docs :100:

app/assets/stylesheets/govuk-component/_button.scss
fofr: Nevermind, you did this 👍

fofr:
* How should we add the `!important` for forcing button colour?
* How do we have a fixture showing specific link styles in container?

fofr:
Is there a branch showing usage on government-frontend?

nickcolley:
@fofr the color leakage is not an issue with this component since it sets color using `:link` which has a higher specificity than our leaky styles. 🙏 

(Would be great to have this as a test though)

nickcolley:
@fofr I've updated this to address the concerns around the govspeak button that'll be removed from government-frontend. (Note: I've tested static with master government-frontend, and the cascading overrides work out OK so this wont mean we need to co-ordinate a deployment)

test/govuk_component/button_test.rb
```diff
@@ -0,0 +1,43 @@
...
+  test "renders a button if href not set" do
+    render_component(text: "Start now")
+    assert_select "button.pub-c-button"
+    assert_select "a.pub-c-button", false
+  end
+
+  test "renders info text" do
+    render_component(text: "Start now", info_text: "Information text")
+    assert_select ".pub-c-button", text: "Start now"
+    assert_select ".pub-c-button__info-text", text: "Information text"
```
andysellick: Should we also test that the info-text element is not rendered unless the info_text parameter is included?

app/assets/stylesheets/govuk-component/govspeak/_button.scss
```diff
@@ -0,0 +1,12 @@
+
+/* im really sorry */
```
nickcolley: Accidentally committed that, I'm going to update with a better comment :P

app/views/govuk_component/docs/button.yml
```diff
@@ -0,0 +1,52 @@
...
+  (This is a challenge to the reader :))
+accessibility_criteria: |
+  The button must:
+
+  - accept focus
+  - be focusable with a keyboard
+  - indicate when it has focus
+  - activate when focused and space is pressed
+  - activate when focused and enter is pressed
+  - have a role of button
```
nickcolley: I took this as to say, the role is defined by default if you use a button element, whereas the role for an anchor needs to be added by specifying the role attribute.

test/govuk_component/button_test.rb
nickcolley: Added a test to check defaults 👍

app/views/govuk_component/docs/button.yml
andysellick: That makes sense, but might be worth clarifying.

nickcolley: I like @tvararu 's approach which is that the accessibility criteria does not refer to implementation details so I think this is fine for now.

PR #1106, opened 2017-07-28
Add user defined link text to emergency banner

bevanloon:
Emergency banners have the option of displaying a hyperlink for more information. The hyperlink text is currently hard coded to "More information" in the view. This PR makes the anchor text user definable.

* Add a parameter to the EmergencyBanner::Deploy class which writes the emergency banner data to Redis.
* Expose the new parameter that we store in Redis above via EmergencyBanner::Display. Also include a little logic to set a default value of "More information" since it isn't a mandatory field.
* Using the new "property" in EmergencyBanner::Display, show the anchor text in the view
* Update the rake task that we call from Jenkins to deploy the banner to include the new variable.

This should be released in conjunction with:

- [ ] [PR 6170 in GOV.UK Puppet](https://github.com/alphagov/govuk-puppet/pull/6170)
- [ ] [PR 352 in GOV.UK Developer Docs](https://github.com/alphagov/govuk-developer-docs/pull/352)

[Trello](https://trello.com/c/ZPkk7hnk/58-add-new-field-for-link-text)

### DIFF
--- a/lib/emergency_banner/display.rb
+++ b/lib/emergency_banner/display.rb
@@ -30,7 +30,15 @@ module EmergencyBanner
       content[:link] if content[:link].present?
     end
 
+    def link_text
+      return nil if link.blank?
+      return content[:link_text] if content[:link_text].present?
+      MORE_INFORMATION
+    end
+
   private
+
+    MORE_INFORMATION = "More information"
 
     def content
       return @content if defined? @content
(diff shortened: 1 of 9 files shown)


lib/emergency_banner/display.rb
```diff
@@ -30,8 +30,16 @@ def link
...
 
+    def link_text
+      return nil if link.blank?
+      return content[:link_text] if content[:link_text].present?
+      MORE_INFORMATION
+    end
+
   private
 
+    MORE_INFORMATION = "More information"
```
leenagupte: A very minor thing, but I probably would have named this `DEFAULT_LINK_TEXT`

PR #1108, opened 2017-08-03
Upgrade to Rails 5.1

fofr:
Questions:
* ~~We have some initialisers that remove XML and JSON parsers from the default parsers. These were added in 2013: https://github.com/alphagov/static/pull/177. Parsers have moved so this no longer works, I'm not sure whether they are still needed. Commented out for now.~~ Initializers removed in 9f689d6b224f3cf1f516140bbd2c7d0626d579de
* ~~Sprockets seems to be including asset digest URLs in tests now – I've updated the tests in 053e08a so they are less strict about what they expect but I'm not sure if this is a symptom of the slower sprockets on development.~~ Set digest assets to false in test environment (05bf7cb, based on http://railsdiff.org/4.2.7.1/5.1.2)

As part of the Rails 5 upgrade:
* Refactored the `root` routes. The `controller` method can no longer take two arguments. I believe the change I've added is equivalent.
* Remove quiet_assets gem, it's now config
* Switch from `render text:` and `render nothing:`, they were deprecated and now removed
* Use rack_strip_client_ip 0.0.2 for Rails 5 compatibility
* Update `actionpack-page_caching`to 1.1.0 for Rails 5 support
* Update airbrake to remove middleware errors (matching government-frontend version). Further upgrades possible.
* Stop pinning sprockets-rails
* Update nokogiri to avoid security vulns
* Update jasmine-rails for Rails 5

In 6976fc0 I've also updated a lot of the Rails boilerplate code to be more modern. Though I've kept this intentionally minimal.

The two Snyk vulnerabilities have no known fixes.


Gemfile
```diff
@@ -1,19 +1,18 @@
...
 
 gem 'logstasher', '0.4.8'
-gem 'rack_strip_client_ip', '0.0.1'
-gem 'actionpack-page_caching', '1.0.2'
+gem 'rack_strip_client_ip', '0.0.2'
+gem 'actionpack-page_caching', '1.1.0'
 
 gem 'uglifier', ">= 1.3.0"
 gem 'sass-rails', "5.0.6"
-gem 'airbrake', '~> 4.3.1'
```
h-lame: In most apps we've been using our own fork of airbrake 4 to get rails 5 compatibility and avoid losing some functionality on our errbit install (see: https://docs.publishing.service.gov.uk/manual/upgrade-rails.html#airbrake)

fofr:
We added `gem 'actionpack-page_caching', '1.1.0'` in the Rails 4 upgrade, but it looks like we haven't included the config it needs: https://github.com/rails/actionpack-page_caching#usage

Gemfile
fofr: I've updated in 60798cd and removed 3e6b9d1

h-lame: :+1:

fofr:
Removed old initialisers for XML/JSON after discussing with @boffbowsh.

Sprockets slow downs will not affect production and can be investigated separately to this PR. I will track with an issue.

Gemfile
boffbowsh: So excited to switch this to Sentry soon...

fofr:
@boffbowsh It's on integration now if you wanted to give it a once-over.

boffbowsh:
You'll be able to test it better than me, @fofr

fofr:
I can't see the 3rd issue, only 2 are showing on Snyk itself.
screen shot 2017-08-09 at 15 26 28

binaryberry:
Ah ok. Waiting to get access to Snyk myself, so wasn't able to click through to the details. Interesting there is a discrepancy, but shouldn't be a blocker. I've approved now!

PR #1114, opened 2017-08-17
Send the client ID inside a custom dimension

elenatanasoiu:
For: https://trello.com/c/n22tcj9Q/206-custom-dimension-google-analytics-client-id-and-session-id

In order to track user journeys in GA, it's been requested that we send the client ID as a custom dimension. The docs for extracting the client id can be found here: https://developers.google.com/analytics/devguides/collection/analyticsjs/cookies-user-id?hl=en#getting_the_client_id_from_the_cookie

Using the recommended way means we're expecting a tracker to be loaded, which may or may not have happened at that point. Because our method is asynchronous, it might be called before `trackPageView` and not have a tracker object to query for the client ID.

In order to avoid this, we've put the call for `trackPageView` inside the asynchronous method for setting the client ID. This will guarantee that these two methods are called in the right order.

We've also had to move the rest of the plugin loading after the `trackPageView` event in order to ensure the same order will be kept.

### DIFF
--- a/app/assets/javascripts/analytics/static-analytics.js
+++ b/app/assets/javascripts/analytics/static-analytics.js
@@ -10,17 +10,23 @@
 
     var trackingOptions = getOptionsFromCookie();
 
-    // Track initial pageview
-    this.trackPageview(null, null, trackingOptions);
+    // We're setting the client ID inside a callback function because
+    // we don't have access to the client ID until GA returns a tracker object.
+    ga(function (tracker) {
+      this.gaClientId = tracker.get('clientId');
 
-    // Begin error and print tracking
-    GOVUK.analyticsPlugins.error({filenameMustMatch: /gov\.uk/});
-    GOVUK.analyticsPlugins.printIntent();
-    GOVUK.analyticsPlugins.mailtoLinkTracker();
-    GOVUK.analyticsPlugins.externalLinkTracker();
-    GOVUK.analyticsPlugins.downloadLinkTracker({
-      selector: 'a[href*="/government/uploads"], a[href*="assets.publishing.service.gov.uk"]'
-    });
+      // Track initial pageview
+      this.trackPageview(null, null, trackingOptions);
+
+      // Begin error and print tracking
+      GOVUK.analyticsPlugins.error({filenameMustMatch: /gov\.uk/});
+      GOVUK.analyticsPlugins.printIntent();
+      GOVUK.analyticsPlugins.mailtoLinkTracker();
+      GOVUK.analyticsPlugins.externalLinkTracker();
+      GOVUK.analyticsPlugins.downloadLinkTracker({
+        selector: 'a[href*="/government/uploads"], a[href*="assets.publishing.service.gov.uk"]'
+      });
+    }.bind(this));
   };
 
   StaticAnalytics.load = function () {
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/analytics/static-analytics.js
```diff
@@ -10,17 +10,23 @@
...
-    GOVUK.analyticsPlugins.error({filenameMustMatch: /gov\.uk/});
-    GOVUK.analyticsPlugins.printIntent();
-    GOVUK.analyticsPlugins.mailtoLinkTracker();
-    GOVUK.analyticsPlugins.externalLinkTracker();
-    GOVUK.analyticsPlugins.downloadLinkTracker({
-      selector: 'a[href*="/government/uploads"], a[href*="assets.publishing.service.gov.uk"]'
-    });
+    // We're setting the client ID inside a callback function because
+    // we don't have access to the client ID until GA returns a tracker object.
+    ga(function (tracker) {
```
fofr: What's the risk of nesting these calls within an anonymous function called by ga?

elenatanasoiu: This anonymous function is called as part of the init process where we create the original analytics object. If it didn't get a response, we wouldn't have a GoogleUniversalAnalyticsTracker with which to send data in the first place. So the setup process could not finish without us getting this callback.  

It was inspired by this commit, where the client ID needs to be set before the call to GA via the sendData method: https://github.com/alphagov/govuk_frontend_toolkit/commit/8bd1360d9f2e6a8966824dd67f4935598c0d4e02#diff-7f6da3f6d8cbd22c94ad731b283e9226

nickcolley: I thought we abstracted Google Analytics so we're not interacting with it directly? 🤔

elenatanasoiu: Sorry, I didn't see your review here. 

Yes, you're correct, we aren't interacting with it directly, which is why I've had to do this workaround in order to obtain the client id since it's not being set in static. It's being set in `govuk_frontend_toolkit` where we communicate with GA through 2 trackers: GOVUKTracker and GoogleAnalyticsUniversalTracker.

So we've had to wait for the tracker to be returned to us (in this case GoogleAnalyticsUniversalTracker) so that we could access the client id.

nickcolley: I think we need to follow up, at some point the trackers were intentionally abstracted to avoid a direct link to Google Analytics.

The ga global is Google Analytics which this file shouldnt know about - given the above?

elenatanasoiu: It's already using global.ga inside the init file: https://github.com/alphagov/static/blob/8ab43d8348dcd264302ba4c94ee16c0ba5344ee3/app/assets/javascripts/analytics/init.js.erb#L25

It's receiving global as a param in there (on line 1) so even if we removed the use of ga in the `static-analytics.js` code, we'd still be able to access it because we're receiving ga through the global param in the init file. 

What is the purpose of removing a direct link to Google Analytics?

nickcolley: At some point GOV.UK abstracted the analytics so in theory you could swap out Google Analytics and it'd still work.

@fofr may be able to give more context

fofr: Yeah, I think this breaks that concept.

PR #1116, opened 2017-08-18
Update pagination component

andysellick:
- to use CSS class naming convention for components
- adds accessibility criteria to documentation
- tests for navigation role

https://trello.com/c/3rWOvqBG/95-2-update-pagination-component-to-be-inline-with-component-principles

Should look exactly the same as the current one, but here's some screenshots for examples.

screen shot 2017-08-21 at 08 59 42

screen shot 2017-08-21 at 09 00 11

See also related but non-dependent PR: https://github.com/alphagov/finder-frontend/pull/301

### DIFF
--- a/app/assets/stylesheets/govuk-component/_previous-and-next-navigation.scss
+++ b/app/assets/stylesheets/govuk-component/_previous-and-next-navigation.scss
@@ -1,81 +1,83 @@
-.govuk-previous-and-next-navigation {
+.pub-c-pagination {
   display: block;
   margin-top: $gutter;
   margin-bottom: $gutter;
   margin-left: -$gutter-half;
   margin-right: -$gutter-half;
+}
 
-  ul {
-    margin: 0;
-    padding: 0;
-  }
+// scss-lint:disable SelectorFormat
+.pub-c-pagination__list {
+  margin: 0;
+  padding: 0;
 
-  li {
-    @include core-16($line-height: (20 / 16));
-    float: left;
-    list-style: none;
-    text-align: right;
-    margin: 0;
-    padding: 0;
-    width: 50%;
-
-    a {
-      display: block;
-      padding: $gutter-half;
-      text-decoration: none;
-
-      &:visited {
-        color: $link-colour;
-      }
-
-      &:hover,
-      &:active {
-        background-color: $canvas-colour;
-      }
-
-      .pagination-part-title {
-        @include core-27($line-height: (33.75 / 27));
-        display: block;
-      }
-    }
-  }
-
-  .previous-page {
-    float: left;
-    text-align: left;
-  }
-
-  .next-page {
-    float: right;
-    text-align: right;
-  }
-
-  @include media-down(mobile) {
-    .previous-page,
-    .next-page {
-      float: none;
-      width: 100%;
-    }
-
-    .next-page a {
-      text-align: right;
-    }
-  }
-
-  .pagination-icon {
-    display: inline-block;
-    margin-bottom: 1px;
-    height: .482em;
-    width: .63em;
-  }
-
-  .pagination-label {
-    display: inline-block;
-    margin-top: 0.1em;
-    text-decoration: underline;
-
-    &:empty {
-      display: none;
-    }
+  &:after {
+    content: "";
+    display: block;
+    clear: both;
   }
 }
+
+.pub-c-pagination__item {
+  @include core-16($line-height: (20 / 16));
+  float: left;
+  list-style: none;
+  text-align: right;
+  margin: 0;
+  padding: 0;
+  width: 50%;
+}
+
+.pub-c-pagination__link {
+  display: block;
+  padding: $gutter-half;
+  text-decoration: none;
+
+  &:visited {
+    color: $link-colour;
+  }
+
+  &:hover,
+  &:active {
+    background-color: $canvas-colour;
+  }
+}
+
+.pub-c-pagination__link-title {
+  @include core-27($line-height: (33.75 / 27));
+  display: block;
+}
+
+.pub-c-pagination__item--previous {
+  @include media-down(mobile) {
+    float: none;
+    width: 100%;
+  }
+
+  float: left;
+  text-align: left;
+}
+
+.pub-c-pagination__item--next {
+  @include media-down(mobile) {
+    float: none;
+    width: 100%;
+  }
+
+  float: right;
+  text-align: right;
+}
+
+.pub-c-pagination__link-icon {
+  display: inline-block;
+  margin-bottom: 1px;
+  height: .482em;
+  width: .63em;
+}
+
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/govuk-component/_previous-and-next-navigation.scss
```diff
@@ -1,81 +1,81 @@
-.govuk-previous-and-next-navigation {
+.app-c-pagination {
```
nickcolley: Think this should be `pub` since it's in static?

app/views/govuk_component/docs/previous_and_next_navigation.yml
```diff
@@ -11,6 +11,25 @@ body: |
...
 
   If one of the 2 parameters is nil, no link will appear.
+accessibility-criteria: |
+  The component must:
+
+  - identify itself as pagination navigation
+
+  Links in the component must:
+
+  - include rel next or rel previous
```
nickcolley: This seems like an implementation detail, is it used in an accessibility context?

app/assets/stylesheets/govuk-component/_previous-and-next-navigation.scss
```diff
@@ -1,81 +1,81 @@
...
-    float: right;
-    text-align: right;
-  }
+.app-c-pagination__link-title {
+  @include core-27($line-height: (33.75 / 27));
+  display: block;
+}
 
+.app-c-pagination__item--previous {
   @include media-down(mobile) {
```
nickcolley: Reordering the media queries does this mess with any specificity?

app/assets/stylesheets/govuk-component/_previous-and-next-navigation.scss
```diff
@@ -1,81 +1,81 @@
...
 
-  .pagination-label {
-    display: inline-block;
-    margin-top: 0.1em;
-    text-decoration: underline;
+  float: right;
+  text-align: right;
+}
 
-    &:empty {
```
nickcolley: We've removed this, do we know what the impact is?

app/assets/stylesheets/govuk-component/_previous-and-next-navigation.scss
```diff
@@ -1,81 +1,83 @@
...
-    float: right;
-    text-align: right;
-  }
+.pub-c-pagination__link-title {
+  @include core-27($line-height: (33.75 / 27));
+  display: block;
+}
 
+.pub-c-pagination__item--previous {
   @include media-down(mobile) {
```
nickcolley: Reordering media queries could mess with specificity, is this still alright?

app/assets/stylesheets/govuk-component/_previous-and-next-navigation.scss
andysellick: This element isn't rendered if it doesn't have content (it's wrapped in an if) so this rule seemed redundant.

app/assets/stylesheets/govuk-component/_previous-and-next-navigation.scss
andysellick: Tested at various screen sizes, seems to behave exactly as the current version.

app/views/govuk_component/docs/previous_and_next_navigation.yml
andysellick: That's a fair point, I'll try rewording it.

app/assets/stylesheets/govuk-component/_previous-and-next-navigation.scss
```diff
@@ -1,81 +1,83 @@
-.govuk-previous-and-next-navigation {
+.pub-c-pagination {
   display: block;
   margin-top: $gutter;
   margin-bottom: $gutter;
   margin-left: -$gutter-half;
   margin-right: -$gutter-half;
```
vanitabarrett: Can the margins be simplified to: `margin: $gutter (-$gutter-half)`

fofr:
We need to check no apps are abusing components by overriding internal classes. Are there references to `govuk-previous-and-next-navigation` anywhere outside of static?

app/assets/stylesheets/govuk-component/_previous-and-next-navigation.scss
nickcolley: I prefer long hand since it's super obvious what's happening but I think I'm in the minority.

I compare it to options objects passed to a function

`function render('potato', true, false, true, false)`

`function render('potato', {
  upsideDown: true,
  invisible: false,
  superCool: true
})`

nickcolley:
https://govuk-publishing-frontend-search.glitch.me/?q=govuk-previous-and-next-navigation

app/assets/stylesheets/govuk-component/_previous-and-next-navigation.scss
vanitabarrett: @nickcolley I'd argue the margin scenario is a bit different because it's always the same, i.e: `margin: top right bottom left`, whereas with an options object there's no "standard' parameters to be passing, so long hand is more important then? But I guess it's personal preference 🙂

nickcolley: It's less about them always being the same and more about having to understand how the function expects something to be passed.

With long hand you know exactly how it'll behave without having to consult the docs if you forget.

If you put the two versions in front of developers, there will be a lot of people who couldnt tell you what the short hand is doing whereas everyone will be able to tell you what the long hand does.

I like explicit code that doesnt require someone from having to consult documentation.

nickcolley:
Looks like here:
https://github.com/alphagov/finder-frontend/blob/f11df224c06d9d80d7bd9372450b3faee38e2c41/app/assets/stylesheets/finder_frontend.scss#L334

It'd be good to update finder-frontend so it has this spacing without overriding the component.

This could be done by adding spacing on something above it.

andysellick:
I've just fired up finder-frontend and discovered that the local CSS change overrides the margin top of the prev/next with... wait for it... exactly the same margin top. I'll open a separate PR to remove that :)

app/assets/stylesheets/govuk-component/_previous-and-next-navigation.scss
nickcolley: I wrote this but didn't properly document why (I'm a bad person), I think this might be a workaround for an issue in one of the applications.

My gut was that it was something to do with manuals: https://www.gov.uk/hmrc-internal-manuals/admin-law-manual/adml4000 since they don't have labels, but you're right they dont render at all.

Let's just be extra careful when we deploy to check all the places this is used.

https://govuk-publishing-frontend-search.glitch.me/?q=govuk_component/previous_and_next_navigation

PR #1117, opened 2017-08-21
Update title component to meet chosen conventions

vanitabarrett:
Updating static title component to meet government-frontend component conventions.
This includes:
* Using pub-c namespace for CSS classes
* Use of BEM
* Adding accessibility criteria to docs

screen shot 2017-08-21 at 09 49 50

screen shot 2017-08-21 at 09 50 03

### DIFF
--- a/app/assets/stylesheets/govuk-component/_title.scss
+++ b/app/assets/stylesheets/govuk-component/_title.scss
@@ -16,3 +16,21 @@
   @include bold-36;
 }
 // scss-lint:enable SelectorFormat
+
+// To Be Deleted: Old Styling
+.govuk-title {
+  @include responsive-vertical-margins;
+
+  .context {
+    @include core-24;
+    color: $secondary-text-colour;
+  }
+
+  h1 {
+    @include bold-48;
+  }
+
+  &.length-long h1 {
+    @include bold-36;
+  }
+}
--- a/app/assets/stylesheets/govuk-component/_title.scss
+++ b/app/assets/stylesheets/govuk-component/_title.scss
@@ -1,16 +1,18 @@
-.govuk-title {
+// scss-lint:disable SelectorFormat
+.pub-c-title {
   @include responsive-vertical-margins;
-
-  .context {
-    @include core-24;
-    color: $secondary-text-colour;
-  }
-
-  h1 {
-    @include bold-48;
-  }
-
-  &.length-long h1 {
-    @include bold-36;
-  }
 }
+
+.pub-c-title__context {
+  @include core-24;
+  color: $secondary-text-colour;
+}
+
+.pub-c-title__text {
+  @include bold-48;
+}
+
+.pub-c-title__text--long {
+  @include bold-36;
+}
+// scss-lint:enable SelectorFormat
--- a/app/views/govuk_component/title.raw.html.erb
+++ b/app/views/govuk_component/title.raw.html.erb
@@ -1,10 +1,12 @@
 <%
-  average_title_length ||= :average
+  average_title_length ||= false
   context ||= false
 %>
-<div class="govuk-title length-<%= average_title_length %>">
+<div class="pub-c-title">
   <% if context %>
-    <p class="context"><%= context %></p>
+    <p class="pub-c-title__context"><%= context %></p>
   <% end %>
-  <h1><%= title %></h1>
+  <h1 class="pub-c-title__text <% if average_title_length %>pub-c-title__text--<%= average_title_length %><% end %>">
+    <%= title %>
+  </h1>
 </div>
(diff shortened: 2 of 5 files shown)


app/assets/stylesheets/govuk-component/_title.scss
```diff
@@ -1,16 +1,18 @@
-.govuk-title {
+// scss-lint:disable SelectorFormat
+.pub-c-title {
   @include responsive-vertical-margins;
 
-  .context {
-    @include core-24;
-    color: $secondary-text-colour;
+  &--long .pub-c-title__text {
```
nickcolley: We try to avoid nesting if it's not required see: https://github.com/alphagov/govuk-frontend/blob/master/docs/coding-standards/css.md#nesting

app/views/govuk_component/title.raw.html.erb
```diff
@@ -2,9 +2,9 @@
   average_title_length ||= :average
   context ||= false
 %>
-<div class="govuk-title length-<%= average_title_length %>">
+<div class="pub-c-title pub-c-title--<%= average_title_length %>">
```
nickcolley: I think this means if the long title length is not active we'll have a class of "pub-c-title pub-c-title--" which is not a big deal but seems like it could be avoided?

test/govuk_component/title_test.rb
```diff
@@ -13,11 +13,16 @@ def component_name
...
+    assert_select ".pub-c-title__text", text: "Hello World"
   end
 
   test "title context appears" do
     render_component(title: "Hello World", context: "Format")
-    assert_select ".govuk-title .context", text: "Format"
+    assert_select ".pub-c-title__context", text: "Format"
+  end
+
+  test "applies title length if supplied" do
```
nickcolley: 👍  Nice

app/assets/stylesheets/govuk-component/_title.scss
nickcolley: Seems as if here, since the long modifier only impacts the text element, we could have a modifier class on `.pub-c-title_text` instead?

app/views/govuk_component/title.raw.html.erb
fofr: It defaults to 'average', no?

nickcolley: Yeah think you're right, not sure why though? 

See https://github.com/alphagov/static/pull/1117/files/71356f9b255fe913d6403bdd5006b4633b45037d#r134181601

nickcolley:
[Searching the codebase for any overrides](https://govuk-publishing-frontend-search.glitch.me/?q=govuk-title) I found https://github.com/alphagov/collections/blob/cbaae81eed6cf2dac1b4b7885e864d5f5d1e024d/app/views/taxons/_page_header.html.erb#L3 so we need to update collections to use the govuk-title component and then we can ship this change.

vanitabarrett:
@nickcolley I've had a look at Collections in more detail. It looks like there's about 4 different titles being used on different pages, all with different spacing and uses, among themselves and with our component (see screenshots below, left is current collections title, right is with static title component).

**Spacing differences:**
screen shot 2017-08-21 at 15 19 22

**This title also contains a description, which our title component doesn't take as a parameter.**
screen shot 2017-08-21 at 15 25 45

If we want to integrate the title component, we'd probably have to override some of the component CSS within collections, which doesn't feel like a good thing to be do? What are your thoughts on where to go from here?

fofr:
@vanitabarrett @nickcolley Should we iterate collections to _not_ use the title component?

vanitabarrett:
@fofr @nickcolley The collections title does feel different to the static title - perhaps in the future it should really be an app-specific component in itself (if other apps move over to our component pattern). For now we could remove the govuk-title so it doesn't rely on the static styling, and add the styling it needs into collections?

nickcolley:
I've not closely looked but the title component should be broadly applicable, seems like the main problem is spacing which is something we know we need to solve.

nickcolley:
@vanitabarrett I think we can avoid the headings that are not currently using the govuk-title classes and create a card to consolidate them separately.

I've had a look at the remaining use cases.

That text underneath the title, looks a lot like our lead paragraph component.

screen shot 2017-08-23 at 16 12 09

This means if we can move the lead paragraph component to static, and allow flexible spacing for the title component. We should be good to go?

nickcolley:
screen shot 2017-08-23 at 16 17 54

Screenshot showing the above hacked in

vanitabarrett:
To get this story completed more quickly, we've decided to (for now) remove the use of the govuk-title class in Collections and simply copy over the styling. This has been done in https://github.com/alphagov/collections/pull/373

Have added stories to the Planning board to [move over lead paragraph to static](https://trello.com/c/I99hvEOq/219-move-lead-paragraph-component-from-government-frontend-to-static) and [update collection titles to use static components](https://trello.com/c/ec0YjgQ4/220-update-titles-in-collections-to-use-title-and-lead-paragraph-components-in-static)

fofr:
We need to deploy Collections before merging this.

PR #1120, opened 2017-08-25
Remove old govuk-title and previous and next styling

vanitabarrett:
This needs removing now the title has been updated to use namespace pub-c- and BEM


fofr:
Can you remove pagination styles too?

PR #1122, opened 2017-08-25
Add new user satisfaction survey

h-lame:
For: https://trello.com/c/b4J4OpaM/218-emotional-survey-test-banner  This survey is to run 1 in 6 frequency, on all pages (hence the lack of any `activeWhen` arguments in the definition) and only last for one week (Thursday -> Monday).


h-lame:
Flagged as DNM as I just need to confirm the dates for the survey with Sean.

h-lame:
Removed DNM as confirmed dates with Sean.

PR #1118, opened 2017-08-24
Improve accessibility of pagination component

andysellick:
- add a visually hidden element containing a colon so screenreaders pause between the navigation part of the link (ie. next/previous) and the label part, if present
- update accessibility criteria and test to include this

https://trello.com/c/0kYwSKe4/120-investigate-accessibility-of-next-previous-component

PR #1123, opened 2017-08-30
[Do not merge] Improve next previous accessibility

andysellick:
- hidden pseudo before on label element containing "." causes screen readers to take a breath between navigation and label parts of link
- tested on Voiceover (Mac), Narrator, Jaws 18, NVDA (Windows)
- unable to test for pseudo element content so specific tests for accessibility removed, but comment added to SCSS to prevent accidental removal

screen shot 2017-08-30 at 09 29 17

See https://trello.com/c/0kYwSKe4/222-investigate-accessibility-of-next-previous-component

### DIFF
--- a/app/assets/stylesheets/govuk-component/_previous-and-next-navigation.scss
+++ b/app/assets/stylesheets/govuk-component/_previous-and-next-navigation.scss
@@ -79,5 +79,16 @@
   display: inline-block;
   margin-top: 0.1em;
   text-decoration: underline;
+
+  // this pseudo element makes screen readers pause, preventing navigation and label text confusion
+  &:before {
+    content: ".";
+    position: absolute;
+    left: -9999em;
+    top: auto;
+    width: 1px;
+    height: 1px;
+    overflow: hidden;
+  }
 }
 // scss-lint:enable SelectorFormat
--- a/app/views/govuk_component/previous_and_next_navigation.raw.html.erb
+++ b/app/views/govuk_component/previous_and_next_navigation.raw.html.erb
@@ -15,7 +15,6 @@
             <%= previous_page[:title] %>
           </span>
           <% if previous_page[:label].present? %>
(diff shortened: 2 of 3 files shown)


nickcolley:
This doesn't feel right to me, since it could change at any time and I'm not sure the current version without pauses is a barrier.

The advice in https://accessibility.blog.gov.uk/2017/02/08/advice-for-creating-content-that-works-well-with-screen-readers recommends not do anything specific for assistive tech but this change does seem to make it clearer.

aduggin:
I agree with Nick. Do we know if the lack of a pause is a problem? Can you pose the question in the accessibility community google group- would be good to get some thoughts/feedback from screen reader users.  Can you set up a couple of couple of examples using something like CodePen so that people can test out a rendered version and provide feedback.

fofr:
This is a follow on from #1118 which introduced a colon in an attempt to add a pause.

selfthinker:
To summarise what I said in our internal Slack channel:

I agree that having a link which reads "Previous Terrorism" is not great. I think the best solution is exactly what #1118 was doing, i.e. adding a colon in between, because "Previous: Terrorism" is way more semantic. That also works well when you switch off CSS. And it works for Braille display users (the solution in this PR doesn't).

This PR however is more of a hack to fix various browser or screen reader bugs. I think that's generally a bad idea. And we cannot be sure what other issues it might introduce, especially with different screen reader settings and other speech engines.
That's why I would follow Leonie's advice from the blog post Nick linked above:
> don't write content that works specifically for screen readers, write content that works well for everyone. Use correct punctuation, spelling and grammar, use standard conventions for acronyms and abbreviations, and use words that are appropriate for your audience.

If you want to make sure you're doing the right thing there is no way around asking actual screen reader users, either by asking the cross-government accessibility mailinglist or doing some user research. Because everything else is just guesswork.

PR #1125, opened 2017-08-31
Use govuk_publishing_components to create component guide

fofr:
* Update YAML to match govuk_publishing_components (v 1.0)
* Use govuk_publishing_components component guide

This branch is deployed to Heroku and the component guide is available to view here:
https://govuk-static.herokuapp.com/component-guide/

---

Slimmer is used by the govuk_publishing_components gem, however it inserts itself automatically as middleware in the host Rails application.

* Disable Slimmer middleware for Static and enable for use in the
context of the gem’s engine only
* Inserting slimmer middleware into static risks the app getting stuck
in a loop requesting things from itself over and over.
* The `skip_slimmer` approach wasn’t foolproof


h-lame:
The publishing components gem uses slimmer, which ultimately means that the `/component-guide` pages rendered by static are going to make a network call out and back into itself to fetch the templates.  This is pretty weird, and the reason we needed to disable the slimmer middleware.  We probably want to investigate a way of requiring slimmer that doesn't auto-glom it onto the middleware stack, and maybe a way of configuring it to fetch templates from disk instead of via the network for edgecases like this.  However, the approach taken here is the right one until we do that.

andysellick:
Should we add a link to the component guide to the readme?

fofr:
@andysellick Added in 8c3d58e

doc/govuk-components.md
```diff
@@ -1,124 +0,0 @@
...
-
-* a [Partial View](../app/views/govuk_component) - The template logic and markup, also defines the arguments expected
-* a [SCSS module](../app/assets/stylesheets/govuk-component) - The styling of the component
-* a [Javascript module](../app/assets/javascripts/govuk-component) - JS behaviour of the component
-* [Documentation](../app/views/govuk_component/docs) - a `.yml` per component, describing the component and containing fixture data.
-* a [unit test](../test/govuk_component) - testing the component renders correctly based on parameters passed in
-
-## Principles behind components
-
-### Encapsulate common UI
```
nickcolley: @fofr is this transferred or have an equivalent in our new docs? this is pretty good stuff...

fofr: Yeah, I thought the same. We should transfer to gem.

config/application.rb
```diff
@@ -47,5 +47,13 @@ class Application < Rails::Application
     config.action_dispatch.rack_cache = nil
 
     config.paths["log"] = ENV["LOG_PATH"] if ENV["LOG_PATH"]
+
+    # Slimmer is used by the govuk_publishing_components gem, however it inserts itself
+    # automatically as middleware in the host Rails application
+    # Disable Slimmer middleware for Static and enable for use in gem only
+    if defined?(GovukPublishingComponents) && defined?(Slimmer)
```
alecgibson: I don't have a great understanding of Railties - is this config always guaranteed to be run after the Slimmer railtie has injected itself into the middleware? If so, why do we need to check `defined?(Slimmer)`? (and `defined?(GovukPublishingComponents)` at that rate).

config/application.rb
```diff
@@ -47,5 +47,13 @@ class Application < Rails::Application
...
+      config.middleware.delete Slimmer::App
+      GovukPublishingComponents::Engine.config.middleware.use Slimmer::App
```
alecgibson: Again, my knowledge of Railties is basically nil, but would a conceptually neater approach to this be configuring `Slimmer` with which middleware to target (if possible)?

config/initializers/govuk_publishing_components.rb
```diff
@@ -0,0 +1,12 @@
+if defined?(GovukPublishingComponents)
```
alecgibson: Again, I find it unclear in which situations this shouldn't be defined.

config/initializers/govuk_publishing_components.rb
```diff
@@ -0,0 +1,12 @@
...
+  GovukPublishingComponents.configure do |c|
+    c.component_guide_title = "Static Component Guide"
+    c.static = true
+
+    # Component guide already includes static, which includes the components
```
alecgibson: This sounds pretty circular?

config/initializers/govuk_publishing_components.rb
alecgibson: Oh, is it only defined in Development?

nickcolley: Yes, we only include this Gem in Development and on Heroku

config/initializers/govuk_publishing_components.rb
fofr: Aye, that's the problem this PR has to solve/work around. As @h-lame nicely put it, the shape of the world we've built isn't quite right.

config/application.rb
fofr: We need to check that Slimmer is defined in order to access `Slimmer::App`
We need to know if GovukPublishingComponents is defined to add the slimmer middleware to the engine.

There isn't an easy way to query if some middleware exists.

config/application.rb
fofr: Yes, I agree it would be neater. I decided against this because:

* it required a sizeable amount of refactoring in slimmer
* slimmer hasn't really been worked on fundamentally since 2013
* we aren't sure about the usefulness or shape of slimmer
* the refactor would only be necessary for this one use

config/application.rb
alecgibson: Yes, but what I mean is - shouldn't we _ensure_ we run this after `Slimmer` has been railtied in? If it's not, then surely the component guide won't work properly? The fact that you have to check if it's been loaded is a smell that says we don't understand the load order properly, which probably isn't great.

config/application.rb
alecgibson: Sure fine by me.

doc/govuk-components.md
nickcolley: https://github.com/alphagov/govuk_publishing_components/issues/46

config/application.rb
fofr: We require the gems before configuring the app:
https://github.com/alphagov/static/blame/use-guide-gem/config/application.rb#L9

The publishing components gem requires slimmer which adds the railtie to the host app. The railtie must be loaded during the Rails boot process.

https://github.com/alphagov/govuk_publishing_components/blob/0ffeab9ebe1fd53300c80bea63572ed39b6bf18a/lib/govuk_publishing_components/engine.rb#L5
https://github.com/alphagov/slimmer/blob/868095a99fdf559cdad80b9102d882322214bc60/lib/slimmer/railtie.rb

When reaching this logic, if the gem has been required (ie dev/test/on heroku), then the two modules will be defined and we should move the middleware.

The load order looks correct to me?

My interpretation is that we can't run the Slimmer initialisation any later than this.

alecgibson: Sorry, I misunderstood again. So Slimmer will be loaded automatically by the GovukPublishingComponent gem, but that's only included in the dev or Heroku environment.

That being so, do we need to check `defined?(Slimmer)`? Isn't that implicit in the check for `defined?(GovukPublishingComponents)`?

fofr: Strictly no – I left it in as protection against a version change that later removes Slimmer (or its specific `require` call)

alecgibson: I personally find it a bit confusing - it makes me wonder when Slimmer wouldn't be defined. If it's always going to be defined, I think we shouldn't check. And - especially given that this guide isn't public-facing - it is probably nicer for it to fail loudly than quietly on the off-chance that Slimmer is removed from the gem.

fofr: I've updated and removed the Slimmer part of the logic. :+1:

PR #1128, opened 2017-09-04
Wait for GA Client ID to be loaded before surveys

elenatanasoiu:
For: https://trello.com/c/AokZntDm/221-investigate-why-client-id-isnt-being-loaded-for-all-surveys

## Why 

The assignment of the GA client ID happens by invoking a bound function in `static-analytics.js` (seen below), which is passed to `ga`:

```
// static-analytics.js

ga(function (tracker) {
   this.gaClientId = tracker.get('clientId')

   // Track initial pageview
   this.trackPageview(null, null, trackingOptions);

//   ...
}.bind(this));
```

Normally this happens immediately, and would initialise the GA Client ID which could then be used in `surveys.js` to send the ID along in the survey URL. 

Empirically, we've discovered that this only happens 50% of the time, because `surveys.js` might load before `static-analytics.js`. This means that for 50% of our survey responses we can't track the user in GA using the Client ID, causing our data to be inconsistent.

## What 
Initially `surveys.js` was always loaded after `static-analytics.js` as they were part of the same JS bundle, so the script loading order was always the same. However, due to frequent changes to `surveys.js`, users would have to re-download the entire JS bundle every time we deployed a new change to surveys. To avoid this we've had to extract the `surveys.js` into a separate JS bundle: https://github.com/alphagov/static/pull/1038

To get around the case when `surveys.js` loads **before** `static-analytics.js`, we've modified the surveys code to wait until the `ga` callback has occurred in `static-analytics.js`. Once that happens we trigger a `gaClientSet` event which will be listened for in `surveys.js`. 

We can then initialize surveys and construct the survey URLS using the GA client ID. 

In the alternative case where `surveys.js` loads **after** `static-analytics.js`, then the gaClientID is already set, so we don't setup a listener and just load surveys.

### DIFF
--- a/app/assets/javascripts/analytics/static-analytics.js
+++ b/app/assets/javascripts/analytics/static-analytics.js
@@ -15,6 +15,8 @@
     ga(function (tracker) {
       this.gaClientId = tracker.get('clientId');
 
+      $(window).trigger('gaClientSet');
+
       // Track initial pageview
       this.trackPageview(null, null, trackingOptions);
 
--- a/app/assets/javascripts/surveys.js
+++ b/app/assets/javascripts/surveys.js
@@ -614,8 +614,15 @@
   window.GOVUK.userSurveys = userSurveys
 
   $(document).ready(function () {
-    if (window.GOVUK.userSurveys) {
-      window.GOVUK.userSurveys.init()
+    if (GOVUK.userSurveys) {
+      if (GOVUK.analytics && GOVUK.analytics.gaClientId) {
+        window.GOVUK.userSurveys.init()
+      }
+      else {
+        $(window).on('gaClientSet', function() {
+          window.GOVUK.userSurveys.init()
+        })
+      }
     }
   })
 })(window.jQuery)


nickcolley:
Like the previous pull requests I think this goes against the main principles of the original library, we should track these things as techincal debt.

This seems pretty brittle...

h-lame:
What if we were to fire an event in static analytics when it has received the GA callback and the client id is ready.  Then we can hoo the surveys init code onto receipt of that event?  It'd mean this library is decoupled from GA.  Would that kind of thing address your concenrs at all @nickcolley ?

nickcolley:
Analytics is not really owned by anyone, I don't really have a good mental model of how it works or it's future. I don't want to block anything like this but I feel we need to put together a plan for what happens to analytics in the future since it feels like it's going to be really hard to maintain in the future.

elenatanasoiu:
After discussing with both @nickcolley and @h-lame, we've agreed that the issue is a bit more broad than the current 
scope of this PR. 

Keeping the analytics code as decoupled as possible by using the endpoints defined in `govuk_frontend_toolkit` is a valid concern, and certainly helping to tidy up the way it's being done in static is something our team 
(survey-and-support) could help with. However, we don't really have the authority to decide on 
the types of sweeping changes where we would take the position that either: 

A. in the future we might remove this analytics code and replace it with something else

B. in the future we'll stick with google analytics, so we should focus on extending what already exists in `govuk_frontend_toolkit` and avoid using `ga` in static. 

Depending on the position taken, our definition of "tidying up" will differ and would affect the way we use GA calls in `static`. 

For the moment, we've agreed to merge the PR the way it is, until we get a steer on how we've decided to approach usage of GA in `static` in the future.

PR #1129, opened 2017-09-05
Adding inverse flag to title component

vanitabarrett:
Passing inverse: true to title component means the title can work on a dark background.

screen shot 2017-09-06 at 13 18 26

**Note: needs to be merged before updating HTML publications in government-frontend, but only after update to govuk_publishing_components.**

### DIFF
--- a/app/views/govuk_component/docs/title.yml
+++ b/app/views/govuk_component/docs/title.yml
@@ -27,3 +27,11 @@ examples:
       context: Publication
       title: My page title which is often really long and verbose and has lots of extra words it doesn't need
       average_title_length: long
+  title_on_dark_background:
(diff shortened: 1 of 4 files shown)


app/views/govuk_component/docs/title.yml
```diff
@@ -27,3 +27,10 @@ examples:
       context: Publication
       title: My page title which is often really long and verbose and has lots of extra words it doesn't need
       average_title_length: long
+  title_on_dark_background:
```
fofr: We could call this `html_publication_title`, or use the description to indicate where this is used.

nickcolley: Description seems like a good shout 👍

PR #1134, opened 2017-09-07
Add some more GDS user research panel surveys

DennisTsiang:
For: https://trello.com/c/Asj5HLO6

Adds links to the four surveys on the pages requested, each with a frequency of 1 in 6, running from Aug 25 2017 to Oct 1 2017.

### DIFF
--- a/app/assets/javascripts/surveys.js
+++ b/app/assets/javascripts/surveys.js
@@ -175,6 +175,104 @@
           title: 'Tell us about your visit to GOV.UK',
           surveyCta: 'Answer 2 quick questions to help us make GOV.UK better',
         }
+      },
+      {
+        identifier: 'scrutinise_government',
+        surveyType: 'url',
+        frequency: 6,
+        startTime: new Date('August 25, 2017').getTime(),
+        endTime: new Date('October 1, 2017 23:59:50').getTime(),
+        url: 'https://signup.take-part-in-research.service.gov.uk/?utm_campaign=Scrutinise_government&utm_source=Hold_gov_to_account&utm_medium=gov.uk&t=GDS&id=15',
+        templateArgs: {
+          title: 'Help improve GOV.UK',
+          surveyCta: 'Answer some questions about yourself to join the research community.',
+          surveyCtaPostscript: 'This link opens in a new tab.'
+        },
+        activeWhen: {
+          path: [
+            '^/government/announcements/?$',
+            '^/government/publications/?$',
+            '^/government/statistics/?$',
+            '^/government/organisations/hm-revenue-customs/?$',
+            '^/government/organisations/companies-house/?$',
+            '^/government/organisations/driver-and-vehicle-licensing-agency/?$',
+            '^/government/organisations/uk-visas-and-immigration/?$',
+            '^/government/organisations/land-registry/?$'
+          ]
+        }
+      },
+      {
+        identifier: 'content_history',
+        surveyType: 'url',
+        frequency: 6,
+        startTime: new Date('August 25, 2017').getTime(),
+        endTime: new Date('October 1, 2017 23:59:50').getTime(),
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/surveys.js
```diff
@@ -175,6 +175,111 @@
...
+          surveyCta: 'Answer some questions about yourself to join the research community.',
+          surveyCtaPostscript: 'This link opens in a new tab.'
+        },
+        activeWhen: {
+          path: [
+            '^/government/policies(?:/|$)',
+            '^/government/announcements(?:/|$)',
+            '^/government/publications(?:/|$)',
+            '^/government/statistics(?:/|$)',
+            '^/government/publications?publication_filter_options=consultations(?:/|$)',
```
h-lame: We can drop this URL.  Primarily because the current implementation of the surveys code only looks at the `/path` not the `?query=string` of the current page, so this would never match anything, but also because (luckily for us) we already have `/government/publications` above which would cover it.

app/assets/javascripts/surveys.js
```diff
@@ -175,6 +175,111 @@
...
+        endTime: new Date('October 1, 2017 23:59:50').getTime(),
+        url: 'https://signup.take-part-in-research.service.gov.uk/?utm_campaign=Becoming_a_childminder&utm_source=Group_and_Transform_Content&utm_medium=gov.uk&t=GDS&id=18',
+        templateArgs: {
+          title: 'Help improve GOV.UK',
+          surveyCta: 'Answer some questions about yourself to join the research community.',
+          surveyCtaPostscript: 'This link opens in a new tab.'
+        },
+        activeWhen: {
+          path: [
+            '^/register-childminder-childcare-provider(?:/|$)',
```
h-lame: I think we can drop the other `/register-childminder-chidcare-provider/<whatever>` paths as this one will match any subpages, and we've been asked to run it on all the subpages, so we can get away with just the root.  It does no harm to have the other explicit paths in here, but we can save a few bytes this way.

Need to keep the northern-ireland one though as it's a separate page.

app/assets/javascripts/surveys.js
```diff
@@ -175,6 +175,111 @@
...
+            '^/government/policies(?:/|$)',
+            '^/government/announcements(?:/|$)',
+            '^/government/publications(?:/|$)',
+            '^/government/statistics(?:/|$)',
+            '^/government/publications?publication_filter_options=consultations(?:/|$)',
+            '^/government/organisations/hm-revenue-customs(?:/|$)',
+            '^/government/organisations/companies-house(?:/|$)',
+            '^/government/organisations/driver-and-vehicle-licensing-agency(?:/|$)',
+            '^/government/organisations/uk-visas-and-immigration(?:/|$)',
+            '^/government/organisations/land-registry(?:/|$)'
```
h-lame: As many of these pages are index pages that have lots of subpages (for example /government/publications/ is the root for almost every document published via whitehall, we need to check if the intention is that we only show the survey on the index, or if we show it on all child pages too.  I suspect in this case we only mean the index in which case we have to change the regexs here.  We've got `(?:/|$)` (this matches a `/` or the end of the string) as a suffix which will match `/government/publications`, `/government/publications/` and `/government/publications/anything` (but not `/government/publications-about-something-else`).  To avoid matching the third form we can change the suffix to: `/?$` (this matches the end of the string or a `/` followed by the end of the string).  I've asked Sean to confirm.

h-lame: We've decided that we should treat all these urls as exact matches rather than prefix matches, so the survey only appears on these exact pages and not any of their children pages.  The urls for the other surveys are all ok as they're already specific pages, or (as already mentioned for `becoming_a_childminder` are prefix paths that we want all the child pages of anyway).

PR #1136, opened 2017-09-08
Fix issue preventing deployment

whoojemaflip:
Due to a ruby version mixup, `Numeric#positive?` isn't a method available in our production stack. 

Temporary fix to get deployments moving again. I hope to be able to move the logic out of this Gemfile in a later PR.


fofr:
We also didn't catch this in the Heroku production deployment because Heroku doesn't respect the `.ruby-version` file.

PR #1137, opened 2017-09-08
Remove 'require' logic from Gemfile

whoojemaflip:
Move the decision point for the `require` out of the Gemfile and into application.rb

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,7 +10,7 @@ gem 'sass-rails', "5.0.6"
 gem 'airbrake', github: 'alphagov/airbrake', branch: 'silence-dep-warnings-for-rails-5'
 gem 'nokogiri', "~> 1.7"
 gem 'redis', "~> 3.3.3"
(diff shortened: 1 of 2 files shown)


nickcolley:
Nice! We should probably add this to https://github.com/alphagov/govuk_publishing_components#install-gem and also update government-frontend https://github.com/alphagov/government-frontend/blob/master/Gemfile#L11

nickcolley:
Could you add a comment in the gemfile to explain what we're doing here, so people are not confused at why it's false but still displaying in development/heroku

whoojemaflip:
@nickcolley Isn't that what `requires: false` in a Gemfile signifies? That a gem will be required by hand rather than by magic / bundler.

nickcolley:
@whoojemaflip as someone who doesn't frequent gemfiles that's not something I'd know off the top of my head

fofr:
Seems to be working as expected:
http://govuk-static-pr-1137.herokuapp.com/component-guide/

nickcolley:
I'm going to update the gem and government-frontend with this in mind.

PR #1138, opened 2017-09-08
Add to survey document about testing

DennisTsiang:
This commits add some extra information on how to test that the
surveys are working correctly. Hopefully this will be useful for future
devs.


nickcolley:
I wonder if this is easier now we have heroku review apps for each PR?

PR #1139, opened 2017-09-11
Add more HMRC surveys

DennisTsiang:
For: https://trello.com/c/xcvqIxHr, https://trello.com/c/zXdSm7Ho

Adds 3 new surveys on the pages requested, running from Sept 13 2017
to Oct 14 2017.

### DIFF
--- a/app/assets/javascripts/surveys.js
+++ b/app/assets/javascripts/surveys.js
@@ -137,6 +137,60 @@
         }
       },
       {
+        identifier: 'hmrc_child_benefits',
+        surveyType: 'url',
+        frequency: 6,
+        startTime: new Date('September 13, 2017').getTime(),
+        endTime: new Date('October 14, 2017 23:59:50').getTime(),
+        url: 'https://signup.take-part-in-research.service.gov.uk/?utm_campaign=ChildBenefitGOV&utm_source=Other&utm_medium=other&t=HMRC&id=26',
+        templateArgs: {
+          title: 'Help improve GOV.UK',
+          surveyCta: 'Answer some questions about yourself to join the research community.',
+          surveyCtaPostscript: 'This link opens in a new tab.'
+        },
+        activeWhen: {
+          path: [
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/surveys.js
```diff
@@ -137,6 +139,43 @@
         }
       },
       {
+        identifier: 'hmrc_sept2017',
+        surveyType: 'url',
+        frequency: 6,
+        startTime: new Date('September 13, 2017').getTime(),
+        endTime: new Date('October 14, 2017 23:59:50').getTime(),
+        url: 'https://signup.take-part-in-research.service.gov.uk/?utm_campaign=' + hmrcSurveyUtmCampaignValueMap(),
```
h-lame: I think it'll be clearer to have 2 separate surveys for this one as their urls aren't really that similar.  The reason we have 1 survey for the existing hmrc one and use that `hmrcSurveyUtmCampaignValueMap` is that a) there's a lot of them, and b) their urls only differ by the value of the `utm_campaign` value, but these two a) are only 2, b) have wider differences in their urls.  Adding them to the `hmrcSurveyUtmCampaignValueMap` also means we end up changing that function - before this change the return value is a value for the `utm_campaign` param, after this change it might also be an entire url suffix including extra params, which could cause problems for call-sites that assume the return value is still just a param value.

app/assets/javascripts/surveys.js
```diff
@@ -369,7 +408,7 @@
       var allSurveys = [defaultSurvey].concat(activeSurveys)
       var displayableSurveys = userSurveys.getDisplayableSurveys(allSurveys)
 
-      if (displayableSurveys.length < 1) {
+      if (displayableSurveys.length < 2) {
```
h-lame: Ha! yes.  Good spot!

DennisTsiang:
Changed them to separate survey blocks now

PR #1140, opened 2017-09-11
Update Gemfile to reference ruby-version file

nickcolley:
We have had issues with static not running correctly based on the ruby version being different
on Heroku and production, this change ensures the version is the same

Thanks to @tijmenb for this 👍

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,6 +10,7 @@ ENV APP_HOME /app
 RUN mkdir $APP_HOME
 
 WORKDIR $APP_HOME
+ADD .ruby-version $APP_HOME/
 ADD Gemfile* $APP_HOME/
 RUN bundle install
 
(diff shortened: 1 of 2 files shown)


nickcolley:
@tijmenb I tried a bunch of variations and only the current one gets us a green build, I'd prefer something more verbose and explicit, thoughts?

tijmenb:
So this happens only when building the docker container, not during normal CI. I'm guessing it's caused by this line:

https://github.com/alphagov/static/blob/master/Dockerfile#L12-L14

I'm no docktor, but I think that means that the Gemfile is copied into something, which wouldn't have access to the rest of the files (like `.ruby-version`). Maybe try adding:

```
ADD .ruby-versionb* $APP_HOME/
```

fofr:
I wonder if https://github.com/alphagov/static/blob/master/Dockerfile#L1 risks getting out of date.

PR #1142, opened 2017-09-13
Implement customisable spacing for title component

andysellick:
- remove bottom margin from component
- add margin helper class for original component bottom margin
- default component state will use that class unless an alternative is passed i.e. 0, which means there is no need to update any instance of the component as it will still be as it was before this change

screen shot 2017-09-13 at 16 59 56

https://trello.com/c/KEkNsxG3/142-3-implement-customisable-spacing-for-components

This issue is complicated by our need for responsive margins. The ideal would be a mechanism where we pass a number to the component to interpret as margin-bottom e.g. 0 = no margin, 1 = gutter-half, 2 = gutter, etc. but an [https://docs.google.com/spreadsheets/d/1mL-zAgi_Vfb7BO3KPF8b9xAPq_pIXIZHHSjvlB1Tulg/edit#gid=0](audit of the components) reveals there is no simple scale for doing this when responsive is included (even though this change is for one component it would be good to consider the others).

We could try separating the responsive classes e.g. pass margin-bottom-mobile = 1, margin-bottom-tablet = 1, margin-bottom-desktop = 2 but that seemed overly horrible.

### DIFF
--- a/app/views/govuk_component/title.raw.html.erb
+++ b/app/views/govuk_component/title.raw.html.erb
@@ -2,8 +2,12 @@
   average_title_length ||= false
   context ||= false
   inverse ||= false
+  margin_bottom ||= 1
+  margin_bottom_class = "pub-c-title--bottom-margin" if margin_bottom == 1
 %>
-<div class="pub-c-title <% if inverse %>pub-c-title--inverse<% end %>">
+<div class="pub-c-title
+  <% if inverse %>pub-c-title--inverse<% end %>
+  <%= margin_bottom_class %>">
   <% if context %>
     <p class="pub-c-title__context"><%= context %></p>
   <% end %>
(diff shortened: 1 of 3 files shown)


app/views/govuk_component/title.raw.html.erb
```diff
@@ -2,8 +2,12 @@
   average_title_length ||= false
   context ||= false
   inverse ||= false
+  margin_bottom ||= 1
+  margin_bottom_class = "pub-c-title--bottom-margin" if margin_bottom == 1
```
fofr: Maybe:
```
margin_bottom ||= true
margin_bottom_class = "pub-c-title--bottom-margin" if margin_bottom
```

andysellick: Don't think that works - if you pass 0 to the component that counts as true.

nickcolley:
```
margin_bottom_class = "pub-c-title--bottom-margin" if margin_bottom !== 0
```

?

andysellick: That doesn't work either... margin_bottom is not defined.

PR #1144, opened 2017-09-15
[Do not merge] Make sure we stub locale components in metadata_test

h-lame:
Otherwise it may not pass (sometimes the locales have already been stubbed
by another test which is why it passes sometimes, but not all).

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -55,9 +55,22 @@ module Static
     # Slimmer is used by the govuk_publishing_components gem, however it inserts itself
     # automatically as middleware in the host Rails application
     # Disable Slimmer middleware for Static and enable for use in gem only
+    # Replace Slimmer I18n Backend with a null backend that contains no
+    # translations. We can't remove it from the i18n backend chain because
+    # it's added on every request if it's missing.
     if defined?(GovukPublishingComponents)
       config.middleware.delete Slimmer::App
       GovukPublishingComponents::Engine.config.middleware.use Slimmer::App
+
+      class AvoidSlimmerNullBackend
+        include I18n::Backend::Base
+
+        def available_locales; []; end
+
+        def lookup(locale, key, scope = [], options = {}); nil; end
+      end
+
+      Slimmer::I18nBackend = AvoidSlimmerNullBackend
     end
   end
 end


fofr:
In the context of this test case, it shouldn't be making requests to Slimmer. Static ordinarily doesn't use slimmer, so I'm somewhat concerned that this happening at all.

h-lame:
@fofr - now I stop to think about it rather than just fixating on getting a robust green build, I agree.  I'll see if I can dig in deeper to work out what's going on, but I may need your assistance in unpicking it.

h-lame:
The problem is that once we make a request to `/component-guide` in one of our tests the `GovukPublishingComponents::ComponentGuideController` is loaded, which in turns means the `GovukPublishingComponents::ApplicationController` is loaded, and this includes `Slimmer::GovukComponents`.  This sets up the govuk_publishing_components controllers to use slimmer, which is fine, but it also sets up the global `I18n` to use a slimmer backend.  This means that when we do locale stuff we'll try to load the locale file form slimmer (e.g. over the wire from static).  This is a global setting, so while the specific test case shouldn't be using slimmer, it does because it uses `I18n`.

I am not entirely sure what the best way to fix this might be.  I think it's another unepxected side-effect of putting slimmer into static so that it has to talk to itself over the network.  We might need to build a config option for slimmer that allows it to run in "local" mode to avoid it entirely.  It should be able to collect locale info and template info directly off disk instead of having to go out from the app and back into itself.

h-lame:
It's worth noting that this problem has surfaced in the tests, but I'm pretty sure it can happen in any environment where govuk_publishing_components is loaded and made available.  It's less likely to be a problem in practice because static doesn't actually interpret the templates itself, they are rendered raw to the applications that need them, so static won't make any calls out to itself for the locales.

fofr:
For i18n, eg this line:
https://github.com/alphagov/slimmer/blob/83287c1750342ee88a8a9b6cd177dc9e5d88c900/lib/slimmer/govuk_components.rb#L23

Can we do some reset like we did here: https://github.com/alphagov/static/blob/master/config/application.rb#L58-L61

h-lame:
@fofr - oh, that's a good idea.  I'll redo the PR and see if it helps.

h-lame:
On reflection, I don't think we can remove the backend in `application.rb`.  It's only added after the first request to a `GovukPublishingComponents::ApplicationController` is made; so it won't have been added by the time `aaplication.rb` is evaluated.  Even if it was there and we removed it, the `before_action` that adds it would spot that it's not there and add it back again.

h-lame:
@fofr - we can't remove the backend from the chain (or remove the code that adds it), but we can replace the implementation with one that does nothing.  I've redone the PR to try that and it seems to work.

config/application.rb
```diff
@@ -55,9 +55,22 @@ class Application < Rails::Application
...
+
+      class AvoidSlimmerNullBackend
+        include I18n::Backend::Base
+
+        def available_locales; []; end
+
+        def lookup(locale, key, scope = [], options = {}); nil; end
+      end
+
+      Slimmer::I18nBackend = AvoidSlimmerNullBackend
```
fofr: I wonder if this whole method could be stubbed out:
https://github.com/alphagov/slimmer/blob/83287c1750342ee88a8a9b6cd177dc9e5d88c900/lib/slimmer/govuk_components.rb#L19

We shouldn’t need the component resolver either as templates will be available locally.

fofr: This might not work because our partials in static don't being with an underscore 😕

h-lame:
Closing this hot mess in favour of the much more reasoned #1149 which uses a version of slimmer that knows how to use local things.

PR #1145, opened 2017-09-15
Fix output of finding installed location

maxgds:
This is a fix to the startup file which is currently failing.
The output of "bundle show govuk_frontend_toolkit" currently
contains warnings about us not running the latest version of
bundler which means the script fails. This fix simply
truncates the output at the first line.

(Note this is not a merge into master.)


nickcolley:
Nice 👍

PR #1146, opened 2017-09-18
Fix aria expanded and height of option select

### LINKED ISSUE #860: Option selects aren’t set up correctly when open on page load
If an option select is set to be open (expanded) when the page loads then:
- the button which controls it’s opening/closing has its `aria-expanded` attribute set to `false`, even though it is open
- the code which [sets the height of the options container](https://github.com/alphagov/static/pull/556) doesn’t get executed, so users may not realise that the options are scrollable

***

Both issues can be seen here (on [the Competition and Markets Authority cases page](https://www.gov.uk/cma-cases)):

opt-select

andysellick:
- set aria-expanded to true by default when initialising rather than performing a check as it will always be open on load, unless it's set to be closed in which case the code will change it back to false anyway as it then closes it
- fix for height issue by calling setheight on initialisation, code is structured already so that it won't get called again unless it needs to be
- fixes #860
- includes some CSS optimisation/improvements (tested in various including IE8, 9, 10, 11, Firefox current, 48, Edge, Chrome current, 54, Opera 40, Safari 5.1, iphone 6s, ipad mini)

Before:

screen shot 2017-09-18 at 10 49 26

After:

screen shot 2017-09-19 at 12 50 59

https://govuk-static-pr-1146.herokuapp.com/component-guide/option_select

### DIFF
--- a/spec/javascripts/govuk-component/option-select-spec.js
+++ b/spec/javascripts/govuk-component/option-select-spec.js
@@ -66,24 +66,54 @@ describe('GOVUK.OptionSelect', function() {
 
 
   it('instantiates a closed option-select if data-closed-on-load is true', function(){
-    $closedOnLoadFixture = $('<div class="govuk-option-select" data-closed-on-load=true></div>');
+    closedOnLoadFixture = '<div class="govuk-option-select" data-closed-on-load=true>' +
+                            '<div class="container-head js-container-head"></div>' +
+                          '</div>';
+    $closedOnLoadFixture = $(closedOnLoadFixture);
+
     $('body').append($closedOnLoadFixture);
     optionSelect = new GOVUK.OptionSelect({$el:$closedOnLoadFixture});
     expect(optionSelect.isClosed()).toBe(true);
+    expect($closedOnLoadFixture.find('button').attr('aria-expanded')).toBe('false');
   });
 
   it('instantiates an open option-select if data-closed-on-load is false', function(){
-    $openOnLoadFixture = $('<div class="govuk-option-select" data-closed-on-load="false"></div>');
+    openOnLoadFixture = '<div class="govuk-option-select" data-closed-on-load=false>' +
+                            '<div class="container-head js-container-head"></div>' +
+                          '</div>';
+    $openOnLoadFixture = $(openOnLoadFixture);
+
     $('body').append($openOnLoadFixture);
     optionSelect = new GOVUK.OptionSelect({$el:$openOnLoadFixture});
     expect(optionSelect.isClosed()).toBe(false);
+    expect($openOnLoadFixture.find('button').attr('aria-expanded')).toBe('true');
   });
 
   it('instantiates an open option-select if data-closed-on-load is not present', function(){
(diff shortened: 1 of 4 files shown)


nickcolley:
Could we create a failing test for this?

andysellick:
On it.

spec/javascripts/govuk-component/option-select-spec.js
```diff
@@ -86,6 +86,10 @@ describe('GOVUK.OptionSelect', function() {
     expect(optionSelect.isClosed()).toBe(false);
   });
 
+  it ('has called setupHeight as part of initialisation', function(){
```
nickcolley: Can we make the test name to what we're testing, which is that it's expanded by default, and update the test to reflect this?

fofr:
@andysellick It's helpful to include links for review like:
https://govuk-static-pr-1146.herokuapp.com/component-guide/option_select

spec/javascripts/govuk-component/option-select-spec.js
```diff
@@ -86,6 +86,17 @@ describe('GOVUK.OptionSelect', function() {
     expect(optionSelect.isClosed()).toBe(false);
   });
 
+  it ('sets the height of the options container as part of initialisation', function(){
```
nickcolley: These tests could communicate the bug we're fixing, if we named this test against how we're checking the component is expanded correctly we'd notice that we're testing the height is set but we're not testing that the aria-expanded attribute is set properly.

fofr: We need a separate second test to check that aria-expanded has the correct value(s)

spec/javascripts/govuk-component/option-select-spec.js
```diff
@@ -124,12 +135,6 @@ describe('GOVUK.OptionSelect', function() {
       expect($optionSelectHTML.hasClass('js-closed')).toBe(false);
     });
 
-    it ('calls setupHeight()', function(){
```
fofr: Why delete this test?

fofr:
Visual regression picked up on the changes:
http://govuk-static-pr-1146.surge.sh/gallery.html

spec/javascripts/govuk-component/option-select-spec.js
andysellick: This test assumes setupHeight gets called when the option select is opened, but that's now not always the case. setupHeight is called on init if the option select is in the expanded state, but not if it's collapsed. I'm writing other tests to fill this gap.

spec/javascripts/govuk-component/option-select-spec.js
andysellick: The first two tests in the file now cover this for both open and closed initial states.

spec/javascripts/govuk-component/option-select-spec.js
andysellick: Have now put this test back in a modified form.

nickcolley:
@andysellick can you put in the PR which browsers/devices you've tested the CSS changes against?

PR #1151, opened 2017-09-21
Enable automated accessibility testing

fofr:
* Use aXe tests provided by govuk_publishing_components gem
* Add poltergeist to enable integration tests with JS
* Update gem config to explicitly include styles missing from Slimmer

Fix found accessibility failures:

* Option select: Only add aria-controls if element is present
* Option select: Declare list of checkboxes as checkbox group with label
* GovSpeak: Text in bar charts has a contrast too low against light blue for WCAG
* GovSpeak: Increase font on highlight answers for WCAG AA contrast
* Bump govuk_publishing_components to 1.10 to avoid duplicate ID errors

Caveats:

* Component JS isn't running on the integration tests because I can't easily include the jQuery dependency they need. Will tackle this in separate PR.
* The visual regression tests produce screenshots that aren't always long enough (https://govuk-static-pr-1151.surge.sh/gallery.html)

# Highlight answer

Updated to match https://www.gov.uk/bank-holidays and pass WCAG AA contrast ratio

https://govuk-static-pr-1151.herokuapp.com/component-guide/govspeak/highlight_answer

## Before

screen shot 2017-09-21 at 17 10 06

## After

screen shot 2017-09-21 at 17 09 58

### Before
screen shot 2017-09-21 at 17 12 21

### After
screen shot 2017-09-21 at 17 04 11

## Visual diff
screen shot 2017-09-21 at 17 48 04

Part of:
https://trello.com/c/Yq9FFia1/91-3-move-component-a11y-testing-from-application-into-gem-so-applications-get-the-testing-for-free


fofr:
Test failures are Poltergeist timeouts

fofr:
Somewhat confused about the test failure, which is on the sticky element JS which hasn't been edited.

test/integration_test_helper.rb
```diff
@@ -1,6 +1,18 @@
 require_relative 'test_helper'
 require 'rack/test'
 require 'capybara/rails'
+require 'capybara/poltergeist'
+require 'phantomjs'
+
+# TODO: Remove when PhantomJS updated on Puppet
+# Use modern PhantomJS version via gem
+# Based on: https://github.com/colszowka/phantomjs-gem/blob/master/lib/phantomjs/poltergeist.rb
+Phantomjs.path
```
nickcolley: Is this intentional?

fofr: Yes, see link above.

app/assets/javascripts/govuk-component/option-select.js
```diff
@@ -70,6 +71,15 @@
 
   };
 
+  OptionSelect.prototype.setCheckboxAriaControlsAttributes = function setCheckboxAriaControlsAttributes(){
+    var controls = this.$optionSelect.data('input-aria-controls');
+    if (typeof controls === "string" && $('#' + controls).length > 0) {
+      this.$optionSelect.find('input[type="checkbox"]').each(function() {
```
nickcolley: Do we know how long this takes to execute?

app/views/govuk_component/docs/govspeak.yml
```diff
@@ -608,11 +608,24 @@ examples:
           </li>
         </ol>
   highlight_answer:
+    description: |
+      Used on:
+
+      * [https://gov.uk/vat-rates](https://gov.uk/vat-rates)
```
nickcolley: 👍  nice

test/integration/component_guide_test.rb
```diff
@@ -16,23 +17,30 @@ def skip_slimmer
 
 GovukPublishingComponents::ApplicationController.include(GovukPublishingComponentsSkipSlimmer)
 
+# Tests aren't going through Slimmer so we need to explicitly include styles
+GovukPublishingComponents.configure do |c|
+  c.application_print_stylesheet = "core-layout-print"
+  c.application_stylesheet = "core-layout"
```
nickcolley: do we need the grid? could we use header-and-footer-only?

fofr: I'm using the same thing that the guide uses when you view it in a browser.

app/views/govuk_component/option_select.raw.html.erb
```diff
@@ -1,25 +1,23 @@
...
+>
   <div class="container-head js-container-head">
-    <div class='option-select-label'><%= title %></div>
+    <div class="option-select-label" id="<%= title_id %>">
+      <%= title %>
+    </div>
   </div>
-  <div class='options-container' id='<%= options_container_id %>'>
-    <div class='js-auto-height-inner'>
+  <div role="group" aria-labelledby="<%= title_id %>" class="options-container" id="<%= options_container_id %>">
```
nickcolley: Do we need to check this change in different A.T?

app/assets/javascripts/govuk-component/option-select.js
nickcolley: Before
57.10ms
46.03ms
51.65ms

After
56.92ms
45.76ms
55.98ms

At CPU 6x slowdown

app/views/govuk_component/option_select.raw.html.erb
fofr: Yes. I've only been able to test with VoiceOver on Mac and iOS so far.

fofr:
Changes in 715a9cdc2ae33bf15ce0726fe69e7be12bfdfcd7 are picked up by JAWS 17 and 18. They make the reading of each checkbox more verbose by reading the group name before the checkbox name each time. Unlike on VoiceOver which does it the first time but not the following times. This may slow down users when navigating through a list of items in a group with a long name. 

NVDA behaves the same before and after, ie it doesn't support the aria group role.

JAWS 17 and 18:
> Market sector. Aerospace checkbox. Not checked
> Market sector. Agriculture checkbox. Not checked

NVDA:
>  Aerospace checkbox. Not checked

VoiceOver:
> Market sector. Aerospace checkbox. Not checked
> Agriculture checkbox. Not checked

fofr:
Confirmed that 0a70e29fb9337c4cfd3c4973e30e3eb9e9e65b51 works as expected on VoiceOver for Mac and iOS (checkbox changes read out the loading/results element)

nickcolley:
Talked to Paul, this change is seems good and does not introduce barriers, so we'll move this forwards and get some opinions on verbosity of re-reading of the label with the accessibility mailing list.

PR #1161, opened 2017-10-03
Remove old HMRC webchat

h-lame:
For: https://trello.com/c/hWTtb0OZ/253-remove-old-instance-of-webchat

This implementation of webchat uses the old HMRC webchat so we want to get
rid of it.  There is a newer implementation included in
government-frontend that we could replace this one with, but we realised
that most of the pages the older webchat version is active on link to an
HMRC contact page.  All the HMRC contact pages have the new webchat
implementation on it so there's no need to have the old webchat.  For those
that don't link to an HMRC contact page we decided it wasn't a problem to
not have webchat.


nickcolley:
We have an ongoing dialogue with HMRC on this, lets hold off until that's finished.

nickcolley:
I checked your trello card, it's not clear if this work is included with that conversation...

h-lame:
@nickcolley - on the 21st Sept. we told HMRC we'd be removing this version of webchat in the next couple of weeks so they are aware this is happening and haven't told us not to.

huminmiah:
@nickcolley @h-lame Yep. I've been in touch with HMRC and made the decision to remove the old webchat. We're confident the legacy webchat is no longer in use and they are aware of the content pages it was used on.

PR #1168, opened 2017-10-13
Add lead paragraph component

maxgds:
- Imports the lead paragraph component from government-frontend
  so that it can be shared across applications

Goes with https://github.com/alphagov/government-frontend/pull/504

### DIFF
--- a/app/assets/stylesheets/govuk-component/_lead-paragraph.scss
+++ b/app/assets/stylesheets/govuk-component/_lead-paragraph.scss
@@ -1,0 +1,6 @@
+.pub-c-lead-paragraph {
+  @include core-24;
+  @include responsive-bottom-margin;
+  // Ensure the text has a line-length of around 60 characters
+  max-width: 30em;
+}
--- a/app/views/govuk_component/lead_paragraph.raw.html.erb
+++ b/app/views/govuk_component/lead_paragraph.raw.html.erb
@@ -1,0 +1,10 @@
+<%
+  text ||= ""
+  escaped_text = html_escape_once(text.strip)
+  text = escaped_text.sub(/\s([\w\.\?\!\:]+)$/, '&nbsp;\1').html_safe
+%>
+<% if text.present? %>
+  <p class="pub-c-lead-paragraph">
+    <%= text %>
+  </p>
+<% end %>
--- a/test/govuk_component/lead_paragraph_test.rb
+++ b/test/govuk_component/lead_paragraph_test.rb
@@ -1,0 +1,39 @@
+require 'govuk_component_test_helper'
+
+class LeadParagraphTest < ComponentTestCase
+  NBSP = HTMLEntities.new.decode('&nbsp;')
+
+  def component_name
+    "lead_paragraph"
+  end
+
+  def assert_lead_paragraph_matches(text, expected_text)
+    render_component(text: text)
+    assert_select ".pub-c-lead-paragraph", text: expected_text
+  end
+
+  test "renders nothing without a description" do
+    assert_empty render_component({})
+  end
+
+  test "renders a lead paragraph" do
+    render_component(text: 'UK Visas and Immigration is making changes to the Immigration Rules affecting various categories.')
+    assert_select ".pub-c-lead-paragraph", text: "UK Visas and Immigration is making changes to the Immigration Rules affecting various#{NBSP}categories."
+  end
+
+  [
+    { text: 'this and that', expected: "this and#{NBSP}that", description: 'non breaking space between last two words' },
+    { text: 'this and that.', expected: "this and#{NBSP}that.", description: 'non breaking space between last two words with trailing space' },
+    { text: 'this and that.', expected: "this and#{NBSP}that.", description: 'non breaking space between last two words at end of sentence' },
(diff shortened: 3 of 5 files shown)


app/assets/stylesheets/govuk-component/_lead-paragraph.scss
```diff
@@ -0,0 +1,6 @@
+.app-c-lead-paragraph {
```
nickcolley: The prefix needs to be 'pub' so it doesnt conflict with the app level component

maxgds: Fixed

nickcolley:
Link to test: https://govuk-static-pr-1168.herokuapp.com/component-guide/lead_paragraph

app/views/govuk_component/lead_paragraph.raw.html.erb
```diff
@@ -0,0 +1,13 @@
+<%
+  text ||= ""
+  def nbsp_between_last_two_words(text)
+    return text if text.nil?
```
fofr: This first line will never happen so can be omitted – we set it to an empty string in the line above

app/views/govuk_component/lead_paragraph.raw.html.erb
```diff
@@ -0,0 +1,13 @@
+<%
+  text ||= ""
+  def nbsp_between_last_two_words(text)
```
fofr: This helper has some tests in government-frontend: https://github.com/alphagov/government-frontend/blob/master/test/helpers/typography_helper_test.rb – we should probably have something similar here in static.

maxgds: Added.

app/views/govuk_component/lead_paragraph.raw.html.erb
maxgds: Fixed this. (Hopefully also squashed the linting problem.)

app/views/govuk_component/lead_paragraph.raw.html.erb
```diff
@@ -0,0 +1,12 @@
...
+    escaped_text = html_escape_once(text.strip)
+    escaped_text.sub(/\s([\w\.\?\!\:]+)$/, '&nbsp;\1').html_safe
+  end
+%>
```
fofr: I'm not sure we need the method definition in here as the logic isn't going to be reused. Could rewrite as:

```erb
<%
  text ||= ""
  escaped_text = html_escape_once(text.strip)
  text = escaped_text.sub(/\s([\w\.\?\!\:]+)$/, '&nbsp;\1').html_safe
%>
```

Then further down use:
`<%= text %>`

The tests can then test different component input, checking that the outputted HTML is as expected.

I'm not sure what scope the `nbsp_between_last_two_words` method has in a view, and I'm surprised the tests worked without needing to call render on a component.

test/govuk_component/lead_paragraph_test.rb
```diff
@@ -0,0 +1,46 @@
...
+      assert_lead_paragraph_matches(test_params[:text], test_params[:expected])
+    end
+  end
+
+  test "lead_paragraph:nbsp_between_last_two_words leaves alone single words" do
+    assert_lead_paragraph_matches("this", "this")
+  end
+
+  test "lead_paragraph:nbsp_between_last_two_words isn't unsafe" do
+    AMPERSAND = HTMLEntities.new.decode('&amp;')
```
fofr: This line isn't used. Surprised the linter hasn't complained about that.

test/govuk_component/lead_paragraph_test.rb
```diff
@@ -0,0 +1,46 @@
...
+    assert_lead_paragraph_matches("&lt;b&gt;this&lt;b&gt; &amp; that&nbsp;thing", "<b>this<b> & that#{NBSP}thing")
+    assert_lead_paragraph_matches("&lt;b&gt;this&lt;b&gt; &amp;&nbsp;that", "<b>this<b> &#{NBSP}that")
```
fofr: These tests and `assert_lead_paragraph_matches("this", "this")` could probably be bundled up into the list above, as they're all doing the same thing.

maxgds: Fixed.

test/govuk_component/lead_paragraph_test.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  [
+    { text: 'this and that', expected: "this and#{NBSP}that", description: 'non breaking space between last two words' },
+    { text: 'this and that.', expected: "this and#{NBSP}that.", description: 'non breaking space between last two words with trailing space' },
+    { text: 'this and that.', expected: "this and#{NBSP}that.", description: 'non breaking space between last two words at end of sentence' },
+    { text: 'this and that. ', expected: "this and#{NBSP}that.", description: 'non breaking space between last two words with trailing space after sentence' },
+    { text: "multiline\nthis and that", expected: "multiline\nthis and#{NBSP}that", description: 'non breaking space between last two words with new lines' },
+    { text: "this and that\n\n", expected: "this and#{NBSP}that", description: 'non breaking space between last two words with multiline input' },
+    { text: "this", expected: "this", description: 'non breaking space single word input gives single word output' },
```
fofr: Don't need the `non breaking space` at the start of this description.

PR #1173, opened 2017-10-26
Add tasklist component

andysellick:
A tasklist is meant to help users achieve a goal on GOV.UK, such as learning to drive. Links to pages of information are presented in a series of steps, each of which containing one or more expandable sections.

- based on the accordion code
- the JS should be quite similar to the accordion but tasklist does not have the 'last section open remembered in the URL' functionality enabled by default
- the JS spec file should be almost identical to the one from accordion, but with an addition for the data-remember attribute for 'last section open remembered in URL' functionality

Large tasklist:

screen shot 2017-10-26 at 15 58 09

Small tasklist (large should also look like this on mobile):

screen shot 2017-10-26 at 15 58 23

Component guide link: https://govuk-static-pr-1173.herokuapp.com/component-guide/task_list

Related trello card: https://trello.com/c/o0BCuFub/190-rewrite-the-accordion-to-meet-our-teams-needs

**Accessibility and compatibility**

This work included a session with the GDS accessibility team, who made a number of recommendations which have since been implemented. It will go into public beta first, during which we'll be testing it with users with accessibility needs.

Tested with IE8 -> 11, Edge latest, Chrome latest and 54, FF latest and 48, Opera latest, Safari latest, Safari 5.1 (Windows), ipad mini 2, iphone 6s, Android Nexus 6.

Tested using Voiceover on Chrome and Safari.


fofr:
This will need a design review.

nickcolley:
Can you list what devices/browsers (including assistive tech) this has been tested with?

andysellick:
It's gone through numerous iterations with Kate, including a couple of pairing sessions to tweak things. Mia's also seen it.

app/assets/javascripts/govuk-component/tasklist.js
```diff
@@ -0,0 +1,333 @@
+// Most of this is taken from the service manual
+
+(function (Modules) {
+  "use strict";
+  window.GOVUK = window.GOVUK || {};
+
+  Modules.Tasklist = function () {
```
alecgibson: Is the thinking to eventually commonise this with the Accordion?

app/views/govuk_component/tasklist.raw.html.erb
```diff
@@ -0,0 +1,84 @@
...
+  <div
+    data-module="tasklist"
+    class="pub-c-tasklist js-hidden <% if not small %>pub-c-tasklist--large<% end %>"
+    <% if remember_last_section %>data-remember<% end %>
+    >
+    <ol>
+      <% steps.each_with_index do |step, index| %>
+        <%
+          step_active = false
+          step.each do |this_step|
```
alecgibson: How do you enumerate on a single `step`? I find this naming confusing. Why do we have `steps` -> `step` -> `this_step` ?

alecgibson: Also, if this whole loop construct exists to determine `step_active`, then a more idiomatic way to achieve this is something like:

```ruby
step_is_active = step.any? do |this_step|
  this_step[:panel_links].to_a.any? { |s| s[:active] }
end
```

app/views/govuk_component/tasklist.raw.html.erb
```diff
@@ -0,0 +1,84 @@
...
+            if this_step[:panel_links].to_a.any? { |s| s[:active] == true }
+              step_active = true
+            end
+          end
+        %>
+        <li class="pub-c-tasklist__step <% if step_active %>pub-c-tasklist__step--active<% end %>" <% if step_active %>aria-current="step"<% end %>>
```
alecgibson: (I personally prefer removing the need for `end` like so: `class="<%= 'pub-c-tasklist__step--active' if step_active' %>"`)

app/views/govuk_component/tasklist.raw.html.erb
```diff
@@ -0,0 +1,84 @@
...
+  steps ||= false
+  small ||= false
+  heading_level ||= 2
+  open_section ||= false
+  remember_last_section ||= false
+
+  section_count = 0
+%>
+<% if steps %>
+  <div
```
alecgibson: This whole file is hugely nested. Potentially consider breaking up into smaller partials.

app/assets/images/icons-plus-minus.svg
```diff
@@ -0,0 +1,3 @@
+<?xml version="1.0"?>
```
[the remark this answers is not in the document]
andysellick: Funnily enough, I've already started work on a PR in collections that removes this JS, but I've been waiting to get this all approved and merged first. Good point on the images though, hadn't thought of those.

app/assets/javascripts/govuk-component/tasklist.js
```diff
@@ -0,0 +1,333 @@
...
+        buttonText: "Open all",
+        eventLabel: "Open All"
+      },
+      closeAll: {
+        buttonText: "Close all",
+        eventLabel: "Close All"
+      }
+    };
+
+    var rememberOpenSection = 0;
```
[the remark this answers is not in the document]
andysellick: Good suggestion, thanks, will change.

app/views/govuk_component/tasklist.raw.html.erb
andysellick: It is a bit. I'll try some extra line spaces first to see if that helps.

app/views/govuk_component/tasklist.raw.html.erb
```diff
@@ -0,0 +1,84 @@
...
+    data-module="tasklist"
+    class="pub-c-tasklist js-hidden <% if not small %>pub-c-tasklist--large<% end %>"
```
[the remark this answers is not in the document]
andysellick: Will change to 'unless'.

app/views/govuk_component/tasklist.raw.html.erb
andysellick: I knew there must be a cleaner way of doing this. Thank you!

app/views/govuk_component/tasklist.raw.html.erb
andysellick: Again, my Ruby knowledge. That's helpful, thanks. I'm definitely learning..

andysellick:
Fair comment @alecgibson about the duplication of accordion code. This all relates to https://github.com/alphagov/govuk-rfcs/pull/84, if we move components into static it then breaks integration tests. Basically we chose this approach (creating a new comp in static based on the accordion) because it's the quickest and easiest solution for right now. When the RFC is complete we'll hopefully have a better solution and can then rework the tasklist to fit.

app/assets/javascripts/govuk-component/tasklist.js
```diff
@@ -0,0 +1,333 @@
+// Most of this is taken from the service manual
```
fofr: Does this JS differ in any way to the accordion component?
Are the differences only in the class names of elements?
If so, can those class names be abstracted somehow?
Is it appropriate to avoid duplication of this JS?

app/assets/stylesheets/govuk-component/_tasklist-print.scss
```diff
@@ -0,0 +1,20 @@
+// scss-lint:disable SelectorFormat
```
fofr: We need to upgrade the linter on static in a separate PR.

app/assets/stylesheets/govuk-component/_tasklist.scss
```diff
@@ -0,0 +1,358 @@
+.pub-c-tasklist {
```
fofr: Is it possible to write this in a way that shows what's copied from an accordion, and what's specific to a tasklist?

fofr: At the moment I fear it'll be difficult to make parts of this reusable, and too easy for the two patterns to grow out of sync.

app/assets/stylesheets/govuk-component/_tasklist.scss
```diff
@@ -0,0 +1,358 @@
...
+  position: relative;
+  margin-top: $gutter-half;
+  margin-left: $gutter-half;
+  margin-bottom: $gutter;
+}
+
+// scss-lint:disable SelectorFormat
+
+// when js is enabled
```
fofr: You could do `.js-enabled …` rather than use the comment.

app/assets/stylesheets/govuk-component/_tasklist.scss
```diff
@@ -0,0 +1,358 @@
...
+  .pub-c-tasklist--large & {
+    @include media(tablet) {
+      padding-left: $first-indent;
+    }
+  }
+}
+
+.pub-c-tasklist__title {
+  .pub-c-tasklist--active & {
+    padding-left: $second-indent-small;
```
fofr: Could reduce indentation using:
`.pub-c-tasklist--active .pub-c-tasklist__title`

app/assets/stylesheets/govuk-component/_tasklist.scss
```diff
@@ -0,0 +1,358 @@
...
+  background: image-url("icon-plus-minus.svg"), linear-gradient(transparent, transparent); // http://pauginer.com/post/36614680636/invisible-gradient-technique
+  background-repeat: no-repeat;
+
+  .pub-c-tasklist--large & {
+    @include media(tablet) {
+      left: $first-indent;
+      height: $icon-size-large;
+      width: $icon-size-large;
+      margin-top: -($icon-size-large / 2) - 2;
+      // repeating these two lines for the benefit of IE8, which doesn't support background size
```
fofr: Where's the background size line this is referring to?

fofr: Could you indicate why this needs repeating?

app/views/govuk_component/docs/tasklist.yml
```diff
@@ -0,0 +1,342 @@
+name: Tasklist
```
fofr: I prefer "Task list", "tasklist" isn't a word

fofr: "Task list" and `task-list` will be consistent with other list components like:
https://government-frontend.herokuapp.com/component-guide/document-list 
https://govuk-collections.herokuapp.com/component-guide/taxon-list
https://govuk-collections.herokuapp.com/component-guide/topic-list

app/views/govuk_component/docs/tasklist.yml
```diff
@@ -0,0 +1,342 @@
...
+description: Numbered steps containing expanding/collapsing sections
+body: |
+  Tasklists are designed to show a sequence of steps towards a specific goal, such as 'learning to drive'. Each step of a tasklist can contain one or more sections. Each section can contain one or more links to pages. Each section is collapsed by default so that users are not overwhelmed with information.
```
fofr: Should we specifically refer to the research about why task lists are collapsed?

app/views/govuk_component/docs/tasklist.yml
```diff
@@ -0,0 +1,342 @@
...
+  - creating a new component allows further iteration without impacting the accordion
+
+  If JavaScript is disabled the tasklist expands fully. All of the tasklist functionality (including the icons and aria attributes) are added using JavaScript.
+accessibility_criteria: |
+  The tasklist must:
+
+  - indicate to users that each section can be expanded and collapsed
+  - inform the user when a section has been expanded or collapsed
+  - be usable with a keyboard
+  - use a button element for section headings
```
fofr: Think this needs some explanation or rephrasing.

fofr: This is explained below, this line can be deleted.

app/views/govuk_component/docs/tasklist.yml
```diff
@@ -0,0 +1,342 @@
...
+        [
+          {
+            title: 'Second header',
+            id: 'something2',
+            panel: 'Second panel'
+          }
+        ]
+      ]
+  with_different_heading_level:
+    description: Sections have a H2 by default, but this can be changed. The heading level does not change any styling.
```
fofr: :+1:

app/views/govuk_component/docs/tasklist.yml
```diff
@@ -0,0 +1,342 @@
...
+        ],
+        [
+          {
+            title: 'This is also a heading 3',
+            panel: 'Some more content'
+          }
+        ]
+      ]
+  open_a_section_by_default:
+    description: Pass the number of the section of the accordion to open by default.
```
fofr: Pass the _index_?

app/assets/javascripts/govuk-component/tasklist.js
andysellick: It is different (details below) and I think it's going to continue to get more different. In terms of avoiding duplication or building upon the existing accordion, that might be possible except we'd have to separate out the tracking, which is fairly heavily intertwined (the accordion tracks some content clicks based on an assumed content, which isn't ideal - I'm learning more about this all the time).

Differences are:

- 'remember last section opened in URL hash' functionality disabled by default
- new functionality to allow a section to be opened by default

I'm recommending that we remove the first item altogether as it's not needed when the its in the sidebar and having it on the main one only would be inconsistent.

Also, coming soon are some changes to how the GA tracking works.

andysellick: I might update that comment to be clearer..

fofr: The first point
> 'remember last section opened in URL hash' functionality disabled by default

That feature is still needed for users who browse from a task list to a content page and then hit "Back", the section should remain open so they can continue from where they left off.

app/assets/stylesheets/govuk-component/_tasklist.scss
andysellick: I think the tasklist has already deviated significantly from the accordion that keeping them in sync is neither possible or necessarily desirable. In fact, we're currently looking at a design proposal that would make the tasklist look very different. Happy to discuss though as I feel like I'm unaware of the merits involved.

alecgibson: Hmm. I'm also worried that these two look very similar at the moment. At the very least, I'd hope that we could base both components on a "parent" component that encapsulates as much common behaviour as possible (eg expandable sections), and then make it as configurable as we need to meet the needs of both accordions and task lists. I think this is doable if we do more work with blocks, etc.

app/assets/stylesheets/govuk-component/_tasklist.scss
andysellick: I could, yes. I wonder why it wasn't that way in the original accordion? All I'm doing is changing the class name.

app/assets/stylesheets/govuk-component/_tasklist.scss
andysellick: Will clarify.

app/views/govuk_component/docs/tasklist.yml
andysellick: Ok.

app/views/govuk_component/docs/tasklist.yml
andysellick: Will add.

app/assets/javascripts/govuk-component/tasklist.js
andysellick: Two thoughts about that.

Yes, in an ideal world. But the solution doesn't remember all the accordion sections that were opened, only the last one - so the page might look different when they click 'back' anyway.

Behavioural consistency is important, and we're not planning on having this when it's in the sidebar.

app/assets/stylesheets/govuk-component/_tasklist.scss
andysellick: That would be a good solution. Just to let you know I'm now looking at further changes to how the GA stuff works for task lists, which will further change the code involved.

alecgibson: If tracking turns out to be a pain to commonise, a possible solution is to have no tracking by default, and people will have to implement their own tracking on a per-use basis (which could be done through JS).

andysellick: That would be great. I think I'd have done that from the start if the tracking on the accordion hadn't been so detailed and complex.

nickcolley:
You've mentioned this has been tested with VoiceOver, since this is a navigation component that'll be used on a lot of GOV.UK's important journeys, this should be tested with the assistive technologies that are [more commonly used](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices#testing-with-assistive-technologies).

When will this be put in front of users? Has this been through a lab session with users of assistive tech? Again since this is a complicated piece of UI for an important journey it'd be good to have this.

app/views/govuk_component/task_list.raw.html.erb
```diff
@@ -0,0 +1,84 @@
+<%
+  steps ||= false
+  small ||= false
```
nickcolley: Should this be 'size'? (Not a big deal)

nickcolley:
This is looking good, I agree with Alec, this feels like an uneasy amount of duplication of code.

@andysellick can you record this as techincal debt alongside the static/slimmer discussion?

andysellick:
Recorded - https://github.com/alphagov/govuk-rfcs/pull/84#issuecomment-340756025

nickcolley:
Spoke to @sihugh about it breaking when used in alphagov/frontend.

We talked through three approaches.

1. Add reset.css to alphagov/frontend
2. Add additional rules to the component in static to reset default styling at a component basis
3. Add an additional style sheet called 'shame.css' that includes 2. overrides but only in alphagov/frontend

2. Seems the most per convention way forward, since components should not rely on global state to work well.

Out of scope: We'd probably want to raise an issue also on static to consider removing reset.css for other components.

PR #1177, opened 2017-10-31
Deprecate taxonomy sidebar component

fofr:
Depends on https://github.com/alphagov/government-frontend/pull/507

Bring taxonomy links in line with related links, leaving a single
pattern that can be iterated in multivariate tests.

In time, uses of this component will be switched to the related links
component or a new alternative.

* Coerce taxonomy sidebar into related links component
* Use the “More” link to point to taxon pages
* Use the “Elsewhere on GOV.UK” section to group taxon links without children
* Don’t show taxon descriptions

screen_shot_2017-10-30_at_11 58 57

https://trello.com/c/FN3ppmn2/60-update-taxonomy-related-links-in-right-column-to-match-the-design-used-by-mainstream-browse-related-links


fofr:
Closing for now, as this might not be the right approach.

PR #1180, opened 2017-11-02
Add ability to pass rel attributes to buttons

nickcolley:
This is needed for transactions that applies rel="external"

See https://github.com/alphagov/frontend/blob/master/app/views/transaction/show.html.erb#L20

https://govuk-static-pr-1180.herokuapp.com/component-guide/button/start_now_button/preview

### DIFF
--- a/app/views/govuk_component/button.raw.html.erb
+++ b/app/views/govuk_component/button.raw.html.erb
@@ -2,6 +2,7 @@
   start ||= false
   href ||= false
   info_text ||= false
+  rel ||= false
   text ||= ""
   tag_name = href ? "a" : "button"
 %>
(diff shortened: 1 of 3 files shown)


app/views/govuk_component/button.raw.html.erb
```diff
@@ -16,6 +17,10 @@
     href="<%= href.try(:html_safe) %>"
     role="button"
   <% end %>
+
+  <% if rel %>
+    rel="<%= rel.try(:html_safe) %>"
```
fofr: What's the `html_safe` for?

nickcolley: To be honest I was just following the convention for the previous passed arguments.

I think we could remove this since we're expecting plain text strings, I'll pop an update in, thanks Paul 👍

PR #1183, opened 2017-11-02
Allow configuring postcode pii stripping in analytics via meta tag

h-lame:
For: https://trello.com/c/ilkmeS5E/239-investigate-preventing-pii-being-sent-to-ga

This relies on the changes in alphagov/govuk_frontend_toolkit#435 so will need a gemfile bump once that is released.  Our goal here is to make sure we strip PII from all GA data, by default `GOVUK.Analytics` will strip emails, but can be configured to also strip postcodes.  The changes to static make it possible to set that config via a meta tag.  We also add that meta-tag by default if your app is using the analytics_meta_tag component and the content is in a format where we expect postcode data (`smart_answer` or `search` at the moment) might be present.  We also allow passing an argument to the component to turn postcode stripping on or off regardless of the content format.

This is best reviewed commit-by-commit as the early commits concern themselves with tidying up and re-organising the tests for static-analytics to pave the way for these changes.


boffbowsh:
Is this still relevant @h-lame? The tests aren't passing

h-lame:
@boffbowsh It's been a while, but I think the tests on this won't pass until https://github.com/alphagov/govuk_frontend_toolkit/pull/435 is merged, released, and that version included in our Gemfile.

boffbowsh:
@h-lame Can you rebase this please and I'll have a proper look?

h-lame:
For review purposes, look at the last 2 commits only:

* Allow configuring stripping postcode PII in analytics via a meta-tag
* Always add the strip-postcodes meta tag for some content items

I've extracted the rest to their own PR as they're useful regardless of this one.  I will rebase this one on top of that once it's merged to master.

PR #1184, opened 2017-11-03
Fix search button radius in Chrome 62 (Mac)

36degrees:
Chrome 62 [updates the default styles for buttons to add a default `border-radius` of 4px][1], which is affecting the buttons in the search box (noticed on the homepage).

This sets an explicit border radius of `0`, which corrects the search button to look like it does in Chrome 61.

## Before
govuk-static herokuapp com_component-guide_search

## After
govuk-static herokuapp com_component-guide_search 1

[1]: https://www.chromestatus.com/features/5743649186906112

### DIFF
--- a/app/assets/stylesheets/govuk-component/_search.scss
+++ b/app/assets/stylesheets/govuk-component/_search.scss
@@ -25,6 +25,7 @@
     padding: 6px;
     border: 0;
     cursor: pointer;
+    border-radius: 0;
   }
 
   // IE6 + IE7 always get the simplest version, regardless of whether javascript is enabled


36degrees:
I haven't seen any other buttons exhibiting this behaviour - the button component in frontend toolkit already has `border-radius: 0` – and that is not _currently_ commented there.

nickcolley:
Thanks so much for the contribution! Looks a good fix 💯  👍

nickcolley:
It'd be cool to think of this search component as a composition between the text input and a button in the future, which'd avoid this duplication of button styles

PR #1185, opened 2017-11-06
Style taxonomy sidebar like related links

fofr:
We will be iterating both components as part of a multivariate test. This commit avoids a more thorough refactor to make them identical, in favour of making them alike. It’s an alternative to #1177.

* Port related links styles directly from related links component
* Makes two components similar
* Increases spacing around links for easier clicking on mobile
* Adds more examples

This PR does not address:
* Need to use `pub-c` as the prefix
* Inconsistent API between related links and taxonomy sidebar
* Style re-use between components

https://govuk-static-pr-1185.herokuapp.com/component-guide/taxonomy_sidebar

Part of:
https://trello.com/c/FN3ppmn2/60-update-taxonomy-related-links-in-right-column-to-match-the-design-used-by-mainstream-browse-related-links

## Screenshots
Old on left, new on right

screen shot 2017-11-06 at 11 58 46

screen shot 2017-11-06 at 11 59 21

## Thin viewport

screen shot 2017-11-06 at 12 00 20

### DIFF
--- a/app/assets/stylesheets/govuk-component/_taxonomy-sidebar.scss
+++ b/app/assets/stylesheets/govuk-component/_taxonomy-sidebar.scss
@@ -1,38 +1,29 @@
 .govuk-taxonomy-sidebar {
   border-top: 10px solid $mainstream-brand;
-  padding-bottom: $gutter * 2;
+  padding-top: 5px;
   @include core-16;
 
-  .sidebar-taxon {
-    padding-top: 1.25em;
+  h2 {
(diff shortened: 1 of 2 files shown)


fofr:
@andysellick PR aims for consistency with related links rather than perfection in both. Better spacing is a side effect of related links being more spacious.

I have observed users interact with related links on mobile in user research without a problem, but the spacing might not be ideal for everyone.

We'll be doing design iterations on these elements in the next few sprints.

PR #1182, opened 2017-11-02
Improve task list

andysellick:
- fix issues with number vertical alignment and panel top spacing on mobile
- add click tracking for panel links
- code update to rename the parts of a task list, now contains groups each containing steps, instead of steps containing sections

Before:

screen shot 2017-11-02 at 16 26 53

After:

screen shot 2017-11-02 at 16 31 33

Component guide link: https://govuk-static-pr-1182.herokuapp.com/component-guide/task_list

Trello card: https://trello.com/c/hjuxqyuT/275-visual-and-code-tweaks-to-the-task-list

Tested on iphone 6s, Google Nexus 6, iphone 6, Samsung Galaxy s5, ipad air 2, Google nexus 7, Samsung galaxy tab 4, Google Pixel
IE8, IE9, IE10, FF 56, Opera 48, Chrome latest, Safari 5.1, Edge 16

PR #1187, opened 2017-11-06
Revert "Improve task list"

andysellick:
Reverts alphagov/static#1182


fofr:
Why is this being reverted?

sihugh:
We breaking change, so for the period of the cache the task list is borked.  We're going to think about it and come back tomorrow.

PR #1188, opened 2017-11-07
Revert "Revert "Improve task list""

andysellick:
Reverts alphagov/static#1187


sihugh:
We're going to disable the AB test that displays this in the sidebar for the duration of the test, and clear the Rails cache on Collections to get this back up and running quickly.

PR #1191, opened 2017-11-08
Add "Coming to the UK" survey

elenatanasoiu:
For: https://trello.com/c/kjg67jpH/287-coming-to-the-uk-top-tasks-survey

A 1 in 20 survey which will run on pages which have the `Visas and immigration` breadcrumb.

### DIFF
--- a/app/assets/javascripts/surveys.js
+++ b/app/assets/javascripts/surveys.js
@@ -251,6 +251,24 @@
             '^/hmrc-internal-manuals(?:/|$)'
           ]
         }
+      },
+      {
+        identifier: 'cttuk_tt',
+        surveyType: 'url',
+        frequency: 20,
+        startTime: new Date('November 10, 2017').getTime(),
+        endTime: new Date('December 22, 2018 23:59:50').getTime(),
+        url: 'https://www.smartsurvey.co.uk/s/CTTUKsurvey/?c={{currentPath}}',
+        templateArgs: {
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/surveys.js
```diff
@@ -251,6 +251,24 @@
...
+        endTime: new Date('December 22, 2018 23:59:50').getTime(),
+        url: 'https://www.smartsurvey.co.uk/s/CTTUKsurvey/?c={{currentPath}}',
+        templateArgs: {
+          title: 'Help improve GOV.UK',
+          surveyCta: 'Answer 4 quick questions to help us make GOV.UK better.',
+          surveyCtaPostscript: 'This will open a short survey on another website'
+        },
+        activeWhen: {
+          path: [
+            '^/visas-immigration(?:/|$)'
```
h-lame: It's not super clear from the ticket, but I think what we need here is a breadcrumb match rather than a path match.  For example https://www.gov.uk/asylum-screening-centre is a page that is listed on https://www.gov.uk/browse/visas-immigration but the page doesn't have `visas-immigration` in the path.  I reckon we should clarify this with Sean though.

h-lame: I checked with Sean and it's pages linked to from there, so definitely a breadcrumb match, not a path match.

elenatanasoiu: Changed to "breadcrumb"

PR #1193, opened 2017-11-10
Add global cross domain tracking

nickcolley:
This allows a generic method of cross domain tracking for links on GOV.UK.

This will be useful for when our govspeak extension is used in content,
where only content designers can author what goes into the body of the page.

Here we're using native DOM methods rather than jQuery fragments because, for
whatever reason jQuery produces a fragment that does not allow inspection of hostname.
(I think this is because we can't inject a fragment into a real dom)

So rather than add additional overhead to the users' javascript, the tests are a little more verbose.

Will also allow us to replace https://github.com/alphagov/frontend/blob/master/app/views/transaction/_cross_domain_analytics.html.erb

Trello: https://trello.com/c/ppqXaGPk/122-create-a-govspeak-component-for-buttons

### DIFF
--- a/app/assets/javascripts/modules/cross-domain-tracking.js
+++ b/app/assets/javascripts/modules/cross-domain-tracking.js
@@ -1,0 +1,33 @@
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules, global) {
+  'use strict'
+
+  var $ = global.$
+
+  Modules.CrossDomainTracking = function () {
+    this.start = function ($context) {
+      var trackableLinkSelector = '[href][data-tracking-code][data-tracking-name]'
+
+      if ($context.is(trackableLinkSelector)) {
+        addLinkedTrackerDomain($context)
+      } else {
+        $context
+          .find(trackableLinkSelector)
+          .each(function () {
+            addLinkedTrackerDomain($(this))
+          })
+      }
+    }
+
+    function addLinkedTrackerDomain ($element) {
+      var code = $element.attr('data-tracking-code')
+      var name = $element.attr('data-tracking-name')
+      var hostname = $element.prop('hostname')
+
+      if (GOVUK.analytics !== 'undefined') {
+        GOVUK.analytics.addLinkedTrackerDomain(code, name, hostname)
+      }
+    }
+  }
+})(window.GOVUK.Modules, window)
--- a/spec/javascripts/modules/cross-domain-tracking.spec.js
+++ b/spec/javascripts/modules/cross-domain-tracking.spec.js
@@ -1,0 +1,97 @@
+describe('Cross Domain Tracking', function () {
+  'use strict'
+  var module
+
+  beforeEach(function () {
+    GOVUK.analytics = GOVUK.analytics || {}
+    GOVUK.analytics.addLinkedTrackerDomain = function () {
+    }
+
+    spyOn(GOVUK.analytics, 'addLinkedTrackerDomain')
+    module = new GOVUK.Modules.CrossDomainTracking()
+  })
+
+  it('tracks realistic example', function () {
+    var anchorToTest = document.createElement('a')
+    anchorToTest.href = 'https://www.registertovote.service.gov.uk/register-to-vote/start'
+    anchorToTest.className = 'button button--start'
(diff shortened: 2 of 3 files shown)


nickcolley:
Talked to @fofr he suggested implementing this as a 'module'

https://github.com/alphagov/govuk_frontend_toolkit/blob/master/docs/javascript.md#modules

e.g. `<div data-module=cross-domain-tracking data-tracking-id=1234>`

This seems nicer, going to refactor.

app/assets/javascripts/modules/cross-domain-tracking.js
```diff
@@ -0,0 +1,31 @@
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  'use strict'
+  Modules.CrossDomainTracking = function () {
+    this.start = function ($context) {
+      var trackableLinkSelector = '[href][data-tracking-code][data-tracking-name]';
+      var $element;
```
fofr: Any reason $context isn't $element? It looks like `$element` isn't being used in this scope.

spec/javascripts/modules/cross-domain-tracking.spec.js
```diff
@@ -0,0 +1,97 @@
...
+
+  it('tracks realistic example', function () {
+    var anchorToTest = document.createElement('a')
+    anchorToTest.href = 'https://www.registertovote.service.gov.uk/register-to-vote/start'
+    anchorToTest.className = 'button button--start'
+    anchorToTest.setAttribute('role', 'button')
+    anchorToTest.setAttribute('data-module', 'cross-domain-tracking')
+    anchorToTest.setAttribute('data-tracking-code', 'UA-23066786-5')
+    anchorToTest.setAttribute('data-tracking-name', 'transactionTracker')
+    anchorToTest.textContent = 'Start Now'
```
fofr: Would this be easier to read if the test anchor was setup as a simple string, like:
`"<a href="https://www.registertovote.service.gov.uk/register-to-vote/start" …>Start now</a>"`

app/assets/javascripts/modules/cross-domain-tracking.js
fofr: Like https://github.com/alphagov/static/blob/master/spec/javascripts/modules/track-click.spec.js#L33-L40

I think this makes it clearer in the tests how to use the module.

spec/javascripts/modules/cross-domain-tracking.spec.js
```diff
@@ -0,0 +1,97 @@
...
+      'UA-23066786-5'
+    )
+    secondAnchorToTest.setAttribute(
+      'data-tracking-name',
+      'transactionTracker'
+    )
+
+    var wrapperDiv = document.createElement('div')
+    wrapperDiv.appendChild(anchorToTest)
+    wrapperDiv.appendChild(secondAnchorToTest)
```
fofr: Again, some markup like the following would make it clearer what the test is doing and the expected markup needed for the module to work:

```
<div>
  <a href="#"></a>
  <a href="#"></a>
</div>
```

fofr: Like: https://github.com/alphagov/static/blob/master/spec/javascripts/modules/track-click.spec.js#L140-L156

spec/javascripts/modules/cross-domain-tracking.spec.js
nickcolley: These tests do not pass if given a string, I think trying to get the 'hostname' of a link in a jQuery fragment does not work since it's not been created in a real dom.

I reference this in the commit message https://github.com/alphagov/static/pull/1193/commits/de832f6bd625e53de4f5cf54074a6ef666f03f4d

spec/javascripts/modules/cross-domain-tracking.spec.js
nickcolley: Same as above

app/assets/javascripts/modules/cross-domain-tracking.js
nickcolley: Looks like I left this in, I should configure standard.js in my editor!

PR #1196, opened 2017-11-15
Update task list tracking

andysellick:
- add task list size ('Big' or 'Small') to Open all / Close all clicks eventLabel
- do the same for panel links i.e. links in lists in the accordion-like bits of the task list

Trello: https://trello.com/c/pLUMn48c/260-check-the-all-data-is-rolling-in-from-the-learning-to-drive-a-car-beta-%F0%9F%9A%97


andysellick:
Awesome, thanks @maxgds

PR #1208, opened 2017-12-05
Add margin bottom support to buttons

nickcolley:
By default buttons should have no margin bottom,
but sometimes it's useful to break up a page, for example if a button is at the bottom of a page.

### DIFF
--- a/app/views/govuk_component/button.raw.html.erb
+++ b/app/views/govuk_component/button.raw.html.erb
@@ -5,6 +5,7 @@
   rel ||= false
   text ||= ""
   tag_name = href ? "a" : "button"
+  margin_bottom ||= false
 %>
 <<%= tag_name %>
   class="
@@ -12,6 +13,7 @@
     <% if start == true %>
       pub-c-button--start
     <% end %>
+    <%= " pub-c-button--bottom-margin" if margin_bottom %>
   "
   <% if href %>
     href="<%= href.try(:html_safe) %>"
--- a/app/views/govuk_component/docs/button.yml
+++ b/app/views/govuk_component/docs/button.yml
@@ -49,6 +49,7 @@ examples:
       start: true
       info_text: "Sometimes you want to explain where a user is going to."
   with_margin_bottom:
+    description: "Sometimes it's useful to break up a page, for example if a button is at the bottom of a page."
     data:
       text: "Submit"
       margin_bottom: true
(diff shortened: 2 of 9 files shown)


app/views/govuk_component/button.raw.html.erb
```diff
@@ -5,13 +5,15 @@
   rel ||= false
   text ||= ""
   tag_name = href ? "a" : "button"
+  margin_bottom ||= false
 %>
 <<%= tag_name %>
```
maxgds: What's going on here with the double <? Is it some magic syntax I'm not aware of, or a mistake?

test/govuk_component/button_test.rb
```diff
@@ -55,4 +55,13 @@ def component_name
     render_component(text: "Start now", rel: "nofollow preload")
     assert_select ".pub-c-button[rel='nofollow preload']", text: "Start now"
   end
+
+  test "renders margin bottom class correctly" do
+    render_component(text: "Submit")
+    assert_select ".pub-c-button", text: "Submit"
```
maxgds: I'm not sure this line is necessary, if the test fails at this point it doesn't prove anything to do with the margin class but a different fail in the component, I'd have thought the selector test on its own was enough.

app/views/govuk_component/button.raw.html.erb
nickcolley: There's no changes to this in this PR. This allows us to make this 'button' into an anchor or a button depending on what is passed to it.

test/govuk_component/button_test.rb
nickcolley: This ensures the component renders as expected, with the next assertion making sure that it is then applied the right class.

Without this assertion you could in theory only render the modifier class, or this option could break how the text renders etc.

maxgds: I'd argue that your previous tests test that the component renders ok and it should fail earlier against relevant cases, but I'm not going to argue that hard :) Just personal preference, really.

nickcolley: If I were to introduce behaviour that removed the text output *only* when passing the 'margin_bottom' option, this test would fail.

If I didn't test this, it would not fail.

Ideally we'd be doing a form of [snapshot testing](https://facebook.github.io/jest/docs/en/snapshot-testing.html)

app/views/govuk_component/docs/button.yml
```diff
@@ -48,6 +48,10 @@ examples:
       href: "#"
       start: true
       info_text: "Sometimes you want to explain where a user is going to."
+  with_margin_bottom:
+    data:
```
andysellick: Maybe worth adding a description here, similar to the description of this PR? Just to explain why this option is here.

PR #1221, opened 2017-12-11
Remove task list components

andysellick:
These components are being moved to the govuk_publishing_components gem.

The components are being updated in the gem to use a different namespace so no CSS/JS conflicts should occur if we deploy using the gem's components with these ones still in static. However once that process has been done we should deploy this change to remove the duplications.

Trello card: https://trello.com/c/ELorPDMS/393-do-all-the-task-list-gem-component-things

Related PR to add components into the gem: https://github.com/alphagov/govuk_publishing_components/pull/105


vanitabarrett:
@andysellick Do we also need to remove the following:

- [current-location.js](https://github.com/alphagov/static/blob/master/app/assets/javascripts/current-location.js)
- [history-support.js](https://github.com/alphagov/static/blob/master/app/assets/javascripts/history-support.js)

PR #1222, opened 2017-12-11
Update taxonomy sidebar design

steventux:
https://trello.com/c/zzot5dZq/157-taxonomy-sidebar-add-link-to-collections
https://trello.com/c/NtD5rC4S/121-taxonomy-sidebar

screenshot from 2017-12-12 16-36-52

This combines the changes from https://github.com/alphagov/static/pull/1200 to make the taxonomy sidebar styles similar to the related-navigation component and also the feature of rendering collections links below the taxon links.

### DIFF
--- a/app/assets/stylesheets/govuk-component/_taxonomy-sidebar.scss
+++ b/app/assets/stylesheets/govuk-component/_taxonomy-sidebar.scss
@@ -1,29 +1,37 @@
 .govuk-taxonomy-sidebar {
-  border-top: 10px solid $mainstream-brand;
-  padding-top: 5px;
+  border-top: 2px solid $govuk-blue;
+  padding-top: $gutter-half;
   @include core-16;
 
-  h2 {
-    @include bold-24;
-    margin-top: 0.3em;
-    margin-bottom: 0.5em;
+  .govuk-taxonomy-sidebar__heading {
+    margin-top: 0;
+    margin-bottom: $gutter-one-third;
+    @include bold-19;
   }
 
   .taxon-description {
-    margin-bottom: 0.75em;
+    margin-top: 0;
+    margin-bottom: $gutter-one-third;
+  }
+
+  a {
+    text-decoration: none;
+  }
+
+  nav {
+    margin-bottom: $gutter;
+  }
+
+  li {
+    padding: 0;
+    list-style-type: none;
+    margin-bottom: $gutter-one-third;
   }
 
   ul {
     // reset the default browser styles
     padding: 0;
     margin: 0;
-    list-style: none;
     margin-bottom: 1.25em;
-
-    li {
-      // reset the default browser styles
-      padding: 0;
-      margin-bottom: 0.75em;
-    }
-  }
+ }
 }
--- a/app/assets/stylesheets/govuk-component/_taxonomy-sidebar.scss
+++ b/app/assets/stylesheets/govuk-component/_taxonomy-sidebar.scss
@@ -33,5 +33,16 @@
     padding: 0;
     margin: 0;
     margin-bottom: 1.25em;
- }
+  }
+
+  .govuk-taxonomy-sidebar__collections-heading {
+    border-top: 1px solid $border-colour;
+    margin-bottom: $gutter-half;
+    margin-top: $gutter;
+    padding-top: $gutter-half;
+  }
+
+  .govuk-taxonomy-sidebar__collections-link {
+    @include bold-19;
+  }
 }
--- a/app/assets/stylesheets/govuk-component/_taxonomy-sidebar.scss
+++ b/app/assets/stylesheets/govuk-component/_taxonomy-sidebar.scss
@@ -1,48 +1,50 @@
 .govuk-taxonomy-sidebar {
+  @include core-16;
   border-top: 2px solid $govuk-blue;
   padding-top: $gutter-half;
-  @include core-16;
+}
 
-  .govuk-taxonomy-sidebar__heading {
-    margin-top: 0;
-    margin-bottom: $gutter-one-third;
-    @include bold-19;
-  }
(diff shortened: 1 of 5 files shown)


app/assets/stylesheets/govuk-component/_taxonomy-sidebar.scss
```diff
@@ -1,29 +1,48 @@
...
+    @include bold-19;
   }
 
   .taxon-description {
-    margin-bottom: 0.75em;
+    margin-top: 0;
+    margin-bottom: $gutter-one-third;
+  }
+
+  a {
```
maxgds: We usually wouldn't style an "a" directly. Sometimes we are forced to, but we absolutely need to ensure we're not styling all a tags on the page, so 
.taxon-description a { blah } 
or something may be necessary - ideally though the anchor would have a styling class such as .govuk-taxonomy-sidebar__link

app/assets/stylesheets/govuk-component/_taxonomy-sidebar.scss
```diff
@@ -1,29 +1,48 @@
...
+    text-decoration: none;
+  }
+
+  nav {
```
maxgds: again, this needs to be scoped

app/assets/stylesheets/govuk-component/_taxonomy-sidebar.scss
```diff
@@ -1,29 +1,48 @@
...
+    margin-bottom: $gutter;
+  }
+
+  li {
```
maxgds: scoping needed

app/assets/stylesheets/govuk-component/_taxonomy-sidebar.scss
```diff
@@ -1,29 +1,48 @@
...
+    padding: 0;
+    list-style-type: none;
+    margin-bottom: $gutter-one-third;
   }
 
   ul {
```
maxgds: scope

app/assets/stylesheets/govuk-component/_taxonomy-sidebar.scss
nickcolley: These seem to be nested within the selector at the top, but massive 👍 to always avoiding styling base elements if possible.

We should only be doing this if there's no other way as it directly conflicts with our naming conventions that allow us to have confidence with components across so many applications.

https://github.com/alphagov/govuk_publishing_components/blob/master/docs/component_conventions.md

nickcolley: Specifically we're using [BEM](https://github.com/alphagov/govuk_publishing_components/blob/master/docs/component_conventions.md#bem) and [namespacing](https://github.com/alphagov/govuk_publishing_components/blob/master/docs/component_conventions.md#namespacing)

steventux:
@maxgds @nickcolley I'm confused about requested changes now, could you clarify what, if anything, needs changing on this PR please?

maxgds:
I hadn't noticed the things I commented on were nested, as such they are they are safe to go out.

However, we shouldn't really be applying the css directly to elements even when embedded. I'd say mark it as tech debt, put it on the board in the tech debt column Vanita created and let this go out as-is for now.

maxgds:
The diff is confusing - Steve hasn't actually introduced this, hence my comment. And it's relatively safe as it's nested in the scss, not completely bare.

steventux:
@nickcolley forgive my ignorance, but it's not clear which changes are necessary to make this PR acceptable production-ready component code.
It seems like some unconventional CSS scoping from previous work is being conflated with the intended scope of this PR (some cursory design changes and adding the collections links, none of which add unscoped styles).
Is the fixing of the unconventional CSS a blocker to any other work on this component? If so can you detail what remedial work needs doing please? I'm not certain what _we should just make the updates in this review_ means in terms of fixing scope.

nickcolley:
@steventux if we're writing new CSS we should be using the correct naming conventions (linked above). Old code is fine, and is covered by https://trello.com/c/SCio6yb4/45-static-components-need-updating-to-new-component-principles

steventux:
My bad - had cherry-picked a commit from https://github.com/alphagov/static/pull/1200, was unaware this contained the offending CSS

steventux:
@nickcolley @maxgds hopefully this https://github.com/alphagov/static/pull/1222/commits/fa16a2138443603c9027512fabb8b7b3345fc766 sorts out the element selectors. Let me know if I'm off with the BEM naming.

steventux:
@nickcolley I've amended the component styles so they follow the BEM naming convention, apologies @maxgds for not understanding the original review comments.
Anything else to do on this PR?

app/assets/stylesheets/govuk-component/_taxonomy-sidebar.scss
```diff
@@ -1,29 +1,52 @@
 .govuk-taxonomy-sidebar {
-  border-top: 10px solid $mainstream-brand;
-  padding-top: 5px;
   @include core-16;
+  border-top: 2px solid $govuk-blue;
+  padding-top: $gutter-half;
+}
+
+.govuk-taxonomy-sidebar__related-heading,
```
nickcolley: Non blocking:

These classes look to be the same, if we decouple how we present parts of this component from the semantics of this component, could these be change to a single '.govuk-taxonomy-sidebar__heading`?

steventux:
Thanks @nickcolley your comment makes sense to me, I've renamed the heading and link classes to simplify: https://github.com/alphagov/static/pull/1222/commits/dd4211811e9342ba8d5952d256b8655ef2ff1303#diff-b2af3eb92cd75ba5d8c0d1ab14ce7424R7
I will think of a better name for the collections classes as it suffers from the same naming problem.

PR #1237, opened 2017-12-21
Add mobile TLS survey

elenatanasoiu:
For: https://trello.com/c/quB5X0AU/314-tls-survey

## Background

There are a lot of people visiting gov.uk with out of date security. A lot of them come from mobile. 
We want to display a survey to ask them more questions. 
There is a different survey URL for desktop and mobile.

## What this does 

1. Adds a TLS survey that will only be displayed for people with an out of date TLS version. 
This version number is stored in a `TLSversion` cookie, if present. When that version is < 1.2 we display the TLS survye.  

2. Adds the ability to display surveys on mobile. For the TLS survey, we also serve a different survey URL for mobile than we do for desktop.


app/assets/stylesheets/helpers/_header.scss
```diff
@@ -245,8 +245,21 @@
   @include core-16;
 
   &.visible {
-    @include media(desktop) {
-      display: block;
+    display: block;
+  }
+
+  &.mobile-survey {
```
h-lame: Can we do this without introducing a new class?  The design card (see: https://trello.com/c/jDmt8lpf/146-mobile-survey-design) for this story has versions of both url and email surveys, so I think we are good to let both types of survey appear on mobile (and the CSS changes here are probably enough to cover both designs).

Could we get away with dropping this `&.mobile-survey` rule, and having it's contents apply to the entire `&.visible` block ?

app/assets/javascripts/surveys.js
```diff
@@ -140,6 +140,15 @@
     }
   }
 
+  var tlsSurveyUrl = function () {
+    if (window.matchMedia( "(max-width: 800px)" ).matches) {
```
h-lame: Is there any chance that there is something already on the page that can tell us if something is / isn't mobile?  Wary of having this `max-width: 800px` conditional get out of date with conditionals we use for real CSS styling.

app/assets/javascripts/surveys.js
```diff
@@ -613,13 +641,18 @@
...
 
     isSurveyToBeDisplayed: function (survey) {
-      if (GOVUK.cookie(userSurveys.surveyTakenCookieName(survey)) === 'true') {
-        return false
-      } else if (userSurveys.surveyHasBeenSeenTooManyTimes(survey)) {
+      if (userSurveys.isBeingViewedOnMobile() && !userSurveys.surveyIsAllowedOnMobile(survey)) {
         return false
       } else {
-        return userSurveys.randomNumberMatches(survey.frequency)
+        if (GOVUK.cookie(userSurveys.surveyTakenCookieName(survey)) === 'true') {
```
h-lame: Minor formatting nitpick: I reckon this could be:

```js
} else if ... {
```

rather than:

```js
} else {
  if ...
}
```

app/assets/javascripts/surveys.js
```diff
@@ -720,6 +753,14 @@
       return generateCookieName('survey_seen_' + survey.identifier)
     },
 
+    isBeingViewedOnMobile: function () {
+      return window.matchMedia( "(max-width: 800px)" ).matches
```
h-lame: Same comment as above, but if not, can we somehow make this conditional, and the one in `tlsSurveyUrl` match up?  A global constant for the query or something?

spec/javascripts/surveys-spec.js
```diff
@@ -1032,28 +1083,79 @@ describe('Surveys', function () {
...
 
   describe('getDisplayableSurveys', function () {
-    it('does not include a survey that should not be displayed', function () {
-      spyOn(surveys, 'isSurveyToBeDisplayed').and.returnValue(false)
-      var testSurvey = {
-        startTime: new Date('July 5, 2016').getTime(),
-        endTime: new Date('July 10, 2016 23:50:00').getTime(),
-        url: 'example.com/small-survey'
-      }
+    describe('when the page is viewed on desktop', function () {
```
h-lame: I don't think we need to make changes to this test.  The implementation only relies on the result of `isSurveyToBeDisplayed` and the tests (for better or worse) mirror that via the `spyOn` calls to `isSurveyToBeDisplayed`.  It's saying, there's a bunch of logic in `isSurveyToBeDisplayed` and all this method cares about is true or false coming out of that, if you want to know more look at the tests for `isSurveyToBeDisplayed` (for which you've written a bunch more tests, yay!).  To me it feels like adding the `isBeingViewsOnMobile` stuff is redundant in this context.

app/assets/javascripts/surveys.js
```diff
@@ -795,7 +847,20 @@
     currentBreadcrumb: function () { return $('.govuk-breadcrumbs').text() || '' },
     currentSection: function () { return $('meta[name="govuk:section"]').attr('content') || '' },
     currentThemes: function () { return $('meta[name="govuk:themes"]').attr('content') || '' },
-    currentOrganisation: function () { return $('meta[name="govuk:analytics:organisations"]').attr('content') || '' }
+    currentOrganisation: function () { return $('meta[name="govuk:analytics:organisations"]').attr('content') || '' },
+    currentTlsVersion: function () {
```
h-lame: As there's a decent amount of logic going on in this method I think we should have some tests around it.

app/assets/javascripts/surveys.js
h-lame: Actually, from chatting to #frontend folks it seems there isn't anything in our JS bundle that does this already, so this is probably fine.  It's also likely to be temporary as eventually all our surveys should on mobile.  Let's leave this as-is.

PR #1289, opened 2018-02-13
Add tracking attributes to button component

steventux:
Part of https://trello.com/c/3VwTvDrj/256-use-button-component-in-frontend-formats

Updates the button component to accept `extra_attrs`, primarily to support data attributes for tracking eg. https://github.com/alphagov/frontend/blob/master/app/views/transaction/show.html.erb#L22
This should allow us to use the component for Frontend app buttons.

I've refactored this to use Rails `link_to` and `button_tag` helpers as it seemed less noisy.

### DIFF
--- a/app/views/govuk_component/docs/button.yml
+++ b/app/views/govuk_component/docs/button.yml
@@ -62,3 +62,12 @@ examples:
       text: "I'm a start now button with lots of text to test how the component scales at extremes."
       start: true
       href: "#"
+  with_data_attributes:
+    data:
+      text: "Track this!"
+      margin_bottom: true
+      data_attributes: {
+        "module": "cross-domain-tracking",
+        "tracking-code": "GA-123ABC",
+        "tracking-name": "transactionTracker"
+      }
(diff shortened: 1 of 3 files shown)


app/views/govuk_component/docs/button.yml
```diff
@@ -62,3 +62,12 @@ examples:
       text: "I'm a start now button with lots of text to test how the component scales at extremes."
       start: true
       href: "#"
+  with_data_attributes:
+    data:
+      text: "Track this!"
+      margin_bottom: true
+      extra_attrs: {
```
fofr: There's an existing pattern for passing in data attributes:
https://github.com/alphagov/government-frontend/blob/master/app/views/components/_document-list.html.erb#L13

I wonder how we can document/standardise these patterns.

fofr:
Nice refactor, one minor comment about how the data attrs are passed in.

steventux:
@fofr updated to use the same pattern you've referenced, also added tests for links and buttons. Will update https://github.com/alphagov/frontend/pull/1402

PR #1300, opened 2018-02-20
Add specialist user/super group survey

steventux:
https://trello.com/c/z22WEjzo/277-put-specialist-user-supergroup-subgroup-study-live-on-govuk

This will display for 1 in 3 users on selected specialist
finder and document pages. It takes the user to an Optimal
Workshop task.

screenshot from 2018-02-20 16-11-57

### DIFF
--- a/app/assets/javascripts/surveys.js
+++ b/app/assets/javascripts/surveys.js
@@ -123,6 +123,11 @@
     }
   }
 
+  var specialistUserGroupSurveysFeb2018Url = function() {
+    var path = window.location.pathname;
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/surveys.js
```diff
@@ -266,6 +271,28 @@
         },
         allowedOnMobile: true
       },
+      {
+        identifier: 'specialist_user_groups',
+        surveyType: 'url',
+        frequency: 3,
+        startTime: new Date('February 19, 2018').getTime(),
```
emmabeynon: We can change this to February 20, as the 19th has already passed.

steventux:
@emmabeynon I've updated the start date and rebased with master.

PR #1303, opened 2018-02-21
Add title attribute to button component

steventux:
Title is a default HTML attribute and we do make use of this attribute in completed transactions.

### DIFF
--- a/app/views/govuk_component/button.raw.html.erb
+++ b/app/views/govuk_component/button.raw.html.erb
@@ -6,6 +6,7 @@
   text ||= ""
   margin_bottom ||= false
   data_attributes ||= false
+  title ||= false
   css_classes = %w(pub-c-button)
   css_classes << "pub-c-button--start" if start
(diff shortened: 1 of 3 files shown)


selfthinker:
Under which circumstances do we use titles on buttons or links? I would generally discourage from using it because it is not accessible to many people. Most users wouldn't know how to access it and on mobile devices it's completely inaccessible.
What's the user need?

steventux:
@selfthinker https://github.com/alphagov/frontend/blob/master/app/views/completed_transaction/show.html.erb#L19 uses the title attribute on a button (link).
We'd like to use the button component in frontend formats, this PR provides support for the use of this (global) attribute.

selfthinker:
I tried to find what the reason is for that particular use of the title attribute within the git history, but it was there from the beginning. I assume it was added to let the link make sense out of context for screen readers. That would be fine. But it could be easily misused.

fofr:
@steventux @selfthinker Could we include an accessibility acceptance criteria about appropriate use of the title param?

PR #1315, opened 2018-03-02
Add new taxonomy survey CTTUK_iteration 8

cbaines:
https://trello.com/c/CI3NujXJ/157-test-v8-of-cttuk-using-treejack-and-govuk-survey-banner

### DIFF
--- a/app/assets/javascripts/surveys.js
+++ b/app/assets/javascripts/surveys.js
@@ -154,8 +154,8 @@
         identifier: 'CTTUK_iteration_8',
         surveyType: 'url',
         frequency: 3,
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/surveys.js
```diff
@@ -305,6 +305,39 @@
           ]
         },
         allowedOnMobile: true
+      },
+      {
+        identifier: 'CTTUK_iteration 8',
```
tomsabin: This is the only thing that I am uncertain about that perhaps we should clarify whether the identifier should be `CTTUK_iteration_8`. The field seems to be only used for analytics, so perhaps this needs to link to something there?

PR #1318, opened 2018-03-06
Add inverse flag to lead paragraph component

vanitabarrett:
We need to be able to display the lead paragraph component as white text on a blue background for the new taxonomy topic page designs.

screen shot 2018-03-06 at 11 29 25

[Component Guide link](https://govuk-static-pr-1318.herokuapp.com/component-guide/lead_paragraph)


vanitabarrett:
@maxgds Good spot, I'll add that now 👍

PR #1322, opened 2018-03-07
Add inverse flag to breadcrumbs component

maxgds:
This allows breadcrumbs to be rendered in white so they can be used on blue backgrounds in Topic pages. 
screen shot 2018-03-07 at 14 14 44

Component guide: https://govuk-static-pr-1322.herokuapp.com/component-guide/breadcrumbs

### DIFF
--- a/app/assets/stylesheets/govuk-component/_breadcrumbs.scss
+++ b/app/assets/stylesheets/govuk-component/_breadcrumbs.scss
@@ -34,3 +34,8 @@
     }
   }
 }
+
+.breadcrumb-for-current-page.govuk-breadcrumbs--inverse,
+.govuk-breadcrumbs .govuk-breadcrumbs--inverse {
+  color: $white;
+}
--- a/app/views/govuk_component/breadcrumbs.raw.html.erb
+++ b/app/views/govuk_component/breadcrumbs.raw.html.erb
@@ -1,18 +1,21 @@
 <%
   breadcrumbs ||= []
   collapse_on_mobile ||= false
+  inverse ||= false
+  collapse_class =  collapse_on_mobile && breadcrumbs.any? { |crumb| crumb[:is_page_parent] } ? "collapse-on-mobile" : ""
+  invert_class = inverse ? "govuk-breadcrumbs--inverse" : ""
 %>
 
-<div
-  class="govuk-breadcrumbs <%= "collapse-on-mobile" if collapse_on_mobile && breadcrumbs.any? { |crumb| crumb[:is_page_parent] } %>"
-  data-module="track-click">
+<div class="govuk-breadcrumbs <%= collapse_class %>" data-module="track-click">
   <ol>
   <% breadcrumbs.each_with_index do |crumb, index| %>
     <%
       is_link = crumb[:url].present? || crumb[:is_current_page]
       path = crumb[:is_current_page] ? '#content' : crumb[:url]
       aria_current = crumb[:is_current_page] ? 'page' : 'false'
-      css_class = crumb[:is_current_page] ? 'breadcrumb-for-current-page' : ''
+      # css_class = crumb[:is_current_page] ? 'breadcrumb-for-current-page ' : ''
+      # css_class << invert_class
+      css_class = invert_class.concat(crumb[:is_current_page] ? ' breadcrumb-for-current-page ' : '')
     %>
 
     <li class='<%= "parent-breadcrumb" if crumb[:is_page_parent] %>'>
--- a/app/views/govuk_component/docs/breadcrumbs.yml
+++ b/app/views/govuk_component/docs/breadcrumbs.yml
@@ -4,6 +4,9 @@ body: |
   Accepts an array of breadcrumb objects. Each crumb must have a title and a URL.
 shared_accessibility_criteria:
   - link
+accessibility_criteria:
+  The breadcrumb links must have a text contrast ratio higher than 4.5:1 against the background colour to meet WCAG AA
+  (this especially applies when [using the inverse flag](/component-guide/breadcrumbs/inverse)).
 examples:
   default:
     data:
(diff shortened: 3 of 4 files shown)


app/assets/stylesheets/govuk-component/_breadcrumbs.scss
```diff
@@ -34,3 +35,11 @@
     }
   }
 }
+
+.govuk-breadcrumbs--inverse a {
+  color: $white;
+}
+
+.breadcrumb-for-current-page--inverse a {
```
vanitabarrett: This is being applied to the `a` element, so we don't need the `a` part of the selector.

app/views/govuk_component/breadcrumbs.raw.html.erb
```diff
@@ -1,18 +1,20 @@
...
-  data-module="track-click">
+<div class="govuk-breadcrumbs <%= invert_class %> <%= collapse_class %>" data-module="track-click">
   <ol>
   <% breadcrumbs.each_with_index do |crumb, index| %>
     <%
       is_link = crumb[:url].present? || crumb[:is_current_page]
       path = crumb[:is_current_page] ? '#content' : crumb[:url]
       aria_current = crumb[:is_current_page] ? 'page' : 'false'
       css_class = crumb[:is_current_page] ? 'breadcrumb-for-current-page' : ''
+      css_class << 'breadcrumb-for-current-page--inverse' if inverse && crumb[:is_current_page]
```
vanitabarrett: This currently results in an element getting the 2 classes without any space between them, e.g:

`breadcrumb-for-current-pagebreadcrumb-for-current-page--inverse`

vanitabarrett: I think we can simplify this to get `govuk_breadcrumbs--inverse` to apply to links and current links. We could:

- Continue to construct the inverse class at the top
- Instead of applying it to the overall govuk-breadcrumbs div, apply it to each link
- This would replace the last line with:

`css_class << invert_class`

We would then need to add the following to the CSS, under the current `breadcrumb-for-current-page` rule

```
  .govuk-breadcrumbs--inverse,
  .govuk-breadcrumbs--inverse a {
    color: $white;
  }
```

app/views/govuk_component/docs/breadcrumbs.yml
```diff
@@ -4,6 +4,9 @@ body: |
   Accepts an array of breadcrumb objects. Each crumb must have a title and a URL.
 shared_accessibility_criteria:
   - link
+accessibility_criteria:
+  The breadcrumb links must have a text contrast ratio higher than 4.5:1 against the background colour to meet WCAG AA
+  (this especially applies when using the inverse flag).
```
vanitabarrett: We could link to the inverse flag example here.

app/views/govuk_component/docs/breadcrumbs.yml
```diff
@@ -12,6 +15,19 @@ examples:
         url: '/section'
       - title: 'Sub-section'
         url: '/section/sub-section'
+  inverse:
+    description: On a dark background
```
vanitabarrett: It's useful to say where this will be used in the description, e.g: on topic pages. When the topic pages go live, we can then link to a live page here so the component can be seen in context.

app/views/govuk_component/breadcrumbs.raw.html.erb
maxgds: See the latest commit - I've implemented this, but had to be more specific with the css to override what we're pulling in previously, which is starting to get a bit ugly - seemed better than throwing !important all over it, though. Thoughts? If you're happy with that css I'll squash the commit.

maxgds:
@vanitabarrett Implemented suggested changes. Started out having to get more specific with the class selectors in order to override the css we pull in from elsewhere - it was ugly and the linter shouted, so I've switched to using !important which I'm not thrilled about but it works. Could you take another look?

vanitabarrett:
@maxgds The linter was complaining about use of `a` before the class selector. I think you can do the following instead and that will also mean you don't have to use `!important`:

```
.breadcrumb-for-current-page.govuk-breadcrumbs--inverse,
.govuk-breadcrumbs .govuk-breadcrumbs--inverse {
  color: $white;
}
```

app/assets/stylesheets/govuk-component/_breadcrumbs.scss
```diff
@@ -34,3 +34,12 @@
     }
   }
 }
+
+.breadcrumb-for-current-page--inverse {
```
vanitabarrett: We don't need this any more as the class isn't being used anywhere

maxgds: Thanks! Amended!

app/views/govuk_component/breadcrumbs.raw.html.erb
```diff
@@ -1,18 +1,20 @@
...
+<div class="govuk-breadcrumbs <%= collapse_class %>" data-module="track-click">
   <ol>
   <% breadcrumbs.each_with_index do |crumb, index| %>
     <%
       is_link = crumb[:url].present? || crumb[:is_current_page]
       path = crumb[:is_current_page] ? '#content' : crumb[:url]
       aria_current = crumb[:is_current_page] ? 'page' : 'false'
-      css_class = crumb[:is_current_page] ? 'breadcrumb-for-current-page' : ''
+      css_class = crumb[:is_current_page] ? 'breadcrumb-for-current-page ' : ''
+      css_class << invert_class
```
vanitabarrett: Now that this is always applied, we can probably these 2 lines into one:

```
invert_class.concat(crumb[:is_current_page] ? ' breadcrumb-for-current-page ' : '')
```

maxgds: Done that, too. Thanks!

PR #1326, opened 2018-03-09
Remove the CTTUK_iteration_8 survey

cbaines:
It has ended.


tomsabin:
This has been replaced with the CTTUK v10 survey.

PR #1338, opened 2018-03-14
Reduce frequency of content pages design surveys

maxgds:
Surveys currently appear too frequently, this PR reduces that.


vanitabarrett:
Do we also need to reduce the frequency of `GOVUK_content_pages_design_survey_2` or is it just `GOVUK_content_pages_design_survey_1`?

maxgds:
Good spot, it should be both. Will fix.

maxgds:
@vanitabarrett now applied to both surveys

vanitabarrett:
@maxgds Changes all look good to me, not sure what the Jenkins failure means though.

PR #1352, opened 2018-03-22
Add stepnavs to GA custom dimension 96

selfthinker:
A page can be part of none, one or multiple step by step navigations.
The GA custom dimension 96 should include all step navs the current page is part of in order to be able to differentiate specific step nav elements which might appear in more than one step nav.

[Trello card](https://trello.com/c/QCdTZ8cg)

### DIFF
--- a/app/views/govuk_component/analytics_meta_tags.raw.html.erb
+++ b/app/views/govuk_component/analytics_meta_tags.raw.html.erb
@@ -104,7 +104,7 @@
 
   stepnavs = links_hash[:part_of_step_navs] || []
   stepnavs_content = stepnavs.map { |stepnav| stepnav[:content_id] }.join(",")
-  meta_tags["govuk:stepnavs"] = stepnavs_content if stepnavs_content
+  meta_tags["govuk:stepnavs"] = stepnavs_content if stepnavs_content.present?
 %>
 <% meta_tags.each do |name, content| %>
   <meta name="<%= name %>" content="<%= content %>">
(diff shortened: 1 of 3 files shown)


app/views/govuk_component/analytics_meta_tags.raw.html.erb
```diff
@@ -101,6 +101,10 @@
   end
 
   meta_tags["govuk:static-analytics:strip-postcodes"] = "true" if should_strip_postcode_pii?(content_item_hash, local_assigns)
+
+  stepnavs = links_hash[:part_of_step_navs] || []
+  stepnavs_content = stepnavs.map { |stepnav| stepnav[:content_id] }.join(",")
```
selfthinker: I'm not quite sure how to pass several content IDs because our performance analyst, Tim, suggested to use a space instead of a comma to separate them so that CSVs are easier to handle. But we're using commas in other places, so I rather opted for being consistent.

sihugh: 👍

app/views/govuk_component/analytics_meta_tags.raw.html.erb
```diff
@@ -101,6 +101,10 @@
...
+  meta_tags["govuk:stepnavs"] = stepnavs_content if stepnavs_content
```
sihugh: I think this needs to finish `if stepnavs_content.present?`.  
At the moment, content with no `part_of_step_navs` links will be defaulted to an empty array which results in an empty string (`""`) from the `map`.  The empty string is truthy which is why the tests are failing.

PR #1367, opened 2018-04-05
Remove search component

andysellick:
This PR removes the search component from static, as it has been moved to the publishing components gem in [this PR](https://github.com/alphagov/govuk_publishing_components/pull/267). 

This component is used in frontend and finder-frontend, the following PRs will need to be merged before this PR can be merged.

- PR to update frontend: https://github.com/alphagov/frontend/pull/1478 (merged)
- PR to update finder-frontend: https://github.com/alphagov/finder-frontend/pull/462 (merged)


tijmenb:
Looks like it's also used in frontend:

https://github.com/alphagov/frontend/blob/b6ac36f6324bc24b29421d219b6ebc889a2018e3/app/views/homepage/index.html.erb#L21

config/routes.rb
```diff
@@ -2,6 +2,7 @@
   mount GovukPublishingComponents::Engine, at: "/component-guide" if defined?(GovukPublishingComponents)
 
   scope format: false do
+    get '/templates/govuk_component/search.raw.html.erb', to: proc { [200, {}, ['This component has moved to the gem']] }
```
andysellick: @tijmenb just a quick thought on this - isn't it better if this line isn't included and an error gets thrown? Wouldn't we be more likely to spot it before deploy?

tijmenb: Yeah, maybe just remove it. We better check the kibana logs for any calls to this template before deploying this PR.

andysellick:
@tijmenb the frontend PR to update the app to use the gem component has been merged.

andysellick:
Build is currently failing because static now includes the search component from the gem, but it doesn't include the stylesheets from the gem, which is causing the accessibility testing to throw an error (text isn't a suitable contrast, because the element in question isn't styled as it should be).

andysellick:
Stylesheet problem in the gem/static is now fixed, but need to reversion the gem and use it in this PR in order to make the tests pass.

PR #1374, opened 2018-04-10
Add feedback component to 400 error pages

andysellick:
Doesn't include it on any 500 error pages otherwise we might end up with a feedback loop.

Before:

screen shot 2018-04-10 at 11 54 50

After:

screen shot 2018-04-10 at 11 54 28

Trello card: https://trello.com/c/UtH4gcM3/74-feedback-form-on-404-pages


config/initializers/assets.rb
```diff
@@ -8,4 +8,5 @@
   header-footer-only*.css
   core-layout*.css
   static-print.css
+  error-page.css
```
tijmenb: Does `error-page.js` need to be here too?

tijmenb:
Once this is merged and deployed, govuk-puppet will copy over the error templates so the router can use it:

https://github.com/alphagov/govuk-puppet/blob/28ae5fbaf8deece6896e8b8ed2f4d693f4fb56c8/modules/router/manifests/errorpage.pp

(That is, if I understand that code correctly)

PR #1377, opened 2018-04-11
Add an optional environment variable for the Google Analytics ID

barrucadu:
If the env var `GA_UNIVERSAL_ID` is set, that is used for the Google Analytics ID.  If it's not set, the ID defaults to `UA-26179049-1` in production, and `UA-UNSET` in development and test.

---

This will let us use a different Google Analytics ID in staging and integration: https://trello.com/c/uLUgPcvI/50-stop-sending-staging-and-integration-data-to-same-ga-property-as-live-site-2

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -52,6 +52,9 @@ module Static
 
     config.paths["log"] = ENV["LOG_PATH"] if ENV["LOG_PATH"]
 
+    # Google Analytics ID
+    config.ga_universal_id = ENV.fetch("GA_UNIVERSAL_ID", "UA-UNSET")
+
     # Slimmer is used by the govuk_publishing_components gem, however it inserts itself
     # automatically as middleware in the host Rails application
     # Disable Slimmer middleware for Static and enable for use in gem only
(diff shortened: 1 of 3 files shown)


config/application.rb
```diff
@@ -52,6 +52,9 @@ class Application < Rails::Application
 
     config.paths["log"] = ENV["LOG_PATH"] if ENV["LOG_PATH"]
 
+    # Google Analytics ID
+    config.universalId = ENV["GA_UNIVERSAL_ID"]
```
kevindew: Nitpick - the convention for variables is to use snake case.

kevindew: Also can we set this as a default value here and then only override the one in "environments/production.rb" rather than needing entries in config/development and config/test ?

thomasleese: I was gonna say this could be `ENV.fetch` to ensure it's always set (if for some reason the environment is defined), but what @kevindew suggested works too.

kevindew: you want to be careful of the ENV.fetch as they can cause you real pain points when you're building assets 😢

thomasleese:
Once we've got the environment variables set in all environments, we can remove the special case for production here and just leave it as `GA-UNSET`.

PR #1386, opened 2018-04-19
Allow configuring date PII stripping in analytics via meta tag

barrucadu:
~~**This will *not* work until the changes to govuk_frontend_toolkit are made (PR: https://github.com/alphagov/govuk_frontend_toolkit/pull/459), and the gemfile here updated.  I'll rebase this branch then.**~~

Smart answers can include a date of birth in the URL, which is potentially bad.  So we remove them.  Same approach as #1183.

---

[Trello card](https://trello.com/c/zZzJHa8j/99-strip-dob-from-smart-answer-urls-before-sending-to-ga-2)


barrucadu:
I've asked if there's any preference about when we deploy this.  Will merge when I hear back.

PR #1389, opened 2018-04-24
Remove taxonomy sidebar

tijmenb:
This component was moved to the gem in https://github.com/alphagov/govuk_publishing_components/pull/253.

Is this right @steventux?

https://trello.com/c/pU7YINt1


steventux:
Yes that's right, are other frontend apps now using the component taxonomy sidebar (eg. alphagov/frontend) ?

tijmenb:
Yes! All of them are using the `contextual_sidebar` component, which uses the component from the gem.

PR #1390, opened 2018-04-24
Deprecate button & analytics meta tags

tijmenb:
Link them to the new components in the gem.

cc @andysellick


tijmenb:
@andysellick yes!

PR #1409, opened 2018-05-14
Add a breadcrumb schema to the breadcrump component

thomasleese:
This follows the `BreadcrumbList` specification defined at http://schema.org/BreadcrumbList.

[Trello Card](https://trello.com/c/t11eJLxC/138-implement-breadcrumb-schema-3)

### DIFF
--- a/app/views/govuk_component/breadcrumbs.raw.html.erb
+++ b/app/views/govuk_component/breadcrumbs.raw.html.erb
@@ -7,7 +7,7 @@
 %>
 
 <div class="govuk-breadcrumbs <%= collapse_class %>" data-module="track-click">
-  <ol>
+  <ol itemscope itemtype="http://schema.org/BreadcrumbList">
   <% breadcrumbs.each_with_index do |crumb, index| %>
     <%
       is_link = crumb[:url].present? || crumb[:is_current_page]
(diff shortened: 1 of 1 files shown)


app/views/govuk_component/breadcrumbs.raw.html.erb
```diff
@@ -33,11 +32,15 @@
...
-          }
-        ) %>
+          },
+          itemprop: "item",
+        ) do %>
+          <span itemprop="name"><%= crumb[:title] %></span>
+        <% end %>
       <% else %>
-        <%= crumb[:title] %>
+        <span itemprop="name"><%= crumb[:title] %></span>
```
thomasleese: I'm not a fan of the duplication of `span` here, but I'm not familiar enough with ERB to know the best solution so any suggests welcome!

thomasleese:
@tijmenb sounds good, I didn't realise we had picked `JSON/LD` as the way to show schema data on GOV.UK. I'm not super familiar with `JSON/LD` but can I basically put in that `<script>` tag as the first child of the the `<div>` of the breadcrumb?

thomasleese:
https://github.com/alphagov/govuk_publishing_components/pull/309

PR #1412, opened 2018-05-18
Remove feature to target surveys by breadcrumb

tijmenb:
Using this feature is currently broken because the CSS of the breadcrumbs changed from `govuk-breadcrumbs` to `gem-c-breadcrumbs` in https://github.com/alphagov/govuk_publishing_components/pull/309.

Instead of repairing it we remove it here, because the matching on breadcrumbs seems like a fragile method of implementing the matching since it relies on the internals of a component.

We're not currently using this feature, so we can remove it now without problems. If we want a replacement, perhaps looking at the `govuk:taxons` meta tag would be a more reliable way of determining the subject of the page.

https://trello.com/c/XLuzv0pB

### DIFF
--- a/app/assets/javascripts/surveys.js
+++ b/app/assets/javascripts/surveys.js
@@ -657,15 +657,6 @@
       }
     },
 
-    breadcrumbMatch: function (breadcrumbs) {
-      if (breadcrumbs === undefined) {
-        return false
-      } else {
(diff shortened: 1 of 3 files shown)


tijmenb:
Closing because it's easier to fix the existing implementation than to fix the tests. It's probably worth removing at some point though.

PR #1424, opened 2018-05-31
Revert "Add a meta noindex tag to any page which isn't first"

thomasleese:
This reverts commit 6dbf470a23e3bc6052f3cb1dc3704d030633055d.

Some pages use this component to move between different parts and we don't want to noindex those pages. In particular this is the case for mainstream and travel advice.


barrucadu:
The tags came from https://trello.com/c/PHjO8hP2/139-noindex-pagination-from-external-search-2, is that ticket still done with this revert?

thomasleese:
@barrucadu unfortunately not, we'll be bring that card back from done.

PR #1476, opened 2018-07-25
Supergroup survey

oscarwyatt:
Managed by Content Pages and Navigation

### DIFF
--- a/app/assets/javascripts/surveys.js
+++ b/app/assets/javascripts/surveys.js
@@ -326,7 +326,38 @@
           ]
         },
         allowedOnMobile: true
-      }
+      },
+      {
+        identifier: 'Business_iteration 1',
+        surveyType: 'url',
+        frequency: 6,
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/surveys.js
```diff
@@ -326,9 +326,37 @@
...
+          surveyCta: 'Answer 2 quick questions',
+          surveyCtaPostscript: 'This activity will open in a separate window'
+        },
+        activeWhen: {
+          section: ['business/all']
+        },
+        allowedOnMobile: true
+      },
+      {
+      url: 'https://gdsuserresearch.optimalworkshop.com/treejack/dpo18280-0',
```
vanitabarrett: To make it easier to read when skimming through this file, can we follow the pattern of other survey entries above, e.g: `identifier` first, then `surveyType` etc

vanitabarrett: Do we need to set `allowedOnMobile: true`?

PR #1482, opened 2018-08-02
Enable Brakeman

thomasleese:
This allows us to find any security vulnerabilities. I've also ignored a false positive about dynamic templates.

[Trello Card](https://trello.com/c/JyrOAef3/363-fix-sql-injection-vulnerabilities-in-govuk-applications)

### DIFF
--- a/config/brakeman.ignore
+++ b/config/brakeman.ignore
@@ -1,0 +1,26 @@
+{
+  "ignored_warnings": [
+    {
+      "warning_type": "Dynamic Render Path",
+      "warning_code": 15,
+      "fingerprint": "0a4ebe31686d3a336e8746a0562a643ef742503fdf160807cf8ee2109f539609",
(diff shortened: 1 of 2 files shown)


config/brakeman.ignore
```diff
@@ -0,0 +1,26 @@
...
+      "code": "render(action => params[:template], { :layout => \"govuk_template\" })",
+      "render_path": null,
+      "location": {
+        "type": "method",
+        "class": "RootController",
+        "method": "template"
+      },
+      "user_input": "params[:template]",
+      "confidence": "High",
+      "note": "We valid the parameter in validate_template_param to check it contains no slashes or underscores."
```
binaryberry: validATE?

thomasleese: Good catch! I've amended the commit.

PR #1489, opened 2018-08-07
Add publishing-app custom dimension

binaryberry:
This will help differentiate smart-answers from transactions.

Related to https://github.com/alphagov/govuk_publishing_components/pull/470.

https://trello.com/c/vhfMGngX/239-investigate-and-correct-smart-answer-google-analytics-tracking


cbaines:
What's the reasoning behind the ordering change? Numerical ordering is really useful when looking up dimensions.

app/assets/javascripts/analytics/custom-dimensions.js
```diff
@@ -47,6 +47,7 @@
       'navigation-legacy': {dimension: 30, defaultValue: 'none'},
       'navigation-page-type': {dimension: 32, defaultValue: 'none'},
       'political-status': {dimension: 7},
+      'publishing-app': {dimension: 89},
```
cbaines: I don't disagree with adding this, but I'm surprised that dimensions like `format` and `schema-name` don't separate Smart Answer pages from Transactions.

binaryberry:
@cbaines I hadn't noticed the custom dimension ordering! I thought it was a near-alphabetical ordering so wanted to improve it. Have removed that commit now.

app/assets/javascripts/analytics/custom-dimensions.js
binaryberry: @cbaines : smart-answers are an unusual case as their start page is being rendered as a transaction. The effect of this is that they are categorised as having a format and schema-name of transaction in GA. 

ie: https://www.gov.uk/api/content/marriage-abroad

PR #1583, opened 2018-12-04
Add banner support

steventux:
https://trello.com/c/ItwW03vg/48-build-a-generic-banner-based-on-an-older-banner-pattern

This PR resurrects previous banner code as the rules and design pattern works well for general site wide announcements which are not suitable for emergency banner content.

This can be merged as the banner is not enabled: https://github.com/alphagov/static/pull/1583/files#diff-e17896925164fa4471101aca24b30733R2

screenshot from 2018-12-05 14-24-03

This is based on code from https://github.com/alphagov/static/pull/769

Future iterations should address: 
- Converting global-bar styles to BEM naming convention
- Addressing inline minified JS


spec/javascripts/global-bar-class-toggle.spec.js
```diff
@@ -0,0 +1,119 @@
+describe("toggling a global bar HTML class based on cookie", function () {
+  var root = window;
+
+  function globalBarSource(fakeWindow) {
+    var window = fakeWindow || root;
+
+    /* --------------------------------------- */
```
andysellick: What are these dividing lines for? They're pretty ugly, and I don't think we have anything like this in our current codebase.

app/assets/stylesheets/helpers/_header.scss
```diff
@@ -341,3 +341,47 @@
...
   }
 }
+
+
+.show-global-bar #global-header-bar {
+  display: none;
+}
+
+#global-bar {
+  margin-bottom: 15px;
```
andysellick: We should probably be using some mixins for the spacing - I think you can get 15px with `govuk-spacing(3)` - https://design-system.service.gov.uk/styles/spacing/

app/assets/stylesheets/helpers/_header.scss
```diff
@@ -341,3 +341,47 @@
...
+
+  .global-bar-message {
+    margin-top: 0;
+    margin-bottom: 0;
+
+    @include media(tablet) {
+      max-width: 66.67%;
+    }
+
+    strong {
```
andysellick: This should be a class rather than styling on the element. Suggest `global-bar-message-title`?

app/assets/stylesheets/helpers/_header.scss
```diff
@@ -341,3 +341,42 @@
     margin-right: .5em;
   }
 }
+
+#global-bar {
```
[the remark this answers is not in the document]
steventux: _I think_ this would require reworking toggling in general as this element id is assumed by the toggle module (needed by the `Hide message` link in the banner) https://github.com/alphagov/static/blob/a815620cada7ea1c65428c1c3b3ac4dbe28977bf/app/assets/javascripts/modules/toggle.js#L57

spec/javascripts/global-bar-class-toggle.spec.js
steventux: Clarification added here https://github.com/alphagov/static/pull/1583/commits/39ce98444f066724a5a44d9179dd2a625fcf9dc0

app/assets/stylesheets/helpers/_header.scss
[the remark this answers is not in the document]
steventux: Done in https://github.com/alphagov/static/pull/1583/commits/3fdc23e8df288099f364658e2445577790b59d67

app/assets/stylesheets/helpers/_header.scss
steventux: Done in https://github.com/alphagov/static/pull/1583/commits/3fdc23e8df288099f364658e2445577790b59d67

app/assets/stylesheets/helpers/_header.scss
steventux: We're not going to address this immediately because this helper file contains a very mixed bag of styling via variables and pixel values. The reasoning is we're either going to spend the time maturing this banner code into something more permanent or we'll remove after understanding user engagement better.

steventux:
@andysellick are you OK with review replies/updates?

app/assets/stylesheets/helpers/_header.scss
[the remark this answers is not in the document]
andysellick: There's still one in there... `.show-global-bar #global-header-bar {`

andysellick:
@steventux all but one, I've added a comment.

app/assets/stylesheets/helpers/_header.scss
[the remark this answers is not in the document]
steventux: `#global-header-bar` refers to the blue bar which appears below the black GOV.UK header, it's a selector that was already part of the GOV.UK template. We're just using it to hide the blue bar and make the banner appear flush with the header.

andysellick:
@steventux all looks good now.

PR #1590, opened 2018-12-11
Fix global bar dupe attribute definition

steventux:
Bug found via e2e test output: https://ci.integration.publishing.service.gov.uk/job/publishing-e2e-tests/job/test-against/17090/artifact/tmp/errors-verbose.log

2nd commit is to demonstrate the bug is fixed and will be removed as we don't want to enable this banner.

cc @nickcolley

### DIFF
--- a/app/views/notifications/_global_bar.html.erb
+++ b/app/views/notifications/_global_bar.html.erb
@@ -12,7 +12,7 @@
     <!--<![endif]-->
   <% end %>
   <!--[if gt IE 7]><!-->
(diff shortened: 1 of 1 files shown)


app/views/notifications/_global_bar.html.erb
```diff
@@ -1,5 +1,5 @@
 <%
-  show_global_bar ||= false # Toggles the appearance of the global bar
+  show_global_bar ||= true # Toggles the appearance of the global bar
```
nickcolley: You sure you want this to be true?

PR #1637, opened 2019-02-18
Make barchart govspeak enhancement resilient to re-initialisation

kevindew:
Trello: https://trello.com/c/VDgVNurM/544-extract-govspeak-javascript-functionality-to-govukpublishingcomponents

This barchart initialisation is vulnerable to breaking a component if
the initialisation is called twice. As the code for this is moving to
govuk_publishing_components there is going to be a period of time where
this can quite easily occur while apps are being updated and deployed.

This updates this initialisation code to mark when a chart has been
initialised and be able to ignore being run on charts that have already
been initialised.

### DIFF
--- a/app/assets/javascripts/govuk-component/govspeak-convert-html-pub-charts.js
+++ b/app/assets/javascripts/govuk-component/govspeak-convert-html-pub-charts.js
@@ -1,8 +1,9 @@
 function govspeakBarcharts() {
-    $('.govuk-govspeak .js-barchart-table').each(function() {
+    $('.govuk-govspeak .js-barchart-table:not(.mc-chart):not(.js-barchart-table-init)').each(function() {
         $.magnaCharta($(this), {
             toggleText: "Change between chart and table"
         });
+      $(this).addClass('js-barchart-table-init')
     })
 }
 $(govspeakBarcharts);


app/assets/javascripts/govuk-component/govspeak-convert-html-pub-charts.js
```diff
@@ -1,8 +1,9 @@
 function govspeakBarcharts() {
-    $('.govuk-govspeak .js-barchart-table').each(function() {
+    $('.govuk-govspeak .js-barchart-table:not(.mc-chart):not(.js-barchart-table-init)').each(function() {
```
alex-ju: This selector seems rather flaky. Isn't there a way to check for a `magnaCharta` instance instead? Or update the script to ensure is doesn't load twice.

kevindew: It is grim I agree, it's quite hard to deal with as the library seems rather legacy.

An alternative selector could be `.govuk-govspeak table.js-barchart-table:not(.js-barchart-table-init)` - I don't think there's another way without changing the library code - are you happier with that? I know it's a choice between two non ideal scenarios.

I'm not keen on changing the library itself as that involves picking up a project that hasn't been touched for 4 years and will easily escalate as a piece of work.

It might be more relevant to have this chat on https://github.com/alphagov/govuk_publishing_components/pull/751/files#diff-bd92d59aa72dc924833d655049e37ee0R17 as the code in this repo is intended to be very short lived whereas the code in govuk_publishing_components will have a longer life

alex-ju: Good point about the not-so-long life of this repo. Happy to get it merged.

kevindew:
Thanks @alex-ju

PR #1647, opened 2019-03-08
Auto-generate non-digest versions of JS assets

karlbaker02:
This PR extends the `assets:precompile` Rake task to automatically generate non-digest versions of all JavaScript files, as an intermediary step to ensure that tests on our CI continue to pass until such time as all tests use fingerprinted versions of dependent JS files.

Influenced in large part by https://bibwild.wordpress.com/2014/10/02/non-digested-asset-names-in-rails-4-your-options (Rails 5 offers us no respite in this area either).

### DIFF
--- a/lib/tasks/asset_compile.rake
+++ b/lib/tasks/asset_compile.rake
@@ -1,0 +1,23 @@
+require 'json'
+
+Rake::Task['assets:precompile'].enhance do
+  Rake::Task['asset_precompile:create_non_digest_assets'].invoke
+end
+
+namespace :asset_precompile do
+  task :create_non_digest_assets do
+    relative_asset_path = 'public/static'
+    manifest_path = Dir.glob(File.join(Rails.root, relative_asset_path, '.sprockets-manifest-*.json')).first
+
+    manifest_data = JSON.parse(File.read(manifest_path))
+
+    manifest_data["assets"].each do |logical_path, digested_path|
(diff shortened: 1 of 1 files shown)


lib/tasks/asset_compile.rake
```diff
@@ -0,0 +1,23 @@
...
+    manifest_path = Dir.glob(File.join(Rails.root, relative_asset_path, '.sprockets-manifest-*.json')).first
+
+    manifest_data = JSON.load(File.new(manifest_path))
+
+    manifest_data["assets"].each do |logical_path, digested_path|
+      if logical_path.end_with?('.js')
+        full_digested_path = File.join(Rails.root, relative_asset_path, digested_path)
+        full_nondigested_path = File.join(Rails.root, relative_asset_path, logical_path)
+
+        FileUtils.copy_file full_digested_path, full_nondigested_path, true
```
benthorner: Could we do a link here, instead of copying it?

karlbaker02: Done.

karlbaker02:
To clarify, this PR is an intermediary step until such time as we have test wrappers pointing at fingerprinted versions of assets and all apps which consume static updated to use the latest version. Until then, we are continuing to support non-fingerprinted assets by creating symbolic links from them to the fingerprinted versions.

PR #1676, opened 2019-04-03
Adds new custom dimension to track relevant result shown in finders

DilwoarH:
The new custom dimension is Relevant result shown (83). This should be set to 'Yes' if a  search in the finder search box results in a 'relevant result'. (The null label is 'No'.) 

https://trello.com/c/lwQK2Pi7/336-set-up-a-b-test-for-answer-boxes

### DIFF
--- a/app/assets/javascripts/analytics/custom-dimensions.js
+++ b/app/assets/javascripts/analytics/custom-dimensions.js
@@ -58,7 +58,8 @@
       'taxon-ids': {dimension: 59, defaultValue: 'other'},
       'content-has-history': {dimension: 39, defaultValue: 'false'},
       'publishing-application': {dimension: 89},
-      'stepnavs': {dimension: 96}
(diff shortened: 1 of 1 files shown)


DilwoarH:
This is no longer needed.

DilwoarH:
> Does this file: https://github.com/alphagov/govuk-static-private/blob/1422fb460596d5821dee7963ec227cdad5c033ab/app/views/govuk_component/analytics_meta_tags.raw.html.erb need to be updated as well?
> 
> A bit like in this commit: [alphagov/govuk-static-private@1422fb4#diff-6d770471136872e42f49ea40fb39d63c](https://github.com/alphagov/govuk-static-private/commit/1422fb460596d5821dee7963ec227cdad5c033ab#diff-6d770471136872e42f49ea40fb39d63c) ?

Hey @leenagupte - not needed as we want this for only a specific view and not all of them.

PR #1689, opened 2019-04-15
Adds tracking attributes to footer links

bilbof:
Trello: https://trello.com/c/z9DP4EMK/402-add-tracking-to-govuk-footer-links


bilbof:
Thanks, Andy. I've added `data-module="track-click"`!

bilbof:
Many thanks, Andy! I'll followup with Haur once this is deployed.

PR #1694, opened 2019-04-18
Allow survey banner to be shown with cookie banner

vanitabarrett:
We were previously preventing the survey banner from being shown when any other banners were present, including the cookie banner. Now that the cookie banner is persistent, we want to allow both to show at the same time.

Screen Shot 2019-04-18 at 11 47 17


chao-xian:
https://trello.com/c/cRBewa76

chao-xian:
https://trello.com/c/cRBewa76

PR #1707, opened 2019-05-02
Add cookie banner shown dimension to pageview

vanitabarrett:
We originally added a tracking solution within govuk_publishing_components which fired an event if the cookie banner was not shown on the page. However, this could build up and hit our Google Analytics limit.

Instead, we are adding an additional dimension to our existing pageview event, so we don't need an additional event. This is dimension100 which has been defined as "cookie banner shown". It is set to false if seen_cookie_message is true and set to true if seen_cookie_message is false (i.e: when the banner is shown).

## Example
Screen Shot 2019-05-02 at 17 29 29


vanitabarrett:
[Improve cookie banner tracking](https://trello.com/c/tOKIPmZ3/99-improve-cookie-banner-tracking)

PR #1754, opened 2019-06-13
Set the TLS cookie rather than stubbing getCookie

vanitabarrett:
Stubbing `getCookie` will stub the return value for all calls to `getCookie`, which isn't what we want to do - we really only want to stub the value of TLS cookie.

**Note: these test failures are only present once the govuk_publishing_components has been versioned to include, so a passing build on CI now isn't representative of fixed tests until the Gemfile has also been updated. https://github.com/alphagov/govuk_publishing_components/pull/916**

### DIFF
--- a/spec/javascripts/surveys-spec.js
+++ b/spec/javascripts/surveys-spec.js
@@ -573,23 +573,27 @@ describe('Surveys', function () {
   })
 
   describe('currentTlsVersion', function() {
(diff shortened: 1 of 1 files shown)


vanitabarrett:
[Cookie mechanism fixes](https://trello.com/c/0PhHq7i0/187-cookie-mechanism-fixes)

PR #1755, opened 2019-06-13
Handle GA cookies when usage cookies are denied

vanitabarrett:
**The build is failing because it relies on some fixes from version 17 of govuk_publishing_components and some test fixes in static around TLS cookies, and this hasn't been updated yet.**

If a user doesn't give consent to usage cookies, we need to disable GA tracking. This involves:

- Setting a flag which turns off tracking
- Not automatically setting the value of the GA cookies. This allows us to set the value of the cookies to null on the `govuk_publishing_components` side and not have the cookies be automatically re-created each time we try to do this.

### DIFF
--- a/app/assets/javascripts/analytics/init.js.erb
+++ b/app/assets/javascripts/analytics/init.js.erb
@@ -1,24 +1,34 @@
 (function() {
   "use strict";
 
-  // Load Google Analytics libraries
-  GOVUK.StaticAnalytics.load();
+  var consentCookie = window.GOVUK.getConsentCookie();
 
-  // Use document.domain in dev, preview and staging so that tracking works
-  // Otherwise explicitly set the domain as www.gov.uk (and not gov.uk).
-  var cookieDomain = (document.domain == 'www.gov.uk') ? '.www.gov.uk' : document.domain;
+  // Disable analytics by default
+  // This will be reversed below, if the consent cookie says usage cookies are allowed
+  window['ga-disable-UA-26179049-1'] = true;
 
-  var universalId = '<%= Rails.application.config.ga_universal_id %>';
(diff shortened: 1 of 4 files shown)


vanitabarrett:
[Remove GA cookies if consent for 'usage' cookies is withdrawn](https://trello.com/c/wo6wNeeq/181-remove-ga-cookies-if-consent-for-usage-cookies-is-withdrawn)

app/assets/javascripts/analytics/init.js.erb
```diff
@@ -1,24 +1,34 @@
...
-    universalId: universalId,
-    cookieDomain: cookieDomain,
-    allowLinker: true,
-    // This is served by Fastly in production, and returns an empty 200 response
-    govukTrackerUrl: '<%= asset_path "/static/a" %>'
-  });
-
-  // Make interface public for virtual pageviews and events
-  GOVUK.analytics = analytics;
+   var consentCookie = window.GOVUK.getConsentCookie();
```
chao-xian:
```suggestion
  var consentCookie = window.GOVUK.getConsentCookie();
```

spec/javascripts/analytics/static-analytics-spec.js
```diff
@@ -1058,6 +1059,28 @@ describe("GOVUK.StaticAnalytics", function() {
     });
   });
 
+  describe("when the consent cookie has been set", function() {
+    beforeEach(function() {
+      window.ga.calls.reset();
+    });
+
+    it('does not set analytics cookies as normal when usage cookies are allowed', function() {
```
chao-xian:
```suggestion
    it('does not set analytics cookies as normal when usage cookies are not allowed', function() {
```

PR #1771, opened 2019-06-26
Remove new_cookie_banner flag

vanitabarrett:
## What
Removes `new_cookie_banner` flag on the cookie banner component.

## Why
Following [the change to remove old cookie banner code](https://github.com/alphagov/govuk_publishing_components/pull/959) this flag is no longer required to render the new cookie banner.

**Note: this needs to be deployed only after all frontend apps have been deployed. This is so the new CSS and JS is in place before the `.gem-c-cookie-banner--new` is removed from the HTML via this change.**


vanitabarrett:
[Remove outdated cookie banner code once we've switched to the new banner](https://trello.com/c/Upyyk2yu/215-remove-outdated-cookie-banner-code-once-weve-switched-to-the-new-banner)

PR #1793, opened 2019-07-18
Add a dummy GOVUK.analytics when no cookie consent

peteglondon:
Previously we were not setting GOVUK.analytics
when the user had denied tracking cookie consent.

This is causing other apps to break as they call methods on this
directly, it always existed before.

This commit sets it to a dummy object (kinda like null object pattern)
with empty functions.

I'm not at all familiar with the how the FE stuff fits together, if there's
a more elegant way to do this, let me know.

I've tried it on integration and it does fix the problem.

Trello: https://trello.com/c/t2LSpLSc/1139-cookie-policy-setting-breaks-js-functions-across-govuk-frontend

### DIFF
--- a/app/assets/javascripts/analytics/init.js.erb
+++ b/app/assets/javascripts/analytics/init.js.erb
@@ -3,6 +3,15 @@
 
   var consentCookie = window.GOVUK.getConsentCookie();
 
+  var dummyAnalytics = {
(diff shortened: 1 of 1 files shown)


andysellick:
@peteglondon this is how it's being fixed in finder-frontend for reference: https://github.com/alphagov/finder-frontend/pull/1262

PR #1795, opened 2019-07-22
Add analytics from govuk_frontend_toolkit

andysellick:
## What

Migrate analytics code from `govuk_frontend_toolkit` into static.

## Why

We are trying to retire `govuk_frontend_toolkit` and many other applications depend upon these analytics scripts. Moving these scripts into static is the first step in the eventual plan to move them into `govuk_publishing_components`.

## WIP

Todo:

- fix failing tests (failing tests are currently skipped). Currently this is only the error tracking tests.

Supersedes https://github.com/alphagov/static/pull/1777/

### DIFF
--- a/doc/analytics.md
+++ b/doc/analytics.md
@@ -1,0 +1,272 @@
+# Analytics
+
+The toolkit provides an abstraction around analytics to make tracking pageviews, events and dimensions across multiple analytics providers easier. Specifically it was created to ease the migration from Google’s Classic Analytics to Universal Analytics. It includes:
+
+* a Google Analytics universal tracker wrapper
+* code to asynchronously load universal analytics
+* a generic Analytics wrapper that allows multiple trackers to be configured
+* sensible defaults such as anonymising IPs
+* data coercion into the format required by Google Analytics (eg a custom dimension’s value must be a string)
+* stripping of PII from data sent to the tracker (strips email by default, can be configured to also strip dates and UK postcodes)
+
+## Create an analytics tracker
+
+The minimum you need to use the analytics function is:
+
+1. Include the following files from /javascripts/govuk/analytics in your project:
+  * google-analytics-universal-tracker.js
+  * analytics.js
+2. Copy the following `init` script into your own project and replace the dummy IDs with your own (they begin with `UA-`).
+  * This initialisation can occur immediately as this API has no dependencies.
+  * Load and create the analytics tracker at the earliest opportunity so that:
+    * the time until the first pageview is tracked is kept small and pageviews aren’t missed
+    * javascript that depends on `GOVUK.analytics` runs after the tracker has been created
+
+```js
+(function() {
+  "use strict";
+
+  // Load Google Analytics libraries
+  GOVUK.Analytics.load();
+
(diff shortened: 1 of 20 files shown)


andysellick:
We might need to press ahead with this (for reasons, happy to discuss). 

The only thing that isn't finished here yet is I still haven't figured out a way of testing the error-tracking. The test actually passes, but because we're generating a deliberate error, jasmine notices and counts it as a test failure (I've tried a lot of workarounds but if you think of anything let me know). I've currently marked the two tests in question as 'skip' (see the commit marked WIP).

@alex-ju @vanitabarrett @injms

alex-ju:
@andysellick I think we can create issues for that test and move on with it

andysellick:
@alex-ju awesome. I've just gone through this PR, I've structured the commits to make it easy to review. Hopefully shouldn't take long.

spec/javascripts/analytics_toolkit/govuk-tracker.spec.js
```diff
@@ -52,12 +52,23 @@ describe('GOVUK.GOVUKTracker', function () {
     })
 
     it('adds performance data', function () {
+      if (!window.performance) {
```
alex-ju: I'm not sure if this eases the case, but alternatively we could use a spy to return expected browser values if they're not attached to the window object.

`spyOn(window, 'performance').andReturn({ 'navigation': { type: '0', redirectCount: '0' }, 'timing': { 'domComplete: '123456' } })`

doc/analytics.md
```diff
@@ -13,7 +17,7 @@ The toolkit provides an abstraction around analytics to make tracking pageviews,
 
 The minimum you need to use the analytics function is:
 
-1. Include the following files from /javascripts/govuk/analytics in your project:
+1. Include the following files from /javascripts/govuk/analytics_toolkit in your project:
```
alex-ju: 1. Include the following files from `/javascripts/analytics_toolkit` in your project:

andysellick:
@alex-ju I've now tested this using a locally running copy of `collections` pointing to a locally running copy of `static`. There was an error, which I have fixed, which was related to the 'fix ecommerce tests' commit. GA events are now firing correctly. Good idea to test locally!

I agree that merging all the scripts into the same dir at some point is a good idea but actually for now it's been helpful to have them separate as it's easier to see where they're coming from in the browser. This has been helpful for reasons I can go into offline, if you're interested.

spec/javascripts/analytics_toolkit/govuk-tracker.spec.js
andysellick: I'm not sure I understand the difference between these two options so I'm happy to bow to your expertise on this. Is your change worth doing for a technical reason (does this create a window.performance object outside the scope of the tests that could cause problems?) or is it just neater?

alex-ju: Nope, I guess it's just the `jasmine` way of doing it.

alex-ju:
Tested this by comparing logs from Google Analytics debug. Also checked that beacon messages are being sent and interactions are triggered. 🤞

PR #1838, opened 2019-09-04
Update linked domains

andysellick:
- also change format of the function call so it's more readable and updateable


andysellick:
> Question of whether some of the domains are www or not. We can't be sure until it's live, so let's push this out and see.

Fair enough. I'm going to push it out, but I'll have a revert PR ready just in case.

PR #1863, opened 2019-09-17
Remove some URL filtering

bilbof:
The PII stripper code is bringing up too many false positives (e.g. dates used in search, IDs in step by steps determined incorrectly to be dates).

Examples:

```
https://www.gov.uk/settled-status-eu-citizens-families/what-youll-need-to-apply?step-by-step-nav=0c79b832-75de-4854-8154-d62774a8dfb8
```

The code thinks `4854` is a date.

```
https://www.gov.uk/government/news/uk-programme-assistance-to-2019-2020
```

The code thinks `2019-2020` is a date.

## What does this do?

This change rolls back a change in https://github.com/alphagov/static/pull/1850.

Now PII stripping behaviour is determined by page type. This was the case prior to #1850.

For example, search pages exclude postcodes, but not dates:
https://github.com/alphagov/govuk_publishing_components/blob/master/spec/components/meta_tags_spec.rb#L374

Smart answers exclude dates:
https://github.com/alphagov/govuk_publishing_components/blob/master/spec/components/meta_tags_spec.rb#L349

I have also moved all the PII logic and state into the pii component, which is the majority of the change.

Trello https://trello.com/c/t3mj71ey/118


app/assets/javascripts/analytics_toolkit/google-analytics-universal-tracker.js
```diff
@@ -3,11 +3,11 @@
 
   var $ = global.jQuery
   var GOVUK = global.GOVUK || {}
-  var pii = new GOVUK.pii()
-  pii.stripDatePII = true
-  pii.stripPostcodePII = true
+  var pii
 
   var GoogleAnalyticsUniversalTracker = function (trackingId, fieldsObject) {
+    pii = new GOVUK.pii()
```
andysellick: Is there a difference in initialising this variable here as opposed to where it was? I'm just curious, not suggesting it be changed.

bilbof: Yeah, when GoogleAnalyticsUniversalTracker is initialized, we want the latest version of the DOM. But, if pii is initialized in the global scope it has state determined from the original version of the DOM.

In practice it might not matter, since we don't tend to change meta tags on the fly, but it's helpful for tests.

PR #1869, opened 2019-09-19
Link some more domains

andysellick:
Adding some more domains to the cross linker.


andysellick:
Thanks Ben. Unfortunately there's not really a card for this, it's an ongoing thing. Maybe I could create a card, I'll have a think.

benthorner:
@andysellick sure - I don't think we need a card necessarily. Ideally, we should write something to help a future dev figure out where these came from i.e. why these domains in particular?

andysellick:
I've got some documentation for this whole process in progress, just not quite sure where to put it yet.

andysellick:
@benthorner https://github.com/alphagov/static/pull/1871

PR #1872, opened 2019-09-20
Add more domains to the cross domain linker

andysellick:
Adding more domains. See https://github.com/alphagov/static/blob/master/doc/analytics.md#adding-more-domains-to-govuks-cross-domain-linker


injms:
🔡 ftw!

PR #1877, opened 2019-09-24
Update footer link to Accessibility

chao-xian:
The Accessibility Statement needs to be available from all pages on
the site. Will also replace the old page with a redirect in other
work.


chao-xian:
https://trello.com/c/d1dGtXFd

PR #1887, opened 2019-10-01
Override govuk-template footer colours

chao-xian:
The focus state colours are not consistent across even the homepage,
which is a jarring experience. Explicitly setting the colours for
now until we can get the footer component in to Static.

## Before

www gov uk_reapply-driving-licence-medical-condition

## After

Screenshot at Oct 01 2-34-46 pm


chao-xian:
https://trello.com/c/EkqR7uj9

PR #1897, opened 2019-10-07
Remove displayFeaturesTask from GA code

andysellick:
- seems to be unnecessary for our purposes: `allowAdFeatures:false` is already being set
- according to https://developers.google.com/analytics/devguides/collection/analyticsjs/tasks displayFeaturesTask `Sends an additional hit if display features is enabled and a previous hit has not been sent within the timeout period set by the advertising features cookie (_gat).` and we're not using display features

Trello card: https://trello.com/c/41rsUlcT/157-implement-allowadfeaturesfalse-on-cross-domain-tracking-and-update-guidance

### DIFF
--- a/app/assets/javascripts/analytics_toolkit/google-analytics-universal-tracker.js
+++ b/app/assets/javascripts/analytics_toolkit/google-analytics-universal-tracker.js
@@ -18,11 +18,6 @@
       sendToGa('set', 'anonymizeIp', true)
     }
 
-    function disableAdTracking () {
-      // https://support.google.com/analytics/answer/2444872?hl=en
-      sendToGa('set', 'displayFeaturesTask', null)
-    }
-
     function disableAdFeatures () {
       // https://developers.google.com/analytics/devguides/collection/analyticsjs/field-reference#allowAdFeatures
       sendToGa('set', 'allowAdFeatures', false)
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/analytics_toolkit/google-analytics-universal-tracker.js
```diff
@@ -18,11 +18,6 @@
       sendToGa('set', 'anonymizeIp', true)
     }
 
-    function disableAdTracking () {
-      // https://support.google.com/analytics/answer/2444872?hl=en
-      sendToGa('set', 'displayFeaturesTask', null)
```
chao-xian: According to the [docs](https://developers.google.com/analytics/devguides/collection/analyticsjs/tasks#disabling_a_task) linked to, setting to `null` disables this feature? This feels more explicit, rather than assuming that the default is to not have it enabled?

chao-xian: Have realised that the `sendToGa(name + '.set', 'allowAdFeatures', false)` is all that's required to disable these features, and that the setting `displayFeaturesTask` to `null` is redundant. A rewrite of the commit message and PR description to include that explanation is all that's required. I'll approve in the meantime in expectation of that happening.

PR #1904, opened 2019-10-14
Ability to track click events with a custom event label

kr8n3r:
At the moment all instances of tracking click events in GA are sent with an eventLabel of 'Results' as shown below.
Screen Shot 2019-10-15 at 09 26 47
and this is hardcoded in [ecommerce.js](https://github.com/alphagov/static/blob/6a0d8a619e56063c81c193458bff92618f163306/app/assets/javascripts/analytics/ecommerce.js#L49)

This is sufficient when you're only tracking one set of link clicks with ecommerce.

Ability to track items on separate lists [was added ](https://github.com/alphagov/static/pull/1810) but the clicks always have the hardcoded `eventLabel` as shown above.

For upcoming [spelling suggestion work](https://github.com/alphagov/finder-frontend/pull/1643) we have a requirement to track those links with a different label.

This PR to provide the option to specify the event label sent to GA when clicking a link in a specific list.

This can be done by adding an additional  `data-track-click-label` attribute to the element where we initialise list tracking.

Example of how this would work: 
Screen Shot 2019-10-15 at 09 47 04

While we could reuse the `listTitle` string for the click event label, there's probably a reason why standard search results link tracking has them separate, as shown below, where the list title is 'Search`
Screen Shot 2019-10-15 at 09 50 29

### DIFF
--- a/app/assets/javascripts/analytics/ecommerce.js
+++ b/app/assets/javascripts/analytics/ecommerce.js
@@ -4,6 +4,7 @@
   window.GOVUK = window.GOVUK || {};
 
   var DEFAULT_LIST_TITLE = 'Site search results';
+  var DEFAULT_TRACK_CLICK_LABEL = 'Results';
 
   var Ecommerce = function (config) {
     this.init = function (element) {
@@ -14,6 +15,7 @@
       var startPosition = parseInt(element.data('ecommerce-start-index'), 10);
       var listTitle     = element.data('list-title') || DEFAULT_LIST_TITLE;
       var variant       = element.data('ecommerce-variant');
+      var trackClickLabel = element.data('track-click-label') || DEFAULT_TRACK_CLICK_LABEL;
 
       ecommerceRows.each(function(index, ecommerceRow) {
         var $ecommerceRow = $(ecommerceRow);
@@ -22,7 +24,7 @@
           path = $ecommerceRow.attr('data-ecommerce-path');
 
         addImpression(contentId, path, index + startPosition, searchQuery, listTitle, variant);
-        trackProductOnClick($ecommerceRow, contentId, path, index + startPosition, searchQuery, listTitle, variant);
+        trackProductOnClick($ecommerceRow, contentId, path, index + startPosition, searchQuery, listTitle, variant, trackClickLabel);
       });
     }
 
@@ -41,7 +43,7 @@
       ga('ec:addImpression', data);
     }
 
-    function trackProductOnClick (row, contentId, path, position, searchQuery, listTitle, variant) {
+    function trackProductOnClick (row, contentId, path, position, searchQuery, listTitle, variant, trackClickLabel) {
       row.click(function(event) {
         var data = {
           id: contentId || path,
@@ -55,7 +57,7 @@
 
         ga('ec:setAction', 'click', {list: listTitle});
         GOVUK.analytics.trackEvent('UX', 'click',
-          GOVUK.CustomDimensions.getAndExtendDefaultTrackingOptions({label: 'Results'})
+          GOVUK.CustomDimensions.getAndExtendDefaultTrackingOptions({label: trackClickLabel})
         );
       });
     }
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/analytics/ecommerce.js
```diff
@@ -4,15 +4,17 @@
...
     this.init = function (element) {
       // Limiting to 100 characters to avoid noise from extra longs search queries
       // and to stop the size of the payload going over 8k limit.
-      var searchQuery   = GOVUK.analytics.stripPII(element.attr('data-search-query')).substring(0, 100).toLowerCase();
-      var ecommerceRows = element.find('[data-ecommerce-row]');
-      var startPosition = parseInt(element.data('ecommerce-start-index'), 10);
-      var listTitle     = element.data('list-title') || DEFAULT_LIST_TITLE;
+      var searchQuery     = GOVUK.analytics.stripPII(element.attr('data-search-query')).substring(0, 100).toLowerCase();
+      var ecommerceRows   = element.find('[data-ecommerce-row]');
+      var startPosition   = parseInt(element.data('ecommerce-start-index'), 10);
```
andysellick: What's different about lines 13-15? I can't see a difference - if this is a weird editor thing it'd be good to get that fixed.

app/assets/javascripts/analytics/ecommerce.js
```diff
@@ -4,15 +4,17 @@
...
+      var listTitle       = element.data('list-title') || DEFAULT_LIST_TITLE;
+      var trackClickLabel = element.data('track-click-label') || DEFAULT_TRACK_CLICK_LABEL;
```
andysellick: I think I understand this change but it's a bit confusing. You're adding the ability to specify a different list from `Results` for this event to be included in, but doing so with a new variable called `trackClickLabel` when there's already a variable that does something different called `listTitle`? Does my confusion make sense? Could this be rewritten to be clearer? Or maybe rewrite the PR description?

app/assets/javascripts/analytics/ecommerce.js
kr8n3r: seeing that these lines were previously aligned at the length on the longest variable, i aligned them to the newest longest

app/assets/javascripts/analytics/ecommerce.js
kr8n3r: I'll make it clearer with screenshots, but list title is the name of the ecommerce list, whereas this allow for a flexible eventLabel on click, not always "Results"

kr8n3r: reverted

kr8n3r: updated PR description

spec/javascripts/analytics/ecommerce.spec.js
```diff
@@ -298,6 +298,89 @@ describe('Ecommerce reporter for results pages', function() {
...
+      dimension71: 'search query'
+    });
+    expect(ga).toHaveBeenCalledWith('ec:addProduct', {
+      id: 'AAAA-2222',
+      position: 1,
+      dimension71: 'blah'
+    });
+    expect(ga).toHaveBeenCalledWith('ec:setAction', 'click', {list: 'First list'})
+    expect(ga).toHaveBeenCalledWith('ec:setAction', 'click', {list: 'Second list'})
+    expect(ga).toHaveBeenCalledWith('send', {
```
andysellick: Does the test need to `expect` all of the GA calls? It might be shorter and clearer if it only expected the ones that are being explicitly tested here.

kr8n3r: let me just put one and see what happens :)

andysellick: I'm not 100% this morning but it looks like you've removed the relevant ones and left the non-relevant ones? Where's the tests for 'first list' and 'second list'?

PR #1907, opened 2019-10-16
Handle global_bar_seen cookie not being set

vanitabarrett:
To fix Smokey test failures like this: https://deploy.publishing.service.gov.uk/job/smokey/11502/consoleFull

Some of our Smokey tests don't seem to include the global bar, and so the JS included in the global bar view is not executed, including setting a cookie if it doesn't yet exist.

This commit replicates that logic in the global-bar.js file too, so if the cookie is not set, it will set a basic one. The next time the user navigates, that cookie will be updated with the correct version etc.

### DIFF
--- a/app/assets/javascripts/modules/global-bar.js
+++ b/app/assets/javascripts/modules/global-bar.js
@@ -9,9 +9,16 @@
 
   Modules.GlobalBar = function() {
     this.start = function($el) {
-      var GLOBAL_BAR_SEEN_COOKIE = "global_bar_seen",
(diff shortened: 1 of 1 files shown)


vanitabarrett:
@bilbof The global bar logic doesn't currently have any tests - it's on my list of things to come back and revisit when I have a bit more time 🙂

vanitabarrett:
@bilbof I don't have much time to do this at the moment, as I think it would involve writing a lot more tests for the banner as a whole. I'm going to revisit this at a later date and add more tests for this banner as it doesn't have much test coverage.

PR #1925, opened 2019-10-31
Add base target blank when rendered in iframes

alex-ju:
### What
Add a script to determine whether or not the current page sits within an <iframe>. If it does we change the base target to be blank so that all links will open in a new window (outside the iframe).

### Why
To allow people to check links (especially when they point to an non-secure (http) location) while previewing pages in Content Publisher.

[Trello card](https://trello.com/c/JMWWsYBT)

### DIFF
--- a/app/assets/javascripts/modules/base-target.js
+++ b/app/assets/javascripts/modules/base-target.js
@@ -1,0 +1,26 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  function BaseTarget () { }
+
+  BaseTarget.prototype.start = function () {
+    if (this.pageIsInIframe()) {
+      this.setBaseTarget()
+    }
+  }
+
+  BaseTarget.prototype.setBaseTarget = function () {
+    var base = document.createElement('base')
+    base.target = '_blank'
+    document.getElementsByTagName('head')[0].appendChild(base)
+  }
+
+  BaseTarget.prototype.pageIsInIframe = function () {
+    return window.parent && window.location !== window.parent.location
+  }
+
+  Modules.BaseTarget = BaseTarget
+})(window.GOVUK.Modules)
+
+new window.GOVUK.Modules.BaseTarget().start()
--- a/app/views/root/_javascript.html.erb
+++ b/app/views/root/_javascript.html.erb
@@ -1,3 +1,6 @@
 <%= javascript_include_tag 'libs/jquery/jquery-1.12.4.js', integrity: true, crossorigin: 'anonymous' %>
 <%= javascript_include_tag local_assigns[:js_file] || 'application', integrity: true, crossorigin: 'anonymous' %>
+<% if ENV["DRAFT_ENVIRONMENT"].present? %>
+  <%= javascript_include_tag 'modules/base-target', integrity: true, crossorigin: 'anonymous' %>
+<% end %>
 <%= javascript_include_tag 'surveys', integrity: true, crossorigin: 'anonymous' %>
--- a/spec/javascripts/modules/base-target.spec.js
+++ b/spec/javascripts/modules/base-target.spec.js
@@ -1,4 +1,5 @@
 /* eslint-env jasmine, jquery */
+// = require modules/base-target
 
 describe('A base target module', function () {
   'use strict'
--- a/spec/javascripts/modules/base-target.spec.js
+++ b/spec/javascripts/modules/base-target.spec.js
@@ -1,0 +1,41 @@
+/* eslint-env jasmine, jquery */
+
+describe('A base target module', function () {
+  'use strict'
+
+  describe('when rendered inside an iframe', function () {
+    var windowParent = window.parent
+    var mockWindowParent = {} // window.parent would be different than window when used inside an iframe
+
+    beforeEach(function () {
+      window.parent = mockWindowParent
+      var baseTarget = new window.GOVUK.Modules.BaseTarget()
+      baseTarget.start()
+    })
+
+    afterEach(function () {
+      window.parent = windowParent
+      var head = document.getElementsByTagName('head')[0]
+      var base = document.getElementsByTagName('base')[0]
+      if (head && base) head.removeChild(base)
+    })
+
+    it('should append a <base target="_blank"> tag to <head>', function () {
+      var base = document.getElementsByTagName('base')[0]
+      expect(base.getAttribute('target')).toEqual('_blank')
+    })
+  })
+
+  describe('when not rendered inside an iframe', function () {
+    beforeEach(function () {
+      var baseTarget = new window.GOVUK.Modules.BaseTarget()
+      baseTarget.start()
+    })
+
+    it('should not append a <base target="_blank"> tag to <head>', function () {
+      var head = document.getElementsByTagName('head')[0]
+      var base = head.getElementsByTagName('base')[0]
+      expect(base).toBeFalsy()
+    })
+  })
+})
--- a/app/assets/javascripts/start-modules.js
+++ b/app/assets/javascripts/start-modules.js
@@ -1,5 +1,4 @@
 // = require govuk/modules
-// = require modules/base-target
 // = require modules/global-bar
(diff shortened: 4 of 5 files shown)


app/assets/javascripts/modules/base-target.js
```diff
@@ -0,0 +1,26 @@
...
+
+  BaseTarget.prototype.start = function () {
+    if (this.pageIsInIframe()) {
+      this.setBaseTarget()
+    }
+  }
+
+  BaseTarget.prototype.setBaseTarget = function () {
+    var base = document.createElement('base')
+    base.target = '_blank'
```
ChrisBAshton: I suggest adding `rel="noopener noreferrer"` to the links for [security reasons](https://medium.com/sedeo/how-to-fix-target-blank-a-security-and-performance-issue-in-web-pages-2118eba1ce2f).

app/assets/javascripts/modules/base-target.js
```diff
@@ -0,0 +1,26 @@
...
+  function BaseTarget () { }
+
+  BaseTarget.prototype.start = function () {
+    if (this.pageIsInIframe()) {
+      this.setBaseTarget()
+    }
+  }
+
+  BaseTarget.prototype.setBaseTarget = function () {
+    var base = document.createElement('base')
```
ChrisBAshton: This will fix relative URLs but not absolute URLs, so if a content editor links to `http://www.example.com`, they will still get a preview issue. Would it be better to iterate over all `a` elements and set the `target` on them instead?

spec/javascripts/modules/base-target.spec.js
```diff
@@ -0,0 +1,41 @@
...
+    afterEach(function () {
+      window.parent = windowParent
+      var head = document.getElementsByTagName('head')[0]
+      var base = document.getElementsByTagName('base')[0]
+      if (head && base) head.removeChild(base)
+    })
+
+    it('should append a <base target="_blank"> tag to <head>', function () {
+      var base = document.getElementsByTagName('base')[0]
+      expect(base).toBeTruthy()
```
ChrisBAshton: May be worth asserting `base.target` === `_blank` too.

app/assets/javascripts/modules/base-target.js
alex-ju: I agree with the suggestion, but I'm not sure the `rel` attribute works on a base element and I would avoid querying all the links within the page because of the computational complexity.

app/assets/javascripts/modules/base-target.js
alex-ju: Oh, darn.. I was trying to avoid doing that, but doesn't seem to cover all the links.

alex-ju:
> Had a quick look, code looks great! Was it copied from somewhere? Asking as it seems a mature solution already, and if so the commit could refer to the repository from which it was originally copied.

No, it happens that sometimes I end up writing mature solutions, although this time doesn't seem to cover all of our needs — I may need to iterate over all links. If you're referring to the iframe detection I first wrote it for the [cookie banner component](https://github.com/alphagov/govuk_publishing_components/pull/995/files#diff-31a17edfef3f9fe90f5c6e9068c8b5aaR90), but I don't think it's worth abstracting it yet - especially while this code lives in `static`.

> Have left a couple of comments, though I don't have the context of any of the kickoff conversations so take my comments with a pinch of salt

The was no kickoff for this story.

app/assets/javascripts/modules/base-target.js
alex-ju: I've just tested again now and it's working consistently regardless of links having absolute or relative paths — essentially it works for every link within the page.

ChrisBAshton: Ahh that's good. Thanks 👍 might be worth adding some relative, absolute and external links to the tests and validate that they all open in new tabs?

app/assets/javascripts/modules/base-target.js
ChrisBAshton: Agreed 👍 thanks

app/assets/javascripts/modules/base-target.js
alex-ju: Sounds like a good idea, but how to achieve that with Jasmine - as we don't have any real browser to test with? I know how to do it in Jest, but otherwise open to proposals.

ChrisBAshton: Easier said than done! I tried two approaches:

1) [spying on `window.open`](https://stackoverflow.com/questions/31286162/how-to-test-window-open-using-jasmine) - no longer seems to be supported by Jasmine (`toHaveBeenCalled is undefined`)
2) [monitoring the browser history length](https://www.sitepoint.com/community/t/find-out-when-page-is-opened-in-new-window-or-new-tab/8910/6) - if it's more than 1 then it opened in the current window - but this interferes with the running of the Jasmine test as we've navigated away from the Jasmine spec runner 🤦‍♂ 

Can't think of another way of testing it at the moment, other than manually, which I suppose will have to do!

alex-ju: Tried spying on `window.open` too, but no luck. This adds to our set of cases where the limitations of Jasmine are blocking us from doing more robust testing. If it was part of an application we could rely on feature/integration tests, but otherwise I don't see an alternative either.

alex-ju:
Thank you for the review @ChrisBAshton. Based on @kevindew's comment I extracted this module and only load it if the draft environment is setup (meaning it's not loaded for everyone but only for the draft stack which is used for previews). Do you mind reviewing this again? I left it a separate commit to ease the 2nd review.

app/assets/javascripts/modules/base-target.js
```diff
@@ -0,0 +1,26 @@
...
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  function BaseTarget () { }
+
+  BaseTarget.prototype.start = function () {
+    if (this.pageIsInIframe()) {
+      this.setBaseTarget()
+    }
+  }
```
kevindew: Is it conventional for us to have a `start()` method? I wonder, see we immediately invoke it, whether this should just be done in the constructor?

app/assets/javascripts/modules/base-target.js
kevindew: Do we know what happens if this is run twice? Is two base elements ok or should it look for one and modify it if found create if not?

app/views/root/_javascript.html.erb
```diff
@@ -1,3 +1,6 @@
 <%= javascript_include_tag 'libs/jquery/jquery-1.12.4.js', integrity: true, crossorigin: 'anonymous' %>
 <%= javascript_include_tag local_assigns[:js_file] || 'application', integrity: true, crossorigin: 'anonymous' %>
+<% if ENV["DRAFT_ENVIRONMENT"].present? %>
+  <%= javascript_include_tag 'modules/base-target', integrity: true, crossorigin: 'anonymous' %>
+<% end %>
```
kevindew: I'd have thought we'd want this to be a general draft script that includes this module rather than including this module specifically.

app/assets/javascripts/modules/base-target.js
alex-ju: I tried to align this with the current module pattern that we have in different scripts across `static` (regardless of its complexity - as this could potentially be extended)

app/assets/javascripts/modules/base-target.js
alex-ju: It will indeed generate two base elements - I don't see why would that happen though and by having two of the will not affect the expected behaviour. If you experienced multiple script loading in GOV.UK in the past do let me know and I'll update this script to check for an existing element and replace it if so - otherwise I would investigate the cause of having it run multiple times instead of being too protective within the code.

app/views/root/_javascript.html.erb
alex-ju: I would not generalise this until we have at least two modules running only in the draft environment.

alex-ju:
I'm going to merge this. @kevindew if any of your comments needed to be addressed I'm happy to iterate this over afterwards.

PR #1934, opened 2019-11-11
Add register to vote banner

vanitabarrett:
Trello: https://trello.com/c/Lu2x8C4G/236-add-register-to-vote-banner

## What

- Updates global banner content to point to Register to Vote page
- Adds logic to support an automatic turn off date and time
- Sets automatic turn off datetime to 11:59pm on 26th November 2019

Screen Shot 2019-11-11 at 11 37 39

### DIFF
--- a/app/views/notifications/_global_bar.html.erb
+++ b/app/views/notifications/_global_bar.html.erb
@@ -1,16 +1,20 @@
 <%
-  show_global_bar ||= false # Toggles the appearance of the global bar
-  title = "A Brexit deal has been agreed but needs to be ratified"
-  link_href = "/brexit"
-  link_text = "Find out what this means for you"
+  require 'date'
+
+  show_global_bar ||= true # Toggles the appearance of the global bar
+  title = "Register to vote"
+  link_href = "/register-to-vote"
+  link_text = "Register by 26 November to vote in the General Election on 12 December."
+  close_date = DateTime.new(2019,11,26,23,59,00) # DateTime after which the bar will automatically disappear, or false to unset
 -%>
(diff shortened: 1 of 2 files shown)


app/views/notifications/_global_bar.html.erb
```diff
@@ -1,10 +1,14 @@
...
-  title = "A Brexit deal has been agreed but needs to be ratified"
-  link_href = "/brexit"
-  link_text = "Find out what this means for you"
+  require 'date'
+
+  show_global_bar ||= true # Toggles the appearance of the global bar
+  title = "Register to vote"
+  link_href = "/register-to-vote"
+  link_text = "Register by 26 November to vote in the General Election on 12 December."
+  close_date = DateTime.new(2019,11,26,23,59,00) # DateTime after which the bar will automatically disappear, or false to unset
```
benthorner: Minor: could we just remove the banner manually on the 26th? I'm wondering if it's a strict requirement for it to disappear at midnight, and if so, we might need to scrutinise this code a bit more carefully, because timezones...

benthorner: On second glance it basically looks fine. So yeah, just a minor question if we really need this date conditional.

vanitabarrett: @benthorner I think it's a strict requirement as the deadline for registering to vote is 11:59pm on 26th November. I'm hoping this might actually be a useful feature for the banner to have and which we might want to re-use for other scenarios in the future, so hopefully not a case of writing custom code for this one scenario.

PR #1936, opened 2019-11-11
Switch banner link to title

vanitabarrett:
We want the register to vote link to be on the banner title rather than the descriptive text.

Screen Shot 2019-11-11 at 14 24 16

### DIFF
--- a/app/views/notifications/_global_bar.html.erb
+++ b/app/views/notifications/_global_bar.html.erb
@@ -3,8 +3,9 @@
 
   show_global_bar ||= true # Toggles the appearance of the global bar
   title = "Register to vote"
-  link_href = "/register-to-vote"
+  title_href = "/register-to-vote"
   link_text = "Register by 26 November to vote in the General Election on 12 December."
+  link_href = false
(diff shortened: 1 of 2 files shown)


app/views/notifications/_global_bar.html.erb
```diff
@@ -25,16 +26,24 @@
...
-          rel: "external noreferrer",
-          class: "govuk-link js-call-to-action"
-        ) %>
+      <% if link_text %>
+        <% if link_href %>
+          <%= link_to(
+            link_text,
+            link_href,
+            rel: "external noreferrer",
+            class: "govuk-link js-call-to-action"
```
benthorner: Does this need to go on the new title link?

vanitabarrett: Good spot - done 👍

benthorner: Thanks - do we also need https://github.com/alphagov/static/pull/1936/files#diff-e17896925164fa4471101aca24b30733R41?

vanitabarrett: Hm, not sure that's accurate any more as the link isn't to an external site. Tempted to leave as is and remove in a separate PR so I can double check it's ok to remove with a performance analyst

PR #1942, opened 2019-11-20
Add more domains to the cross linker

andysellick:
Adding some more service domains to the cross domain linking configuration, now that they've turned on cross domain tracking.

See https://github.com/alphagov/static/blob/master/doc/analytics.md#adding-more-domains-to-govuks-cross-domain-linker


andysellick:
Thanks both 😄

PR #1965, opened 2019-12-10
Refactor global banner

vanitabarrett:
## What
Refactors global banner component:

- Replaces minified JS in the view file with a separate JS file
- Rewrites global bar initialize tests (which relied on manually updated minified JS) to test the new JS file mentioned above
- Add missing tests for global bar interaction logic, e.g: dismiss button, click events, tracking, always on behaviour
- Updates documentation to reflect these changes
- Renames surveys-spec.js to surveys.spec.js to be consistent with other spec files

## Why

- The global bar was relying on minified JS in the view file which is very difficult to 1) read and work with and 2) test. 
- The tests were relying on having the minified JS copied and pasted into the test file, so devs were making changes to the global bar and assuming nothing was broken because the tests were still passing, whereas actually the tests were still running against old code. 
- There were no tests for the global bar interaction logic

## Note
I had an issue with the tests passing locally, but failing in CI. I remember having this issue with other cookie tests, where `JSON.parse` was returning escaped strings in CI (but not locally). As a workaround, I've had to add a `parseCookie` function for the tests and in each JS file (which isn't ideal) which conditionally parses twice if necessary. If anyone has any ideas or has solved this problem before, please let me know!


app/assets/javascripts/global-bar-init.js
```diff
@@ -0,0 +1,65 @@
+'use strict'
```
ChrisBAshton: Unless I'm missing something, this code looks like it's been manually de-obfuscated from the minimised JS in `_global_bar.html`. I doubt that's the case, has this been copied from somewhere? If so, should it also be deleted from that place in the same commit as this one?

app/assets/javascripts/modules/global-bar.js
```diff
@@ -73,6 +73,17 @@
...
+      function parseCookie(cookie) {
+        var parsedCookie = JSON.parse(cookie)
+
+        // Tests seem to run differently on CI, and require an extra parse
+        if (typeof parsedCookie !== "object") {
+          parsedCookie = JSON.parse(parsedCookie)
+        }
+
+        return parsedCookie
+      }
```
ChrisBAshton: Is there a way this function can be included from the helper file rather than duplicated here?

spec/javascripts/modules/global-bar.spec.js
```diff
@@ -0,0 +1,157 @@
+describe('Global bar module', function () {
+  'use strict'
+
+  var globalBar
+  var element
+
+  beforeEach(function () {
+    globalBar = new GOVUK.Modules.GlobalBar()
+
+    document.cookie = 'global_bar_seen=; Path=/; Expires=Thu, 01 Jan 1970 00:00:01 GMT;';
```
ChrisBAshton: We seem to have some stray semi-colons in this file, like this one ^

spec/javascripts/modules/global-bar.spec.js
```diff
@@ -0,0 +1,157 @@
...
+
+      document.cookie = 'global_bar_seen=; Path=/; Expires=Thu, 01 Jan 1970 00:00:01 GMT;';
+    })
+
+    it('sets basic global_bar_seen cookie if not already set', function () {
+      expect(GOVUK.getCookie("global_bar_seen")).toBeNull()
+
+      globalBar.start(element)
+
+      expect(parseCookie(GOVUK.getCookie("global_bar_seen")).count).toBe(0)
```
ChrisBAshton: Shouldn't this be 1?

spec/javascripts/modules/global-bar.spec.js
```diff
@@ -0,0 +1,157 @@
...
+      expect(GOVUK.analytics.trackEvent).toHaveBeenCalledWith('Global bar', 'Viewed', {nonInteraction: 1});
+    })
+  })
+
+  describe('always on', function() {
+    beforeEach(function() {
+      document.cookie = 'global_bar_seen=; Path=/; Expires=Thu, 01 Jan 1970 00:00:01 GMT;';
+      spyOn(GOVUK.analytics, 'trackEvent');
+
+      document.cookie = 'global_bar_seen=; Path=/; Expires=Thu, 01 Jan 1970 00:00:01 GMT;';
```
ChrisBAshton: It seems odd we're setting this cookie to the same value twice.

app/assets/javascripts/global-bar-init.js
```diff
@@ -0,0 +1,65 @@
...
+  blacklistedUrl: function() {
+    return /^\/register-to-vote|^\/done|^\/brexit|^\/get-ready-brexit-check/.test(window.location.pathname)
+  },
+
+  setBannerCookie: function() {
+    var expiryDate = new Date(Date.now()+72576e5).toUTCString()
+    var value = JSON.stringify({count: 0, version: globalBarInit.getBannerVersion()})
+    var cookieString = "global_bar_seen=" + value + "; expires=" + expiryDate + ";"
+
+    document.cookie=cookieString
```
injms: Do you think it's worth using the `window.GOVUK.setCookie` helper to set the cookies here?

vanitabarrett: I did this to start with but was getting errors about `GOVUK.getCookie`/`GOVUK.setCookie` not being available. I think because this JS is specifically being pulled in earlier than other JS files, we perhaps can't rely on the cookie helper functions being available at that time?

injms: Argh, that's frustrating!

vanitabarrett:
Note to new reviewers: refactoring this further so hold off reviewing for now!

app/assets/javascripts/global-bar-init.js
vanitabarrett: (sorry if I've read your comment wrong) - this code *is* the JS that was previously in the `global_bar.html` file and it has manually been unminified - there wasn't an existing unminified copy of the code that was up to date. It's still required, but I've moved it this `global-bar-init.js` file so it's more readable and testable.

app/assets/javascripts/global-bar-init.js
vanitabarrett: @injms Having had a quick look, I don't think it actually needs to be pulled in earlier, so it's possible we could move it into `global-bar.js` and use the helper functions. 

I had a lot of trouble with tests when I did that though, so I'm thinking we can keep this as a first pass and iterate in future. It would be good to get this code merged first, so any urgent changes needed post-election can be done with more confidence.

spec/javascripts/modules/global-bar.spec.js
vanitabarrett: It's unlikely this would happen as the initialise code should handle this case, but if the global-bar logic kicks in and the cookie is null or not what is expected, it sets a value of `{count: 0, version: 0}`

https://github.com/alphagov/static/pull/1965/files#diff-146dfe5256358c4209bb150ac180fabeR17

app/assets/javascripts/global-bar-init.js
vanitabarrett: (copied from a reply to Ian's comment below, but provides some context to your question too)

Having had a quick look, I don't think it actually needs to be pulled in earlier, so it's possible we could move this code into `global-bar.js`.

I had a lot of trouble with tests when I did that though, so I'm thinking we can keep this as a first pass and iterate in future. It would be good to get this code merged first, so any urgent changes needed post-election can be done with more confidence.

app/assets/javascripts/global-bar-init.js
injms: Yep, makes complete sense - thanks for explaining 👍

app/assets/javascripts/modules/global-bar.js
vanitabarrett: Good shout. I wasn't sure on the best way of doing this as it's used by spec files too. 

I've gone with: moved parseCookie function into `libs/GlobalBarHelper.js` which is required by the global bar JS files and spec helper, but I'm not sure if it's good to be requiring a non-spec file in a spec. Any ideas?

ChrisBAshton: That sounds good to me 👍

app/assets/javascripts/global-bar-init.js
ChrisBAshton: You understood perfectly - I assumed it must have existed elsewhere in unminified form already. Well done for deobfuscating it!

app/assets/javascripts/global-bar-init.js
```diff
@@ -0,0 +1,56 @@
...
+  getBannerVersion: function() {
+    return BANNER_VERSION
+  },
+
+  blacklistedUrl: function() {
+    return /^\/register-to-vote|^\/done|^\/brexit|^\/get-ready-brexit-check/.test(window.location.pathname)
+  },
+
+  setBannerCookie: function() {
+    var expiryDate = new Date(Date.now()+72576e5).toUTCString()
```
ChrisBAshton: Can you set an explanatory variable instead of a magic value here?

```js
var whateverThisIs = 72576e5
```

app/assets/javascripts/global-bar-init.js
```diff
@@ -0,0 +1,56 @@
...
+// This will reset the view count so all users will see the banner, even if previously seen
+var BANNER_VERSION = 3;
+
+var globalBarInit = {
+  getBannerVersion: function() {
+    return BANNER_VERSION
+  },
+
+  blacklistedUrl: function() {
+    return /^\/register-to-vote|^\/done|^\/brexit|^\/get-ready-brexit-check/.test(window.location.pathname)
```
ChrisBAshton: This would be easier to read if it was split out into an array of paths, can probably then do something like `.test(new RegExp(paths.join("|")))`

app/assets/javascripts/global-bar-init.js
```diff
@@ -0,0 +1,56 @@
...
+        globalBarInit.setBannerCookie()
+        globalBarInit.makeBannerVisible()
+      } else {
+        var currentCookieVersion = parseCookie(currentCookieValue[1]).version
+
+        if (currentCookieVersion !== globalBarInit.getBannerVersion()) {
+          globalBarInit.setBannerCookie()
+        }
+
+        var newCookie = document.cookie.match("(?:^|[ ;])(?:global_bar_seen=)(.+?)(?:(?=;|$))")
```
ChrisBAshton: This repeats line 32 and is quite fiddly regex, can it be moved to a function?

```js
if (parseCookie(getLatestCookie()).count < 3) { //etc
```

vanitabarrett:
@ChrisBAshton I think I've addressed your comments now - would you mind re-reviewing?

app/assets/javascripts/global-bar-init.js
```diff
@@ -0,0 +1,62 @@
...
+
+  makeBannerVisible: function() {
+    document.documentElement.className = document.documentElement.className.concat(" show-global-bar")
+  },
+
+  init: function() {
+    var currentCookieValue = globalBarInit.getLatestCookie()
+
+    if (!globalBarInit.blacklistedUrl()) {
+      if (currentCookieValue === null) {
```
ChrisBAshton: I would just call the function inline here and delete line 39:

```suggestion
      if (globalBarInit.getLatestCookie() === null) {
```

And I would do so in the conditionals on line 48 and 54. It's slightly less efficient than caching the value in a variable, but easier to read & less can go wrong.

app/assets/javascripts/global-bar-init.js
```diff
@@ -0,0 +1,62 @@
...
+        globalBarInit.setBannerCookie()
+        globalBarInit.makeBannerVisible()
+      } else {
+        var currentCookieVersion = parseCookie(currentCookieValue[1]).version
+
+        if (currentCookieVersion !== globalBarInit.getBannerVersion()) {
+          globalBarInit.setBannerCookie()
+        }
+
+        var newCookieCount = parseCookie(globalBarInit.getLatestCookie()[1]).count
```
ChrisBAshton: I find it strange we have to hook into the `[1]` here, I'd expect `getLatestCookie` to return the object we need rather than the array output of the regex matcher. Could you modify the getLatestCookie function to return only the part we need?

vanitabarrett:
@ChrisBAshton Thanks, all done!

PR #1969, opened 2019-12-12
WIP Replace toolkit variables in the footer partial

kr8n3r:
This is to remove frontend_toolkit reliance from the footer Sass partial.

`$govuk-brand-colour` will make the footer top border light blue - just in time to match it with the updated feedback component background colour.


kr8n3r:
if we want to ship incremental removal we'd have to import of all publishing components Sass until https://github.com/alphagov/govuk_publishing_components/pull/1159, right @alex-ju ?

kr8n3r:
closing in favour or #1970 which will be merging into a remove_toolkit branch

PR #1970, opened 2019-12-13
Update footer partial to not use frontend_toolkit

kr8n3r:
Part of https://trello.com/c/fxfRx7dQ/95-remove-govukfrontendtoolkit-from-static
This PR
- imports govuk_publishing_component Sass files into header_footer
- replaces toolkit variables and mixins with publishing components

### DIFF
--- a/app/assets/stylesheets/header-footer-only.scss
+++ b/app/assets/stylesheets/header-footer-only.scss
@@ -1,3 +1,8 @@
+$govuk-compatibility-govuktemplate: true;
+@import "govuk/settings/all";
+@import "govuk/tools/all";
+@import "govuk/helpers/all";
+
 /* govuk_frontend_toolkit includes */
 @import "colours";
 @import "conditionals";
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/header-footer-only.scss
```diff
@@ -1,3 +1,6 @@
+$govuk-compatibility-govuktemplate: true;
+@import "govuk_publishing_components/all_components";
```
nickcolley: What's the performance impact of this?

alex-ju: Missed that line. It's quite bad I would imagine. @kr8n3r can we only fish out the bits we need? I imagine that would be setting and utilities.

kr8n3r: this is not going into master

publishing components gem doesn't expose anything other than all ?

alex-ju: Yeah, aware, but still I think we can do better. `govuk_publishing_components` exposes all `govuk-frontend`, so we can do, for example, `@import  "govuk/settings/all"` in any application that includes the gem.

kr8n3r: ok, wasn't sure. will change the imports

PR #1993, opened 2020-01-07
Remove crossorigin and SRI from our static assets (CSS/JS)

Nooshu:
Change to remove SRI on the JavaScript / CSS and remove the `crossorigin` attribute.

This change is part of [RFC-115](https://github.com/alphagov/govuk-rfcs/pull/115).

Changes have been tested on integration, [seen here](https://github.com/alphagov/static/pull/1993#issuecomment-580287175)

### DIFF
--- a/app/views/layouts/govuk_template.html.erb
+++ b/app/views/layouts/govuk_template.html.erb
@@ -6,12 +6,12 @@
     <meta charset="utf-8" />
     <title><%= content_for?(:page_title) ? yield(:page_title) : "GOV.UK - The best place to find government services and information" %></title>
 
-    <!--[if gt IE 8]><!--><%= stylesheet_link_tag "govuk-template.css", integrity: true, crossorigin: "anonymous" %><!--<![endif]-->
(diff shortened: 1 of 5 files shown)


Nooshu:
I've verified that we are seeing what is expected. Using Chrome DevTools, under the Network tab there is a column called 'Connection ID'. This identifies which TCP connection the asset was loaded on. I exported a HAR file from Chrome and examined it:

```
124231
* HTML

124279
* CSS

124280
* Image
* CSS

124281
* Script
* Image
* CSS

124282
* Script
* Image

124283
* Script
* CSS

124284
* Image
* favicon
* CSS

124288 (anonymous connection)
* Font

124344 (anonymous connection)
* Font
``` 
Still on HTTP/1 so we are still seeing 9 TCP connections, but that is better than the 13/14 we were seeing before.

Proof that it is working is we have CSS and Scripts both coming down the same connection ID as the images. Whereas before in `anonymous mode` they wouldn't be able to be downloaded on the same connection as the images (since the images used a credentialed TCP connection).

Both the CSS and JS are downloading and executing as expected as seen below:

Screenshot 2020-01-30 at 14 44 48

PR #2008, opened 2020-01-23
Remove toolkit dependency - grid layout

andysellick:
## What

Removes govuk_frontend_toolkit imports for the grid layout.

## Why

Part of migrating away from toolkit entirely. Trello card: https://trello.com/c/uVT1Jd0Q/202-migrate-gridlayout-from-toolkit-to-static


app/assets/stylesheets/helpers/_footer.scss
```diff
@@ -18,31 +18,24 @@
     text-decoration: none;
   }
 
-  .footer-categories {
-    @extend %contain-floats;
-    @extend %grid-row;
+  .footer-container {
+    @include media(tablet) {
+      padding-top: 60px;
```
injms: Is it worth using `govuk-spacing()`?

app/assets/stylesheets/helpers/_mixins.scss
```diff
@@ -0,0 +1,19 @@
+// this is a modified copy of the mixin from toolkit, which we're trying to remove our
+// dependency on. Hopefully this mixin will be temporary
```
injms: 🤞

app/assets/stylesheets/helpers/_footer.scss
andysellick: This is interesting - I wanted to do that, but while `static` includes the components gem, it doesn't pull in any sass (to avoid duplication where apps do that and use static) so I can't use any `govuk-frontend` mixins or variables in static. Fun times.

injms: Ouch! Fun times indeed.

PR #2017, opened 2020-01-29
Update global banner path blacklist (Transition pages)

huwd:
Trello: https://trello.com/c/DsPNCfTV/429-update-global-banner-logic-to-stop-it-appearing-on-transition

Stops transition banner showing on:
- /transition
- /transition-check (the newly re-slugged brexit checker)

Also update scroll tracking config to monitor the new /transition path.


vanitabarrett:
Maybe we should let PAs know that it was broken and is now fixed 🤔

PR #2020, opened 2020-01-30
Revert "Bump govuk_publishing_components from 21.21.2 to 21.21.3"

alex-ju:
CI fails in production (and only in production) when trying to release #2016, so we're going to revert it and test again – currently investigating the cause.


andysellick:
@alex-ju just curious - why the revert?

alex-ju:
As expected this was not the cause.

PR #2025, opened 2020-02-04
Remove govuk_toolkit shims

injms:
## What

Removing the shims that static loads in from toolkit.

There were two shims mixing coming in from toolkit. 

One was for `display: inline-block` which wasn't being used, so hasn't been moved across. 

The other was for elements that contain floated elements - a clearfix. Where possible, this has been replaced with the `govuk-clearfix` class on the element. In two places, the styles weren't affecting anything in the templates in static - so the class couldn't be added in static. In those places I've imported the shim specifically into those stylesheets rather than into the base stylesheet. 

## Why

A step along the road towards the complete removal of toolkit from static - and turning off of the compatibility mode.

## Visual differences 

None

### DIFF
--- a/app/assets/stylesheets/helpers/_footer.scss
+++ b/app/assets/stylesheets/helpers/_footer.scss
@@ -1,21 +1,16 @@
 /* Global footer */
 
 #footer {
-  // Replaces the 1px #a1acb2 border from govuk_template
-  border-top: 10px solid $govuk-brand-colour;
+  background-color: govuk-colour("light-grey");
+  border-top: 10px solid $govuk-brand-colour; // Replaces the 1px #a1acb2 border from govuk_template
 
   .govuk-link:focus {
     color: $govuk-focus-text-colour;
   }
 
-  &,
-  .footer-wrapper {
-    background-color: govuk-colour("light-grey");
-  }
-
   .footer-container {
     @include media(tablet) {
-      padding-top: 60px;
+      padding-top: govuk-spacing(9);
     }
   }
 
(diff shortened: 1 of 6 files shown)


andysellick:
Looks like the two places where you've kept the `%contain-floats` shim for are multi-step and info-notice. As far as multi-step goes it looks like only licence-finder uses that class. I'm going to take a look and see if I can remove it from there/componentise it, can this PR wait a bit for that? It would be good to remove `%contain-floats` entirely, if possible.

injms:
> Looks like the two places where you've kept the `%contain-floats` shim for are multi-step and info-notice. As far as multi-step goes it looks like only licence-finder uses that class. I'm going to take a look and see if I can remove it from there/componentise it, can this PR wait a bit for that? It would be good to remove `%contain-floats` entirely, if possible.

Super! Yep, I'll add a `do-not-merge` label so it can wait for those bits to be removed.

andysellick:
@injms licence finder updated: https://github.com/alphagov/licence-finder/pull/678

I'm going to do a second check for other instances now.

andysellick:
I _think_ it'll soon be safe to remove this entire chunk from static: https://github.com/alphagov/static/blob/master/app/assets/stylesheets/helpers/_multi-step.scss#L148-L191

The last thing using `%contain-floats` in static will then be `.info-notice` in `_text.scss`. This seems to be quite widely in use, so my suggestion would be to remove the new shim file and put the code from it directly into `_text.scss` i.e.

```
  .info-notice {
    background: none;
    border-left: 10px solid $border-colour;
    margin-bottom: 1em;

    &:after {
        clear: both;
        content: "";
        display: block;
    }
  }
```

injms:
Super - thanks @andysellick. I've made the suggested changes and pushed them up. To prevent any mishaps I'll keep it labelled as 🚧 and 'do not merge' until the licence finder pull request has been deployed to production.

PR #2026, opened 2020-02-05
Fix global bar multiple cookies

vanitabarrett:
Trello: https://trello.com/c/hA65u9qV/460-global-banner-setting-two-cookies-on-some-pages

## What
Use cookie helper functions in the global bar init logic, to match the rest of the global bar logic.

## Why
The global banner was setting two cookies on some pages, one with a path of '/' and one with the path of the page. This had the side effect of the banner not being able to be dismissed (it also didn't disappear after 3 page views). This looked like it was happening because some of the global bar functions set the cookie manually without specifying a path, whereas other functions use the cookie helper function which always set the path to '/'.

**Example page to check:**
[www.gov.uk/government/organisations/companies-house](www.gov.uk/government/organisations/companies-house)

### DIFF
--- a/app/assets/javascripts/global-bar-init.js
+++ b/app/assets/javascripts/global-bar-init.js
@@ -1,4 +1,5 @@
 //= require libs/GlobalBarHelper.js
+//= require govuk_publishing_components/lib/cookie-functions
 
 'use strict'
 window.GOVUK = window.GOVUK || {}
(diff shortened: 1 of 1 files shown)


vanitabarrett:
(sorry, still need to test on various frontend apps in integration. Will invite for review once I've done more testing)

vanitabarrett:
Ready for review now - seems to fix the issue when I tested on integration

PR #2027, opened 2020-02-05
Reinstate spacing for core layout imports

vanitabarrett:
## What
Different applications import different layouts - it looks like one layout imports the measurements scss, but one wasn't, which meant certain components (e.g: global and survey banners) were rendering without spacing.

## Before
Screenshot 2020-02-05 at 09 56 14

## After
Screenshot 2020-02-05 at 09 59 28


vanitabarrett:
@injms You probably have more context on these changes, please can you review 🙂 I'm not sure if this will affect any other layouts...

PR #2034, opened 2020-02-10
Remove duplicate global_bar_seen cookies

vanitabarrett:
We had a bug which meant that the global_bar_seen cookie was sometimes set more than once.
This bug has now been fixed (https://github.com/alphagov/static/pull/2026), but some users will be left with these duplicate cookies and therefore will continue to see the issue.

This commit introduces a function that checks for these duplicate cookies and deletes them if they're present.

### DIFF
--- a/app/assets/javascripts/global-bar-init.js
+++ b/app/assets/javascripts/global-bar-init.js
@@ -35,6 +35,32 @@ var globalBarInit = {
     return new RegExp(paths.join("|")).test(window.location.pathname)
   },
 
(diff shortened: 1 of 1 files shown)


vanitabarrett:
@alex-ju The dodgy cookie should automatically expire after 84 days, so we could create an issue to clear up this code after that time?

vanitabarrett:
Issue added here to track removal of this code: https://github.com/alphagov/static/issues/2036

alex-ju:
That's perfect, thank you!

PR #2049, opened 2020-02-21
Remove toolkit conditionals

andysellick:
## What

Remove `govuk_frontend_toolkit` conditionals from static. This involves:

- an update to include a new version of the components gem that allows static to import `govuk-frontend` mixins and variables (but no actual CSS weight)
- copying the IE conditional from toolkit
- updating all the media queries to use govuk-frontend, not toolkit
- pulling in the sass file from the gem for the cookie banner and feedback components(see below)

## Why

Just another step along the road to removing toolkit from the gem.

## Cookie banner and feedback components

In this new world of using specific sass for only the components an app is using from the gem, we need to import the sass for the cookie banner and feedback components into static, because static renders them, not the apps. At the moment this all works because all apps include all the components sass, but as soon as we start switching apps this will stop working.

Trello card: https://trello.com/c/knuRo9cF/207-replace-conditionals-from-frontend-toolkit-in-static

### DIFF
--- a/app/assets/stylesheets/header-footer-only.scss
+++ b/app/assets/stylesheets/header-footer-only.scss
@@ -1,12 +1,12 @@
+@import "govuk_publishing_components/govuk_frontend_support";
+
 // settings
 @import "settings/colours";
 @import "settings/measurements";
 @import "settings/spacing";
 
-
 /* govuk_frontend_toolkit includes */
 @import "colours";
-@import "conditionals";
 @import "css3";
 @import "device-pixels";
 @import "measurements";
--- a/app/assets/stylesheets/header-footer-only.scss
+++ b/app/assets/stylesheets/header-footer-only.scss
@@ -1,4 +1,9 @@
+$govuk-compatibility-govuktemplate: true;
+$govuk-use-legacy-palette: false;
+
 @import "govuk_publishing_components/govuk_frontend_support";
+@import "govuk_publishing_components/components/_cookie-banner";
+@import "govuk_publishing_components/components/_feedback";
 
 // settings
 @import "settings/colours";
--- a/app/assets/stylesheets/helpers/_footer.scss
+++ b/app/assets/stylesheets/helpers/_footer.scss
@@ -9,20 +9,20 @@
   }
 
   .footer-container {
-    @include media(tablet) {
+    @include govuk-media-query($from: tablet) {
       padding-top: govuk-spacing(9);
     }
   }
 
   .footer-categories {
-    @include media(tablet) {
+    @include govuk-media-query($from: tablet) {
       padding-bottom: govuk-spacing(6);
     }
 
     h2 {
       padding: govuk-spacing(2) 0 0;
 
-      @include media(tablet) {
+      @include govuk-media-query($from: tablet) {
         margin: 0 0 govuk-spacing(3) 0;
         padding: 0 0 govuk-spacing(4);
         border-bottom: 1px solid $border-colour;
@@ -35,7 +35,7 @@
       border: 1px solid $govuk-border-colour;
       border-width: 1px 0 0 0;
 
-      @include media(tablet) {
+      @include govuk-media-query($from: tablet) {
         margin-bottom: 0;
       }
     }
@@ -46,7 +46,7 @@
       padding: 0;
       margin: 0;
 
-      @include media(tablet) {
+      @include govuk-media-query($from: tablet) {
         padding-bottom: govuk-spacing(9);
       }
 
@@ -54,7 +54,7 @@
         padding: govuk-spacing(2) 0 0;
         margin-bottom: govuk-spacing(1);
 
-        @include media(tablet) {
+        @include govuk-media-query($from: tablet) {
           padding: govuk-spacing(4) 0 0;
           margin-bottom: 0;
         }
@@ -89,7 +89,7 @@
     .copyright {
       text-align: center;
 
-      @include media(tablet) {
+      @include govuk-media-query($from: tablet) {
         text-align: right;
       }
 
--- a/app/assets/stylesheets/helpers/_header.scss
+++ b/app/assets/stylesheets/helpers/_header.scss
@@ -81,7 +81,7 @@
     width: 49%;
     float: right;
 
-    @include media(desktop) {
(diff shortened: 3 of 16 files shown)


app/assets/stylesheets/helpers/_footer.scss
```diff
@@ -9,20 +9,20 @@
...
     }
 
     h2 {
       padding: govuk-spacing(2) 0 0;
 
-      @include media(tablet) {
+      @include govuk-media-query($from: tablet) {
         margin: 0 0 govuk-spacing(3) 0;
         padding: 0 0 govuk-spacing(4);
         border-bottom: 1px solid $border-colour;
```
kr8n3r:
```suggestion
        border-bottom: 1px solid $govuk-border-colour;
```

app/assets/stylesheets/header-footer-only.scss
```diff
@@ -1,12 +1,12 @@
+@import "govuk_publishing_components/govuk_frontend_support";
+
```
vanitabarrett: Sorry I'm not clear on the difference, do we need to import `@import 'govuk_publishing_components/component_support';` too?

app/assets/stylesheets/helpers/_mixins.scss
```diff
@@ -9,11 +9,11 @@
     width: 960px;
   }
 
-  @include media(tablet) {
+  @include govuk-media-query($from: tablet) {
     margin: 0 30px;
   }
 
-  @include media($min-width: (960px + 30px * 2)) {
+  @media (min-width: (960px + 30px * 2)) {
```
kr8n3r: we could you $govuk-page-width and $govuk-gutter

app/assets/stylesheets/helpers/_header.scss
```diff
@@ -483,21 +483,21 @@ $light-blue: #259EDA;
...
+    @include govuk-media-query($from: tablet) {
       .dismiss {
         position: absolute;
         right: 0;
         top: 0;
       }
     }
 
-    @include media($min-width: 320px) {
+    @include govuk-media-query($until: tablet) {
```
injms: This possibly should be from mobile?

```suggestion
    @include govuk-media-query($from: mobile) {
```

app/assets/stylesheets/header-footer-only.scss
```diff
@@ -1,12 +1,16 @@
+$govuk-compatibility-govuktemplate: true;
+$govuk-use-legacy-palette: false;
+
+@import "govuk_publishing_components/govuk_frontend_support";
+@import "govuk_publishing_components/components/_cookie-banner";
+
```
vanitabarrett: We seem to be pulling in the feedback component, do we need to include the css for it too? https://github.com/alphagov/static/blob/6a0d8a619e56063c81c193458bff92618f163306/app/views/root/_error_page.html.erb#L39

app/assets/stylesheets/static.scss
```diff
@@ -1,6 +1,12 @@
 // settings
 @import "settings/measurements";
 
+$govuk-compatibility-govuktemplate: true;
+$govuk-use-legacy-palette: false;
+
+@import "govuk_publishing_components/govuk_frontend_support";
```
injms: :tada: woop for individual importing!

app/assets/stylesheets/header-footer-only.scss
andysellick: Oooh, interesting, hadn't thought about that. I'll look into that, thanks.

app/assets/stylesheets/header-footer-only.scss
andysellick: The component support stuff is if you're pulling in component sass - it's the mixins and variables for the components. It shouldn't be needed (the cookie banner doesn't use anything from it) but in the light of your comment about the feedback component, I might need to rethink it. Will have a look.

app/assets/stylesheets/helpers/_header.scss
injms: Think that this should be from mobile to match `min-width: 320px`:

```suggestion
    @include govuk-media-query($from: mobile) {
```

app/assets/stylesheets/helpers/_mixins.scss
andysellick: Thanks. My initial reaction to this line was I didn't want to touch it with a barge pole! But I'll update it.

app/assets/stylesheets/helpers/_header.scss
andysellick: Updated, thanks.

app/assets/stylesheets/helpers/_mixins.scss
kr8n3r: i'd try `@include govuk-media-query(($govuk-page-width + $govuk-gutter * 2));` and hope

andysellick: Could we just use `$from: desktop` ?

app/assets/stylesheets/header-footer-only.scss
andysellick: Yes, we do now. I've added it. It's not ideal. I originally split them into the two imports (component and govuk_frontend) because some of it adds CSS weight (specifically the branding colours) and static only needs govuk-frontend (not the branding, which would be duplicated in apps as well).

With this development we might as well only have the one support file. Rather than going back to the gem and trying to come up with a solution I think we should go ahead with this addition and I'll try to work up an alternative later. Sorry, slight tech debt introduction here.

app/assets/stylesheets/header-footer-only.scss
andysellick: Feedback component sass added.

app/assets/stylesheets/helpers/_mixins.scss
kr8n3r: nein, because
// As soon as the viewport is greater than the width of the page plus the
// gutters, just centre the content instead of adding gutters.

andysellick: Updated, thanks. I'll make sure this gets tested thoroughly with some apps.

PR #2057, opened 2020-03-02
Hide the global banner

vanitabarrett:
## What
Hide the global banner

## Why
The fact we've left the EU is known, and transition information is well signposted around GOV.UK.


DilwoarH:
Should we remove the code for the eu exit banner?

vanitabarrett:
@DilwoarH We don't want to remove the code as we may need the banner in future. This is essentially a component like the ones we have in `govuk_publishing_components`, but it lives in Static because we don't have a mechanism for deploying banners if they live in the gem yet.

PR #2060, opened 2020-03-04
Hide global bar on the referred page

alex-ju:
Replace hardcoded blacklisted paths with the path of the page the call to action link points to.

Follow up on #2057 and #2059.

### DIFF
--- a/app/assets/javascripts/global-bar-init.js
+++ b/app/assets/javascripts/global-bar-init.js
@@ -26,7 +26,8 @@ var globalBarInit = {
 
   blacklistedUrl: function() {
     var paths = [
-      "/done"
+      "/done",
+      "/transition-check"
     ]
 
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/global-bar-init.js
```diff
@@ -26,12 +26,14 @@ var globalBarInit = {
 
   blacklistedUrl: function() {
     var paths = [
-      "/register-to-vote",
-      "/done",
-      "/transition",
-      "/transition-check"
```
timblair: Please leave `/transition-check` in the blacklist. The others (Transition landing page, and RtV start page) can come out (so the banner shows on them).

chao-xian:
Ah sorry I didn't realise we had this blocklist.

PR #2062, opened 2020-03-04
Move global-bar-init in body_end

alex-ju:
Because the init runs right away (without awaiting the page to load) it's not able to query the markup elements (as they aren't loaded yet). [See this issue in integration](https://www.integration.publishing.service.gov.uk/government/topical-events/coronavirus-covid-19-uk-government-response).

Follow up on #2060.

### DIFF
--- a/app/views/notifications/_global_bar.html.erb
+++ b/app/views/notifications/_global_bar.html.erb
@@ -12,7 +12,7 @@
 -%>
 
 <% if show_global_bar %>
-  <% content_for :head do %>
+  <% content_for :body_end do %>
     <%= javascript_include_tag 'global-bar-init', integrity: false %>
     <script>
       if (GOVUK.globalBarInit) {


app/views/notifications/_global_bar.html.erb
```diff
@@ -12,7 +12,7 @@
 -%>
 
 <% if show_global_bar %>
-  <% content_for :head do %>
+  <% content_for :body_end do %>
     <%= javascript_include_tag 'global-bar-init', integrity: false %>
```
vanitabarrett: I think this was originally included right away to avoid any kind of flash of the banner before disappearing, e.g: if someone has manually hidden it or seen the banner 3 times already. Are we sure this change doesn't lead to that happening?

alex-ju: As I'm reading the `global-bar-init.js` we show the banner when the script kicks in based on a set of conditions. Will double/triple check all scenarios to be sure.

alex-ju: Tested before and after accepting cookies, showing it 3 times or manually hiding it. It's now [deployed in integration](https://www.integration.publishing.service.gov.uk). I don't see any change in behavior.

alex-ju: It will become a component one day 😇 so we'll be more confident it works similarly with any other script.

PR #2063, opened 2020-03-04
Rename blacklist to blocklist

chao-xian:


spec/javascripts/global-bar-init.spec.js
```diff
@@ -10,7 +10,7 @@ describe('Global bar initialize', function () {
   })
 
   it('does not set cookie for blacklisted URL', function() {
```
vanitabarrett: Can we replace it here too?

alex-ju:
```suggestion
  it('does not set cookie for blocklisted URL', function() {
```

Can we replace it for the survey-related code too? 🙏

chao-xian: Good spot! Thanks

alex-ju: Ah, Vanita already spotted it.

chao-xian: Just spotted that - will do

PR #2066, opened 2020-03-05
Add tracking to see where the JS is being used

injms:
Update to (temporarily) use a branch of the frontend toolkit gem, which uses the branch `add-tracking-see-where-js-is-being-used` of govuk_frontend_toolkit.

If a user has given permission, this will fire a GA event to let us know which JavaScript is being used and where. 

We can then see if / where these JavaScripts are being used and know where to provide a suitable replacement.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -38,6 +38,7 @@ group :development, :test do
 end
 
 gem "gds-api-adapters", "~> 63.5"
-gem "govuk_frontend_toolkit", "~> 9.0.0"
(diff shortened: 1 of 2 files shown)


injms:
@andysellick - sure thing: https://github.com/alphagov/govuk_frontend_toolkit/pull/484

alex-ju:
Is there an outcome of this investigation? I was looking at the `govuk_frontend_toolkit` dependency in static and I don't see any obvious reason for keeping it.

PR #2067, opened 2020-03-06
Only import feedback component CSS on error pages

vanitabarrett:
## What

- Remove imports of feedback component CSS in general static stylesheets (e.g: static, core, header-and-footer-only). Instead, only import the feedback component where it's used in the error-page.css.
- Don't import all components in the error pages - we don't need everything
- Don't import error page css in the general css file. The error page template supplies a stylesheet tag for the error-page.css
- Add govuk-link class to error page content for consistent focus states

## Why
The feedback component is only shown on error pages within Static. In other cases, the frontend applications themselves are pulling in the feedback component HTML and CSS.

The existing behaviour meant we had duplicated CSS for the feedback component as it was being imported once by static and once by the frontend app. This commit stops that from happening.

## Visual (should look no different)
static dev gov uk_templates_404 html erb

### DIFF
--- a/app/assets/stylesheets/error-page.scss
+++ b/app/assets/stylesheets/error-page.scss
@@ -1,2 +1,8 @@
 $govuk-compatibility-govuktemplate: true;
-@import "govuk_publishing_components/all_components";
+$govuk-use-legacy-palette: false;
+
+@import "govuk_publishing_components/govuk_frontend_support";
+@import "govuk_publishing_components/component_support";
+@import "govuk_publishing_components/components/_cookie-banner";
+@import "govuk_publishing_components/components/_button";
+@import "govuk_publishing_components/components/_feedback";
--- a/app/assets/stylesheets/header-footer-only-print.scss
+++ b/app/assets/stylesheets/header-footer-only-print.scss
@@ -2,8 +2,6 @@
 $govuk-compatibility-govuktemplate: true;
 $govuk-use-legacy-palette: false;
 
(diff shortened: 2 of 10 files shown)


app/assets/stylesheets/header-footer-only-print.scss
```diff
@@ -2,8 +2,6 @@
 $govuk-compatibility-govuktemplate: true;
 $govuk-use-legacy-palette: false;
 
-@import "govuk_publishing_components/components/print/feedback";
```
andysellick: Do we not need the print styles for the feedback component somewhere? Like in a new error page print stylesheet?

app/assets/stylesheets/error-page.scss
```diff
@@ -1,2 +1,8 @@
 $govuk-compatibility-govuktemplate: true;
-@import "govuk_publishing_components/all_components";
+$govuk-use-legacy-palette: false;
+
+@import "govuk_publishing_components/govuk_frontend_support";
+@import "govuk_publishing_components/component_support";
```
andysellick: Presumably component_support is needed for the button sass? Otherwise this line could be removed.

vanitabarrett: Yeah, without it I get the error: `Undefined mixin 'responsive-bottom-margin'.`

andysellick: Fair enough then.

PR #2068, opened 2020-03-09
Add list subheading as optional attribute to ecommerce tracking

huwd:
trello: https://trello.com/c/wyHXRVll/138-add-ecommerce-tracking-to-mainstream-browse

Depends on: https://github.com/alphagov/static/pull/2072

## What

Adds an optional `list-subheading` attribute that will be sent to google analytics as dimension 94. If the tracked element has an attribute `ecommerce-subtitle` the value will be sent, otherwise dimension94 will not be included in the playload to GA.

## Why

We are adding ecommerce tracking to many of our browse pages. Some lists have subheadings that internally group rows. It is a requirement that we are able to track these too.

This work in an effort to build better baseline data around the browse experience on GOV.UK ahead of considering changes.

### DIFF
--- a/app/assets/javascripts/analytics/ecommerce.js
+++ b/app/assets/javascripts/analytics/ecommerce.js
@@ -19,20 +19,24 @@
 
       ecommerceRows.each(function(index, ecommerceRow) {
         var $ecommerceRow = $(ecommerceRow);
-
+        var listSubheading  = $ecommerceRow.data('ecommerce-subheading') || undefined;
         var contentId = $ecommerceRow.attr('data-ecommerce-content-id'),
           path = $ecommerceRow.attr('data-ecommerce-path');
 
-        addImpression(contentId, path, index + startPosition, searchQuery, listTitle, variant);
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/analytics/ecommerce.js
```diff
@@ -23,18 +24,22 @@
...
 
-    function constructData(contentId, path, position, listTitle, searchQuery, variant) {
+    function constructData(contentId, path, position, listTitle, listSubheading, searchQuery, variant) {
       var data = {
         position: position,
         list: listTitle,
         dimension71: searchQuery
       }
 
+      if (listSubheading !== undefned) {
```
theseanything: I think you can also just do `if (listSubheading)`, which will catch the null case too? Hmm, but will it ever be an empty string?

But otherwise, there's a typo:
```suggestion
      if (listSubheading !== undefined) {
```
Is this code tested anywhere?

huwd: Fixed, thanks!

PR #2083, opened 2020-03-12
Fix :visited :focus state for link in banner

injms:
## What

The focus state for the link in the banner was white when visited, instead of black.

## Why

Fixed to meet the contrast criteria to make the link readable and accessible.

## Visual changes

Before:
image

After:

image


injms:
Thanks @alex-ju 😄

alex-ju:
With 3 approvals it would've been a shame to leave it pending.

PR #2084, opened 2020-03-12
Refactor banner link focus states

injms:
The global link styles from static / govuk_template were clashing with `govuk-link` - so the link styles for the warning banner layout have been updated to not use that class.
 
The focus state styles comes from the GOV.UK Frontend mixin.

### DIFF
--- a/app/assets/stylesheets/helpers/_header.scss
+++ b/app/assets/stylesheets/helpers/_header.scss
@@ -528,12 +528,20 @@ $light-blue: #259EDA;
   padding: govuk-spacing(5) 0;
 
   .global-bar-message {
-    .global-bar-title:link {
+    .global-bar-title {
       color: govuk-colour("white");
+
+      &:focus {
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/helpers/_header.scss
```diff
@@ -528,12 +528,20 @@ $light-blue: #259EDA;
...
     }
   }
 
-  .global-bar__dismiss:link {
+  .global-bar__dismiss {
     color: govuk-colour("white");
+
+      &:focus {
+        @include govuk-focused-text;
+      }
```
andysellick: Nitpick: unindent lines 543 to 545.

injms: That is a tomorrow problem 😄

injms: Fixed here: https://github.com/alphagov/static/pull/2086

PR #2088, opened 2020-03-13
Remove the tracking that fires when old JS is initialised

injms:
This used up a lot - a lot! - of events, so is being reverted to prevent the quota from being exhausted. 

The gem has been commented out rather than deleted, since we will need to use it again to check again where the JavaScript is being initialised in the near future.


chao-xian:
@injms Could you delete the line please rather than comment it out.

PR #2096, opened 2020-03-20
Update links for coronavirus landing page

sihugh:
Because /coronavirus currently redirects to the topical event page, this
can be merged and deployed whenever we like.

The topical event page will be redirected back to the new landing page once it
it published.

https://trello.com/c/9EMnRGEh/38-update-the-banner-link


sihugh:
I've made a new card for the popular links 👍

PR #2101, opened 2020-03-23
Now matches the exact path for url blocking

DilwoarH:
Previously it was checking for segments so if we had `/coranavirus` blocked, `/guidance/coronavirus-covid-19-information-for-the-public` would be blocked too as it has `/coronavirus` in the middle.

This change now forces the regex to check exact match so it would only match if the path is `/coronavirus` not `/guidance/coronavirus-covid-19-information-for-the-public`.

Trello: https://trello.com/c/a83JfZDG/60-check-why-the-banner-doesnt-appear-on-certain-pages

### DIFF
--- a/app/assets/javascripts/global-bar-init.js
+++ b/app/assets/javascripts/global-bar-init.js
@@ -26,14 +26,15 @@ var globalBarInit = {
 
   urlBlockList: function() {
     var paths = [
-      "/done",
-      "/transition-check",
-      "/coronavirus"
+      "^/done",
+      "^/transition-check$",
+      "^/coronavirus$"
     ]
 
     var ctaLink = document.querySelector('.js-call-to-action')
     if (ctaLink) {
-      paths.push(ctaLink.getAttribute('href'))
+      var ctaPath = "^" + ctaLink.getAttribute('href') + "$"
+      paths.push(ctaPath)
     }
 
     return new RegExp(paths.join("|")).test(window.location.pathname)


app/assets/javascripts/global-bar-init.js
```diff
@@ -36,7 +36,11 @@ var globalBarInit = {
       paths.push(ctaLink.getAttribute('href'))
     }
 
-    return new RegExp(paths.join("|")).test(window.location.pathname)
+    var regexForPaths = paths.map(function(path) {
+      return "^" + path + "$";
```
sihugh: I've got a feeling that the other paths in this might be prefix things - `/done` goes before a bunch of pages and I think the transition checker has lots of sub pages too.  Maybe if we drop the trailing `$`?

DilwoarH: Good point.

If I drop the trailing `$` then this would match `/guidance/coronavirus-covid-19-information-for-the-public`.

There are two approaches we can take:
1. Two separate lists, one which has a wildcard match and another which is an exact match. Before we create the regex, we would merge it together.
2. Put the regex directly in the list so that we can specify the paths as we want.

```javascript

var paths = [
      "^/done",
      "^/transition-check$",
      "^/coronavirus$"
    ]

```

DilwoarH: @sihugh - updated

PR #2107, opened 2020-03-24
Reinstate the search box in header

chao-xian:
Reverts alphagov/static#2106


erino:
Revert "Revert "Revert "Revert "Revert

PR #2109, opened 2020-03-25
New COVID-19 banner design

vanitabarrett:
Trello: https://trello.com/c/ATJOnCR2/78-design-and-build-an-update-to-the-sitewide-emergency-banner-that-re-enforces-the-stay-home-message

## What
Introduces:

- a new banner design, using PHE campaign colours
- repurposes the existing global_bar_seen cookie to hide parts of the banner rather than the whole banner
- adds logic to set the global_bar_seen cookie on pages which are included in the URL blocklist (but still continue to hide the banner) - this will help us do things like automatically hide the banner if a person has previously visited a page in the blocklist.

## Before
Screenshot 2020-03-25 at 12 14 42

## After
**Full banner**
Screenshot 2020-03-25 at 15 15 26

**Dismissed**
Screenshot 2020-03-25 at 12 14 59


app/assets/stylesheets/helpers/_header.scss
```diff
@@ -472,76 +472,104 @@
   display: none;
 }
 
-$light-blue: #259EDA;
+$covid-green: #7eff8e;
```
alex-ju: Non-blocker, we may be able to re-use [`$gem-c-covid-green` from the gem](https://github.com/alphagov/govuk_publishing_components/blob/a154b2c60b6d8c80cf41e66ee63557353627548e/app/assets/stylesheets/govuk_publishing_components/components/helpers/_brand-colours.scss#L98).

app/assets/stylesheets/helpers/_header.scss
```diff
@@ -472,76 +472,104 @@
...
 
 .global-bar {
-  background-color: #E5EAF1;
-  border-top: 8px solid $govuk-brand-colour;
+  background-color: govuk-colour("black");
+  border-top: 5px solid $covid-green;
```
alex-ju: can use `$govuk-border-width` now that we're using standard values

app/assets/stylesheets/helpers/_header.scss
```diff
@@ -472,76 +472,104 @@
...
+      }
     }
 
-    @include govuk-media-query($from: tablet) {
-      .global-bar__dismiss {
-        position: absolute;
+    .global-bar__dismiss {
+      display: none;
+
+      &--show {
```
alex-ju: non-blocker, we tend to explicitly expand these `.global-bar__dismiss--show` to be able to find them quickly using search.

alex-ju:
> * if I click 'hide message' then refresh the page it comes back, is that intentional?

Is it happening even after you accepted cookies? Doesn't happen to me

andysellick:
@alex-ju retested, yes, only seems to happen if I haven't accepted cookies.

vanitabarrett:
@andysellick 

> 'Stay at home' text is close to the list text beneath it (closer than on the covid page itself)

Thanks, I'll add a bit more spacing there.

> this new black banner connects with the black sidebar of the homepage, is that odd? Should there be a gap?

This seems to happen when I zoom/magnify the page, but not otherwise - is that the same for you? That's not changed (it happens with the current banner), so given the time pressure I'm going to say it's out of scope here.

> if I click 'hide message' then refresh the page it comes back, is that intentional?
This will happen if you haven't accepted cookies as there is nowhere to store your preference.

> the images disappear on mobile, is that intentional?
Yes, this is as per the design

app/assets/stylesheets/helpers/_header.scss
vanitabarrett: @alex-ju That doesn't seem to work - it says `Undefined variable: "$gem-c-covid-green"`

andysellick:
@vanitabarrett ah yes, the banner touching the black box is normal at tablet (?) I think. Fine, don't worry about it.

app/assets/stylesheets/helpers/_header.scss
alex-ju: Ah, go with a hardcoded value then – it may be that we don't include all those SCSS files at the moment.

PR #2110, opened 2020-03-25
Stop surveys appearing when the global bar is visible

cbaines:
Or on /coronavirus. I believe the concern here is regarding screen
space on these pages.

### DIFF
--- a/app/assets/javascripts/surveys.js
+++ b/app/assets/javascripts/surveys.js
@@ -318,6 +318,7 @@
     pathInBlocklist: function () {
       var blackList = new RegExp('^/(?:' +
         /service-manual/.source +
+        /|coronavirus/.source +
         // add more blacklist paths in the form:
         // + /|path-to\/blacklist/.source
         ')(?:\/|$)'
@@ -390,7 +391,8 @@
       var notificationIds = [
         '.govuk-emergency-banner:visible',
         '#global-browser-prompt:visible',
-        '#taxonomy-survey:visible'
+        '#taxonomy-survey:visible',
+        '#global-bar:visible', // Currently about Coronavirus
       ]
       return $(notificationIds.join(', ')).length > 0
     },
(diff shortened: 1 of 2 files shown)


spec/javascripts/surveys.spec.js
```diff
@@ -468,13 +468,13 @@ describe('Surveys', function () {
       expect(surveys.canShowAnySurvey()).toBeFalsy()
     })
 
-    it('returns true otherwise', function () {
+    xit('returns true otherwise', function () {
```
barrucadu: xit?

app/assets/javascripts/surveys.js
```diff
@@ -390,7 +391,8 @@
       var notificationIds = [
         '.govuk-emergency-banner:visible',
         '#global-browser-prompt:visible',
-        '#taxonomy-survey:visible'
+        '#taxonomy-survey:visible',
+        '#global-bar:visible', // Currently about Coronavirus
```
benthorner: Could you explain how this relates to the commit message? I think the global bar is shown on a lot of pages, so this seems like it would significantly decrease the number of times the survey banner is shown, as the user would have to first dismiss the global bar, and reload the page (or go to another one).

spec/javascripts/surveys.spec.js
cbaines: Apparently the way to mark tests as pending. Because the intent here is to stop surveys showing, this unsurprisingly breaks a couple of tests.

app/assets/javascripts/surveys.js
cbaines: So this change matches up with the "Stop surveys appearing when the global bar is visible" bit of the commit message, with the motivation being "the concern here is regarding screen
space on these pages.".

I think that significantly reducing the survey banner appearing is an intended consequence, but I'll try to confirm that.

cbaines: @benthorner I've double checked, and the intent is indeed to stop the survey banner showing on all pages.

I realise that these changes aren't the simplest route to that point, but I thought it best to try and represent the reasoning around why in the code. Which is why the logic is don't show surveys if the global banner is visible.

benthorner: Thanks for checking and explaining 👍.

PR #2124, opened 2020-04-03
Exclude business page from the coronavirus banner

sihugh:
We'll show a breadcrumb back to the coronavirus page.

We don't need to include it in the "auto-hide the banner if you've seen this page" logic because the banner duplicates /coronavirus content and not /coronavirus/business-support logic

We can remove the explicit coronavirus link here too, as it's picked up by the call to action code below the exclusion list

https://trello.com/c/CXuwza3f/136-hide-the-coronavirus-banner-on-the-business-landing-page

### DIFF
--- a/app/assets/javascripts/global-bar-init.js
+++ b/app/assets/javascripts/global-bar-init.js
@@ -26,7 +26,8 @@ var globalBarInit = {
 
   urlBlockList: function() {
     var paths = [
-      "^/coronavirus$"
+      "^/coronavirus$",
+      "^/coronavirus/business-support$"
     ]
 
     var ctaLink = document.querySelector('.js-call-to-action')
--- a/app/assets/javascripts/global-bar-init.js
+++ b/app/assets/javascripts/global-bar-init.js
@@ -26,7 +26,6 @@ var globalBarInit = {
 
   urlBlockList: function() {
     var paths = [
-      "^/coronavirus$",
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/global-bar-init.js
```diff
@@ -26,7 +26,8 @@ var globalBarInit = {
 
   urlBlockList: function() {
     var paths = [
-      "^/coronavirus$"
+      "^/coronavirus$",
```
vanitabarrett, line 29: Is this line needed? The banner should automatically hide on the page it links to.

sihugh: I'm not sure!

sihugh: I might be brave and try removing it

vanitabarrett: 💪

sihugh: Updated

DilwoarH:
Wouldn't this mean it will show in /coronavirus now?

vanitabarrett:
@DilwoarH No, the code here means that the banner automatically hides on the page it links to
https://github.com/alphagov/static/pull/2124/files#diff-03510c40f1525e0921539b6d1ad823f4R32

PR #2135, opened 2020-04-14
Add icon for banner

DilwoarH:
Iterate the banner design to try to increase the clickthrough rate.

Design details:

- Mobile design matches the landing page header CTA
(breaks across two lines, with a larger arrow icon)
- The CTA link needs a discretionary line break after the colon
- Delete the line break between the last two sentences of the banner on both mobile and desktop (To match landing page)

Our analysis has shown that the switch from the red to black banner increased interaction (hides) but reduced clickthroughs.

CTR for old banner (red) = 4.3%
CTR for current banner (black) = 3.7%

Interaction rate for old banner= 5.2%
Interaction rate for current banner= 5.7%

https://trello.com/c/swfdTfOk/161-strengthen-the-emergency-banner-link


app/assets/images/covid-arrow.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="38" height="38" viewBox="0 0 38 38" fill="none" xmlns="http://www.w3.org/2000/svg">
+  <path fill-rule="evenodd" clip-rule="evenodd" d="M37.9995 19C37.9995 29.4934 29.4929 38 18.9995 38C9.43459 38 1.52044 30.9322 0.194803 21.7343L21.0336 21.7343L13.7809 28.9869L16.6093 31.8153L27.5877 20.8369L29.0019 19.4227L27.5877 18.0085L16.6093 7.03012L13.7809 9.85855L21.6566 17.7343L0.0410156 17.7343C0.692375 7.83098 8.93144 0 18.9995 0C29.4929 0 37.9995 8.50659 37.9995 19Z" fill="#FFE013"/>
```
vanitabarrett, line 2: Worth checking that the fill colour is the correct shade - when we switched to the yellow colour, we noticed that the original SVG colour didn't match the new yellow hex code in the CSS.

app/views/notifications/_global_bar.html.erb
```diff
@@ -15,6 +16,7 @@
   title_classes = %w(global-bar-title)
   title_classes << "js-call-to-action" if title_href
   title_classes << "govuk-link" if title_href
+  title_classes << "global-bar-title__icon" if title_icon
```
andysellick, line 19: Is there a test for this?

app/assets/images/covid-arrow.svg
DilwoarH: ahh It's not - the covid yellow colour is `#ffe114`

app/assets/stylesheets/helpers/_header.scss
```diff
@@ -516,6 +516,22 @@ $covid-grey: #262828;
   }
 }
 
+.global-bar-title__icon {
+  display: inline-block;
+  display: -ms-flexbox;
```
andysellick, line 521: Vertical centering doesn't work in IE11 downwards. I think this is because IE requires an explicit height to be set for vertical centering to work (which I wouldn't recommend doing). Not sure what to suggest.

Screenshot 2020-04-14 at 14 33 39

app/assets/stylesheets/helpers/_header.scss
```diff
@@ -516,6 +516,22 @@ $covid-grey: #262828;
...
+  align-items: center;
+  box-sizing: border-box;
+  padding-left: 50px;
+  min-height: 45px;
+  background-image: image-url("covid-arrow.png"); // PNG fallback for SVG
+  background-image: image-url("covid-arrow.svg");
+  background-repeat: no-repeat;
+  background-position: 0 50%;
+  background-size: 37px auto;
+}
```
andysellick, line 533: Focus state hides the arrow:

Screenshot 2020-04-14 at 15 42 41

app/assets/stylesheets/helpers/_header.scss
andysellick: We seem to have solved this problem for the same link on https://www.gov.uk/coronavirus. Suggest you copy the markup/styles from there.

app/assets/stylesheets/helpers/_header.scss
andysellick: We seem to have solved this problem for the same link on https://www.gov.uk/coronavirus. Suggest you copy the markup/styles from there.

DilwoarH: I've updated it to match.

app/assets/stylesheets/helpers/_header.scss
DilwoarH: Looks like the solution was adding an explicit height

DilwoarH:
> Unfortunately the change I suggested to fix the page in old IE has caused a new layout problem, see below. Let's forget trying to fix that in this PR (delete that commit) and we'll try to solve it separately.
> 
> Screenshot 2020-04-15 at 14 47 00

Looks like it had an old deploy, here are the new changes:
Screenshot 2020-04-16 at 10 17 11

PR #2142, opened 2020-04-23
Add Smart Surveys domain to decorate links

chao-xian:
We already do this for forms[1] but this is a quick and simple
mechanism we already have in using the allowLinker feature in GA.

[1] https://github.com/alphagov/static/blob/master/app/assets/javascripts/surveys.js#L204


chao-xian:
There's a better and cleaner way to do this in government-frontend, plus the clientId from this mechanism isn't exactly what we want.

PR #2146, opened 2020-04-24
Add button component sass to static

andysellick:
Buttons in the cookie banner are unstyled in some applications following the move to individual component sass inclusion, if the application in question does not include buttons already.

e.g. calendars:

Screenshot 2020-04-24 at 09 23 20

Adding the button sass into static like this fixes this problem but introduces duplicate CSS for users where applications already include the button sass, so this is a temporary solution while we figure out something better.

### DIFF
--- a/app/assets/stylesheets/static.scss
+++ b/app/assets/stylesheets/static.scss
@@ -2,6 +2,7 @@ $govuk-compatibility-govuktemplate: true;
 $govuk-use-legacy-palette: false;
 
 @import "govuk_publishing_components/govuk_frontend_support";
+@import "govuk_publishing_components/components/_button";
 @import "govuk_publishing_components/components/_cookie-banner";
 
 @import "header-footer-only";


ChrisBAshton:
Build error - do you need to include something else?

```
Sass::SyntaxError: Undefined mixin 'responsive-bottom-margin'.
[0m[91m/usr/local/bundle/gems/govuk_publishing_components-21.41.4/app/assets/stylesheets/govuk_publishing_components/components/_button.scss:22:in `responsive-bottom-margin'
/usr/local/bundle/gems/govuk_publishing_components-21.41.4/app/assets/stylesheets/govuk_publishing_components/components/_button.scss:22
```

vanitabarrett:
Do we need to add it here too? https://github.com/alphagov/static/blob/master/app/assets/stylesheets/header-footer-only.scss#L5

andysellick:
Argh.

PR #2152, opened 2020-04-29
Remove GOVUKTracker

kevindew:
I came across this code while exploring the options for moving static's
asset files from being served from the www host rather than the assets
one. It proved to be a blocking issue as it required the presence of a
"/static/a" on the assets host.

Further investigation led me to question whether this data is actually
used for anything and I could find no evidence that we do anything other
than record this data nor anyone who knew about it. For reference this has
existed for 3 years, stores ~100 GB in S3 and produces HTTP requests that
mirror each Google Analytics one.

The simplest option to get past the blocking issue seemed to be removing
this code, which then allows cleaning up the configuration in AWS and
Fastly CDN. If we decide we have a use for this later we can always
resurrect it.


kevindew:
> This looks like a positive change to me, not least because we should avoid using two systems to store analytics data. I think this is worth trying out: it shouldn't cause an issue for users, and we can always revert it if it turns out we are using the S3 data.
> 
> Before this goes in, it would help to have a bit more explanation about what this code was/is doing. The code seems to support GA, but also sending to a `gifURL` (?!). It would be helpful to know:
> 
> * What the `/static/a` route actually does. Can we delete it once this is deployed?
> * How we can be sure [the GA functionality](https://github.com/alphagov/static/pull/2152/files#diff-539c184378356356d3dd163357913753L118) really is unused.

Thanks for having a look and sure, I should have added more in the description. It's been a wild journey of discovery around this functionality.

So as far as I have been able to tell (with various slack questions): 

- if you have accepted cookies, any requests to GA also send a request to https://assets.publishing.service.gov.uk/static/a with a lengthy of query string
- Assets CDN has a route set-up for /static/a to return a empty 200 response
- Assets CDN logs all those requests to a bucket
- Nothing else happens

To try verify what would happen without this I did various searches to find any references and came up a blank, no-one in #govuk-developers or #analysis seems to know about it either. So I changed the path used for the bucket so that if there was any processes using this data they would break, that change went through last Friday. So I'm now confident that nothing that runs regularly is using it (well at least something monitored).

Once this is removed and deployed I will proceed to remove the logging from Fastly and the route from our CDN config.

My current conclusion is that this was a well intentioned idea that never actually led to anything.

kevindew:
> @kevindew thanks for adding more comments - it's reassuring to understand the confidence we have this is unused. I'm still unsure about the interaction of this code with GA, though:
> 
> * It looks like [this line](https://github.com/alphagov/static/pull/2152/files#diff-539c184378356356d3dd163357913753L121) means the tracker does send stuff to GA.
> * The "Google Analytics Universal Tracker" also [does something similar](https://github.com/alphagov/static/blob/master/app/assets/javascripts/analytics_toolkit/google-analytics-universal-tracker.js#L191).
> 
> Wouldn't removing the GOVUKTracker mean we lose all the events in GA that aren't covered by the universal one? It _looks_ like they're doing roughly the same thing, so maybe only the universal one is actually active? I'm a bit confused why we've got two.
> 
> However, I think the "fail fast" reasoning still applies. We shouldn't have to do this much investigation to justify someone else's code. Let's ditch it!

Thanks - sorry I didn't appreciate fully what you meant first time around and now looked into it more.

So the call to `global.ga` doesn't actually send any data but is used as an asynchronous callback to access GA properties: https://developers.google.com/analytics/devguides/collection/analyticsjs/command-queue-reference#ready-callback

So what the code is doing is adding the GA tracking id to the data send to the /static/a URL.

PR #2161, opened 2020-05-07
Add new generic unprocessable entity page

benthorner:
https://trello.com/c/Uzpw1bDA/243-document-how-error-pages-work-on-govuk

Related to: https://github.com/alphagov/govuk-puppet/pull/10346

Extracted from: https://github.com/alphagov/email-alert-frontend/pull/743

### DIFF
--- a/app/views/root/422.html.erb
+++ b/app/views/root/422.html.erb
@@ -1,0 +1,7 @@
+<%= render partial: "error_page", locals: {
+    title: "Sorry, there was a problem",
+    heading: "Sorry, there was a problem",
+    intro: '<p>We cannot show you the page you were expecting.</p>
+      <p>Go back and try again.</p>',
+    status_code: 422
+} %>


kevindew:
@benthorner I don't think we should have this one as a generic error page as it stops us from rendering any non-generic 422 pages, unless I've missed some trick we have?

When we render a form with errors it's appropriate to use the 422 status code however this view would clobber the page.

We have a long, long running problem with this same situation with gone pages where they get clobbered by the page from static which means we have to do the rather grim thing of rendering a 200 response for gone pages with a customised response. Example:  https://www.gov.uk/government/publications/public-inquiries-timeliness-of-case-listings--2

benthorner:
@kevindew yes, that's right. We wanted to add this in favour of the previous ["override of an override" approach](https://github.com/alphagov/slimmer/pull/246). The thinking here is that we already return "200" [for some unprocessable entity responses](https://github.com/alphagov/email-alert-frontend/blob/master/app/controllers/subscriber_authentication_controller.rb#L22), in order to get around Slimmer casting them all to a "500". 

While we have [a related PR to remove that problem](https://github.com/alphagov/slimmer/pull/245), I think we still need some kind of fallback error page. The problem is: we can't have a fallback _and_ have specific error handling, unless we replicate the error pages in each app (or remove them from Nginx).

I agree it's not semantically correct to return "200"s in order to handle errors more specifically. The positive side is that our apps don't need to be watertight in order to show something vaguely helpful. I guess the downside is bad caching behaviour.

Looking at [our logs in more details (see comment)](https://trello.com/c/Uzpw1bDA/243-document-how-error-pages-work-on-govuk), perhaps 422 can be an exception to the rule, and the app must cope with it. This means we'd need to do some work on a few apps, though. Do you think we should adopt a different approach?

> Example: https://www.gov.uk/government/publications/public-inquiries-timeliness-of-case-listings--2

Err, this might not be the finest example of a need for this: "duplication error"

kevindew:
> @kevindew yes, that's right. We wanted to add this in favour of the previous ["override of an override" approach](https://github.com/alphagov/slimmer/pull/246). The thinking here is that we already return "200" [for some unprocessable entity responses](https://github.com/alphagov/email-alert-frontend/blob/master/app/controllers/subscriber_authentication_controller.rb#L22), in order to get around Slimmer casting them all to a "500".
> 
> While we have [a related PR to remove that problem](https://github.com/alphagov/slimmer/pull/245), I think we still need some kind of fallback error page. The problem is: we can't have a fallback _and_ have specific error handling, unless we replicate the error pages in each app (or remove them from Nginx).

Yup, that's the real problem. It's frustratingly inconvenient to follow HTTP spec and have the nginx pages. I tried looking into whether you could configure nginx to allow more control but that didn't look simple.

> I agree it's not semantically correct to return "200"s in order to handle errors more specifically. The positive side is that our apps don't need to be watertight in order to show something vaguely helpful. I guess the downside is bad caching behaviour.
> 
> Looking at [our logs in more details (see comment)](https://trello.com/c/Uzpw1bDA/243-document-how-error-pages-work-on-govuk), perhaps 422 can be an exception to the rule, and the app must cope with it. This means we'd need to do some work on a few apps, though. Do you think we should adopt a different approach?

Yup I'm keen for 422 to be something apps support. I worry that once this change settles in it'll be quite hard to get out of (as is the case with gone).
> 
> > Example: https://www.gov.uk/government/publications/public-inquiries-timeliness-of-case-listings--2
> 
> Err, this might not be the finest example of a need for this: "duplication error"

Ha yup, though to me it's even worse that this returns a 200. I don't think there is anything preventing search engines from indexing it 😱 https://www.google.com/search?q=site:www.gov.uk+%22The+information+on+this+page+has+been+removed+because+it+was+published+in+error.%22&sxsrf=ALeKk025PHewu8FVDfQUS4qNvKP26IhULQ:1589302256799&filter=0&biw=1210&bih=754

On a tangent I do wonder what the right way should be to be in a world where you can have fallback error pages and maintain the ability to override them, particularly if we start moving to a world where we do much less nginx configuration (as we already live in with govuk-docker). I guess we need some static error pages to handle cases where routes don't proxy to any app, so they may have to exist in some form. Then app wise we've got a situation where we're configuring individual error pages for each publishing app (that we might want to simplify for re-use) and thus may want to use the same mechanism in frontend apps which would bring it within Rails control for a less rigid approach.

benthorner:
@kevindew so my current plan is to proceed with the work we're doing on https://github.com/alphagov/static/pull/2167, so that we remove a second layer of error handling on GOV.UK. I think this is non-contentious, as it'll be a net improvement in the quality of the error pages seen by the user.

We then have a separate question of whether to revisit this specific issue with 422 response codes, and whether we should remove the fallback. This would mean:

- Coping a generic 422 template into not-sure-how-many apps.
- Adding support for 422 into Slimmer (or maybe just any code)

For me, the first point is a smell. To try and resolve it, perhaps we can introduce this as functionality of GovukPublishingComponents, with consuming apps doing something like:

```
# config/routes.rb
mount GovukPublishingComponents::Errors, at: "/"

# config/application.rb
config.exceptions_app = routes
```

Although this would be beneficial if GOV.UK becomes a more interactive site, it's hard to see the benefits in the medium term. I can add a card for us to _try_ and do this, but we may need to abandon the idea if it starts blowing up, as it's increasingly tangential to our original aim. If we have to abandon it, we could extend the docs we're trying to write, to indicate the future intent.

How does that sound?

kevindew:
Yup, the mounting of an error controller from govuk_publishing_components sounds like a good approach. It'd be nice to have the configuration of Content Publisher easier to drop into apps for error handling and having frontend and backend consistent. Do agree it could blow up in complexity though. 

I end up rather conflicted around this as I do appreciate the convenience and ubiquity of the nginx configuration, but I do feel it encourages dubious HTTP patterns, annoying work arounds and surprising effects. 

Anyway thanks for carrying on the conversation about this 🙂

PR #2162, opened 2020-05-07
Update how we handle styles for IE8

andysellick:
## TL;DR

We don't officially support IE8 or below anymore but we still have some legacy styles in place and some users on these browsers. This change removes some old code, improves how we handle old IE stylesheet compilation and fixes some major layout problems in IE8.

## Detail

Our current IE specific style strategy is as follows:

- have `@include` lines throughout our Sass for specific IE styling
- compile all of the Sass once for modern browsers
- compile all of the Sass again (with the IE specific stuff included) for old IE
- (actually do the previous two steps twice, once for `static`/`static-ie8` and once for `core-layout`/`core-layout-ie8`)
- use conditional comments to include the first CSS for modern or the second CSS for old IE

We don't support IE8 or below anymore and the number of IE specific lines is quite small and scattered. One strategy would be to simply remove all of this code, however I think at last accurate count we had something like 40,000 visitors to GOV.UK using these older browsers, and it seems wrong to simply shut the door in their faces.

Instead, I've tried to remove only those styles that aren't needed (due to modernisation of our CSS in this and other applications) and actually improve things a little for old IE. Specifically:

- remove unneeded old IE `@include` lines (details below)
- move all remaining lines into an IE specific Sass include
- change our compilation strategy and use of conditional comments: compile all the normal styles once and include them regardless of browser, then compile all the IE specific ones as a separate stylesheet and only include that if viewed on IE8 or below

Additionally, the new grid layout from `govuk-frontend` collapses into a single (left aligned) grid column on IE8 because it's mobile first and uses media queries that old IE doesn't understand to do desktop column layout. I've added some reinforcing rules into the IE specific Sass file to help fix this. This means that IE8 users will see the grid layout properly. A version of [these styles](https://github.com/alphagov/frontend/blob/master/app/assets/stylesheets/application-ie8.scss) was already in `frontend` to do the same thing, but putting them here means they should be available to all pages on GOV.UK (I'll raise a separate PR to remove them from `frontend` shortly).

## Screenshots

Screenshots below show the layout improvement for IE8. What they don't show clearly is the new version is now centred on the page (as it should), unlike the old version, which collapsed everything against the left side. See below for a better view of this.

There are also some other obvious problems on this page, but they're not in the scope of this change.

Before | After
--- | ---
old-ie3 | old-ie2

old-ie4

## IE specific changes

The following is a list of the IE specific rules that have been moved or removed as part of this change. For any rule that applies to an element but has been removed, this has been done only where doing so causes only minimal visual change and does not affect the ability of the user to use or interpret the element in question.

`.page-header div h1 span` (removed)

This class occurs in the following applications:

- collections, but no elements match this specific rule and all handled by components for heading in page-header anyway
- frontend, but no elements match this and all handled by components anyway
- licence-finder, but no elements match this
- manuals-publisher, no elements match
- publisher, no elements match
- specialist-publisher, no elements match
- static, no elements match
- whitehall, no elements match

`.cookie-container` (removed)

- not in use anywhere

`#global-header form#search input` (removed)

- in static only, applies to the search input and input button in the main header search, but removing this only makes the elements slightly taller

`%site-width-container` (moved)

- mixin applied to various elements notably the #wrapper element, used throughout govuk. IE8 rule adds an explicit width alongside the max-width already defined, not sure why as IE8 supports max-width, but regardless, the page is not centered in IE8 anyway, so the whole point of this isn’t working. Have replaced with some specific rules in the IE8 stylesheet for the elements that used this to try to fix it, will look at it again later.

Things in _text.scss...

`.content-block header h1`,
`article header h1` (removed)

- applied zoom to a h1 for IE8 and below. Used in frontend for .content-block class but heavy nesting and ambiguous element name (article) makes finding instances of this very difficult, decided to remove

`.content-block .mainstream & .summary`,
`article .mainstream & .summary` (removed)

- see previous note

`.content-block ol.steps > li`,
`article ol.steps > li` (removed)

- see previous note

`.govuk-sticky-element` (moved)


app/assets/stylesheets/helpers/_old-ie.scss
```diff
@@ -0,0 +1,42 @@
...
+#wrapper,
+.survey-wrapper,
+.govuk-emergency-banner div {
+  width: 960px;
+  margin-left: auto;
+  margin-right: auto;
+}
+
+.js-enabled .govuk-sticky-element--hidden {
+  // XXX: Would be nice to `@include .visuallyhidden`.
```
injms, line 11: I think this note needs a `TODO` or a `FIXME` - I don't understand what `XXX` means.

app/assets/stylesheets/helpers/_old-ie.scss
```diff
@@ -0,0 +1,42 @@
...
+  position: absolute;
+  left: -9999em;
+}
+
+// govuk-frontend style reinforcing
+// needs to be redefined for old IE as govuk-frontend media queries confuse it
+// hence also the additional nesting in the body element
```
injms, line 18:
```suggestion
// Needs to be redefined for old IE as govuk-frontend media queries confuse it,
// hence also the additional nesting in the body element.
```

app/views/root/_stylesheet.html.erb
```diff
@@ -1,7 +1,7 @@
 <%
   css_file = local_assigns[:css_file] || 'static'
 %>
-<!--[if gt IE 8]><!--><%= stylesheet_link_tag css_file, integrity: false %><!--<![endif]-->
-<!--[if IE 8]><%= stylesheet_link_tag "#{css_file}-ie8" %><script>var ieVersion = 8;</script><![endif]-->
+<%= stylesheet_link_tag css_file, integrity: false %>
+<!--[if lt IE 9]><%= stylesheet_link_tag "#{css_file}-ie8" %><script>var ieVersion = 8;</script><![endif]-->
```
injms, line 5: Saying that if _less than_ IE 9 then `ieVersion` _is_ 8 feels slightly strange - it's the truth from a certain point of view.

What do we gain from setting this to be less than IE 9 rather than equal to IE 8?

(My other thought was digging through where this global variable is used and seeing if we can remove it - but that's a topic for another pull request.)

app/assets/stylesheets/helpers/_old-ie.scss
```diff
@@ -0,0 +1,42 @@
...
+    width: 33.3333%;
+    padding: 0 15px;
+  }
+
+  .govuk-grid-column-two-thirds {
+    float: left;
+    width: 66.6666%;
+    padding: 0 15px;
+  }
+}
```
injms, line 42: I'd suggest that this code for frontend should be moved to the frontend app to stop every app needing to download and parse these styles. 

Moving it to [frontend's application-ie-8.scss](https://github.com/alphagov/frontend/blob/master/app/assets/stylesheets/application-ie8.scss) could also remove some duplication, as there is overlap between this and frontend's IE8 specific stylesheet. 

For example, frontend's IE8 stylesheet has this as well:
```scss
.govuk-grid-column-two-thirds {
  float: left;
  width: 66.6666%;
  padding: 0 15px;
}
```

app/assets/stylesheets/helpers/_old-ie.scss
andysellick: Me neither! I moved this line of code from somewhere else without modifying it. Probably should be more descriptive. Will change.

app/views/root/_stylesheet.html.erb
andysellick: Sorry, this was me being lazy - it should be `less than or equal to IE8` but I tried what I thought the syntax for that was and it didn't work, but the same file already had `less than IE9` in it, so I used that instead. I'll see if I can make it more explicit.

app/assets/stylesheets/helpers/_old-ie.scss
andysellick: 😄 agreed. This whole PR actually came about partly because I'd already written these styles in `frontend` and then realised if we need them there we need them everywhere. Will raise that PR once I've finished with this one 👍

andysellick: Also massive thanks for the review @injms !!

app/assets/stylesheets/helpers/_old-ie.scss
andysellick: Turns out thanks to the changes in the gem we can now do what this comment aspired to - include a visually hidden mixin from `govuk-frontend`. Hooray! 😄

app/views/root/_stylesheet.html.erb
andysellick: Fixed.

PR #2173, opened 2020-05-15
Change job link in humans.txt

bilbof:
Will change https://www.gov.uk/humans.txt

### DIFF
--- a/public/humans.txt
+++ b/public/humans.txt
@@ -1,1 +1,1 @@
-GOV.UK is built by a team at the Government Digital Service in London. If you'd like to join us, see https://jobs.jobvite.com/gds.
+GOV.UK is built by a team at the Government Digital Service in London (with thanks to Martha!). If you'd like to join us, see https://jobs.jobvite.com/gds.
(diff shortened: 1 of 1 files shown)


bilbof:
The route is here: https://github.com/alphagov/static/blob/c8346267ecbf29ac9c8b1152ce2e8afaf1976e37/lib/tasks/publishing_api.rake#L8-L14

chao-xian:
Where's the "Thanks Martha"?

issyl0:
@chao-xian Also here: https://github.com/alphagov/static/blob/6a0d8a619e56063c81c193458bff92618f163306/app/views/root/homepage.html.erb#L13.

bilbof:
@chao-xian suggest an edit! :-)

public/humans.txt
```diff
@@ -1 +1 @@
-GOV.UK is built by a team at the Government Digital Service in London. If you'd like to join us, see https://gds.blog.gov.uk/jobs.
+GOV.UK is built by a team at the Government Digital Service in London. If you'd like to join us, see https://jobs.jobvite.com/gds.
```
chao-xian, line 1:
```suggestion
GOV.UK is built by a team at the Government Digital Service in London (with thanks to Martha!). If you'd like to join us, see https://jobs.jobvite.com/gds.
```

PR #2175, opened 2020-05-15
Adding google analytics scroll depth tracking

zilnhoj:
Adding the tracking to the new guidance/answers-to-the-most-common-topics-asked-about-by-the-public-for-the-coronavirus-press-conference page 

Added at the end of the config variable and set to percentage of scrolling in 20% increments


kevindew:
Thanks @zilnhoj - I've just switched this over to https://github.com/alphagov/static/pull/2176 to get it merged.

PR #2186, opened 2020-05-19
git mv .github/workflows/tests.yml .github/workflows/ci.yml

issyl0:
- According to the style guide. And so that the repo auto-configurer can detect GitHub Actions.


issyl0:
A single non-Jenkins CI config file will not in any way affect the e2e tests, so I'm skipping them.

PR #2190, opened 2020-05-21
Update global banner content

andysellick:
Changes the global banner from this:

Screenshot 2020-05-21 at 09 14 55

to this:

Screenshot 2020-05-21 at 09 15 03

Specifically, it replaces the custom styles for this link with the new action link component from the gem (this is also used in collections and will make any future maintenance easier).

Trello card: https://trello.com/c/l9xuuuTB/279-update-emergency-banner-design

---

There's a lot of unused code related to this change. Some of it is specific to the coronavirus work, so I have confidence that we don't need it anymore, but some of it is more generic, such as the styles and JS for handling the hide/dismiss option for the global banner (also not now in use). I had two choices:

- **leave it**: we _might_ need it again in the future, but we don't want code sitting around that isn't needed (or JS being downloaded by users that isn't needed)
- **remove it**: I've chosen this option but split the commits up as much as possible in case we decide to keep some things

I realise this change may be controversial so I'm happy to discuss it. Also please let me know if there's any other code related to this that I've overlooked.


vanitabarrett:
I'm a bit torn on this change - on one hand it makes perfect sense and it's nice to neaten up the code, but on the other hand it gets rid of some (at the moment unused) functionality which might be needed in the future and missed if implementing again from scratch (this PR probably wouldn't be quick to revert). 

I'd be tempted to use this PR to try and implement a simpler banner, but one that still had flexibility for the future (e.g: title without a link, description text, dismiss button).

If we decide to go down the route of making the banner specific to its current use case each time (coronavirus at the moment), I think we either need to [update the docs each time](https://github.com/alphagov/static/blob/master/doc/global-banner.md) or make a decision on whether this is something 2ndline is even responsible for or not.

andysellick:
@vanitabarrett have had a rethink and am now closing this PR in favour of https://github.com/alphagov/static/pull/2192

PR #2192, opened 2020-05-21
Update global banner with action link component

andysellick:
This PR is a less drastic alternative to https://github.com/alphagov/static/pull/2190. It does the following things:

- updates the global banner to use the action link component for the coronavirus link, and adds the updated sub text to it
- preserves the existing banner behaviour by separating out the existing banner text from the coronavirus link (means that 2ndline can still switch on an unrelated global message alongside the coronavirus link if needed, or hide the coronavirus link)
- adds some styles for the non-covid message (links were white, but plain text wasn't, also adds a bit of spacing)
- removes most of the content from the now redundant additional banner, which is not in use, we will need to address the code relating to this at some point (mainly JS)

## Visual changes

Banner before:

Screenshot 2020-05-21 at 11 20 54

Banner after:

Screenshot 2020-05-21 at 11 09 33

Banner with both covid link and regular text:

Screenshot 2020-05-21 at 11 16 31

Banner with covid link and regular link:

Screenshot 2020-05-21 at 11 16 50

Banner with covid link, regular link and regular text:

Screenshot 2020-05-21 at 11 17 16

Banner with covid link, regular link and regular text link:

Screenshot 2020-05-21 at 11 17 38

Trello card: https://trello.com/c/l9xuuuTB/279-update-emergency-banner-design


app/assets/stylesheets/header-footer-only.scss
```diff
@@ -3,6 +3,7 @@ $govuk-use-legacy-palette: false;
 
 @import "govuk_publishing_components/govuk_frontend_support";
 @import "govuk_publishing_components/components/_cookie-banner";
+@import "govuk_publishing_components/components/_action-link";
```
vanitabarrett, line 6: Do we need to add this to the error page css too?

app/assets/stylesheets/helpers/_header.scss
```diff
@@ -510,40 +510,26 @@ $covid-grey: #262828;
   font-weight: 700;
   margin-right: govuk-spacing(2);
   margin-bottom: govuk-spacing(1);
+  color: govuk-colour("white");
```
vanitabarrett, line 513: It looks like we're setting the color twice for `global-bar-title`? (once here, and one a few lines below)

andysellick: 🤦 thanks, fixed.

app/assets/stylesheets/header-footer-only.scss
andysellick: Yes, have added. Also noticed that `static.scss` imports this file, so have removed some duplication from it. I've done this as an extra commit, can you sanity check please?

PR #2197, opened 2020-05-22
Iterate design and copy for error pages

benthorner:
https://trello.com/c/gl0JJDev/296-iterate-copy-and-design-for-the-fallback-govuk-error-pages

This updates the copy on the most of the error pages, following a
content review. We also make a few other changes:

- Move and re-style the "Status code" message
- Synchronise the browser title with the page heading

Previously we had a mixture of browser titles being in-sync (or not)
with the page heading, for no apparent reason. The convention on
GOV.UK is generally to use the same text for both.

## Before

Screenshot 2020-05-22 at 17 20 32

## After

Screenshot 2020-05-22 at 17 20 21

### DIFF
--- a/app/views/root/400.html.erb
+++ b/app/views/root/400.html.erb
@@ -1,6 +1,5 @@
 <%= render partial: "error_page", locals: {
-    title: "Bad request",
-    heading: "Bad request",
-    intro: '',
+    heading: "Sorry, there is a problem",
+    intro: '<p>Go back and change the information you entered.</p>',
     status_code: 400
 } %>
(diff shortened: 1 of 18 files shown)


app/views/root/400.html.erb
```diff
@@ -1,6 +1,5 @@
 <%= render partial: "error_page", locals: {
-    title: "Bad request",
-    heading: "Bad request",
-    intro: '',
+    heading: "Sorry, there is a problem",
+    intro: '<p>Go back and change the information you entered.</p>',
```
andysellick, line 3: Can you add a class of `govuk-body` to this and the other `<p>` tags passed to this partial? It probably looks fine now but we're going to be removing a lot of the default styling in static fairly soon.

benthorner: That's a good point. Considering there are a few other items on this page - such as the title - I'm worried this could lead to scope creep. Since it sounds like we're going to audit the whole app anyway (to remove the default styling), could we leave it until then?

andysellick: That's fine, as long as it gets looked at at some point.

PR #2200, opened 2020-05-26
Add comment to sass

andysellick:
Static has the Sass for the action link component because we're using it in the global banner on every page of GOV.UK. Collections is also using the action link component. Rather than duplicate the CSS, collections is relying on the copy in static. 

This PR adds a comment to the Sass to explain that to prevent someone from removing the Sass from static without putting it into collections first.

Should have been part of https://github.com/alphagov/static/pull/2192. See also https://github.com/alphagov/collections/pull/1754

### DIFF
--- a/app/assets/stylesheets/header-footer-only.scss
+++ b/app/assets/stylesheets/header-footer-only.scss
@@ -3,6 +3,8 @@ $govuk-use-legacy-palette: false;
 
 @import "govuk_publishing_components/govuk_frontend_support";
 @import "govuk_publishing_components/components/_cookie-banner";
+// collections uses the action-link component and relies upon this line, see
+// https://github.com/alphagov/collections/pull/1754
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/header-footer-only.scss
```diff
@@ -3,6 +3,8 @@ $govuk-use-legacy-palette: false;
 
 @import "govuk_publishing_components/govuk_frontend_support";
 @import "govuk_publishing_components/components/_cookie-banner";
+// collections uses the action-link component and relies upon this line, see
+// https://github.com/alphagov/collections/pull/1754
```
reggieb, line 7: I'd suggest adding a space between the comment and the line it does not apply to (I think the line above) as that would make it clearer that the comment applies to the line below and not the line above.

PR #2205, opened 2020-05-28
Call cookie banner component JS

andysellick:
Calls the cookie banner component JS wherever the JS in static references the related JS in the gem's lib dir.

### DIFF
--- a/app/assets/javascripts/start-modules.js
+++ b/app/assets/javascripts/start-modules.js
@@ -1,4 +1,5 @@
 // = require govuk_publishing_components/modules
+// = require govuk_publishing_components/components/cookie-banner
 // = require modules/global-bar
 // = require modules/sticky-element-container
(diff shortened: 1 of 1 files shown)


andysellick:
@alex-ju okay, will change. Does that mean I can move all of the instances of `//= require govuk_publishing_components/lib/cookie-functions` into that file as well?

alex-ju:
> @alex-ju okay, will change. Does that mean I can move all of the instances of `//= require govuk_publishing_components/lib/cookie-functions` into that file as well?

yes, the utility functions need to be imported wherever needed, otherwise they would be 'undeclared' dependencies or they would rely on script order. can be improved for sure, but I wouldn't touch them in this.

andysellick:
@alex-ju @kevindew thanks both, have updated.

PR #2207, opened 2020-06-01
Migrate auto track event from toolkit

alex-ju:
After [migrating `frontend` from `wrapper` to `core_layout`](https://github.com/alphagov/frontend/pull/2154) we ended up in a position where `auto-track-event` was missing from our analytics bundle in `static`, causing [a gap in our GA data](https://govuk.zendesk.com/agent/tickets/4096617).

To fix this I migrated the `auto-track-event` from `govuk_frontend_toolkit` to `static`, as this was [planned work in the GOV.UK Frontend and Accessibility backlog](https://trello.com/c/TMYS2FKT/222-migrate-javascript-from-toolkit-into-static) anyway, and added it to the common analytics bundle in static making it available to `frontend` again.

You can test this by checking for `GOVUK.Modules.AutoTrackEvent` in the browser console in `frontend`.

### DIFF
--- a/app/assets/javascripts/analytics.js
+++ b/app/assets/javascripts/analytics.js
@@ -7,6 +7,7 @@
 //= require analytics_toolkit/mailto-link-tracker
 //= require analytics_toolkit/external-link-tracker
 //= require analytics_toolkit/download-link-tracker
+//= require analytics_toolkit/auto-track-event
 
 //= require analytics/page-content
 //= require analytics/custom-dimensions
--- a/app/assets/javascripts/analytics_toolkit/auto-track-event.js
+++ b/app/assets/javascripts/analytics_toolkit/auto-track-event.js
@@ -1,0 +1,30 @@
+;(function (global) {
+  'use strict'
+
+  var GOVUK = global.GOVUK || {}
+  GOVUK.Modules = GOVUK.Modules || {}
+
+  GOVUK.Modules.AutoTrackEvent = function () {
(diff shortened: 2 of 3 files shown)


alex-ju:
> The changes look sensible, but I'm not sure I fully understand this. `GOVUK.Modules.AutoTrackEvent` seems to still work on live, presumably because static is importing that code from toolkit still? So why is it not working?

Assuming you tested this on the homepage – it seems that the homepage is still using `application.js` from static, whereas all the rest of the pages served by `frontend` are using `header-footer-only.js` (you can check https://www.gov.uk/school-term-holiday-dates/ – which is the reported page, or https://www.gov.uk/bank-holidays which is easier to test with on a dev environment).

> And does it need to be removed from toolkit now?

[The whole JavaScript from toolkit is still being imported](https://github.com/alphagov/static/blob/master/app/assets/javascripts/application.js#L4) (and potentially used) by apps using `application.js` from static (`frontend` and `licence-finder`, for example), so I wouldn't remove it until we fully migrate them to use `core_layout`/`header-footer-only.js` as I don't know the full implications of removing them. When we are confident enough, I would rather remove the dependency on toolkit instead of removing various scripts from toolkit.

> Can you provide more explanation?

I hope the above provides a bit more insight, but do let me know if there are things that I might be able to explain.
 
> (should the first commit be titled 'Migrate auto-track-event from toolkit', not static?)

Yes, late-evening brain 😴, will update now.

PR #2230, opened 2020-06-26
Disable additional banner section

sihugh:
My thought is to leave the markup section as is, as this will make it _much_ easier for messages to go up at short notice.


sihugh:
Closing in favour of https://github.com/alphagov/static/pull/2229

PR #2238, opened 2020-07-17
Remove dependency on govuk_frontend_toolkit typography

danacotoran:
Remove dependency to govuk_frontend_toolkit typography file.

Replace deprecated typography mixins with newer govuk frontend ones, as
per the Design System Guidance:
https://design-system.service.gov.uk/get-started/updating-your-code/

https://trello.com/c/RlJmOIOu


app/assets/stylesheets/helpers/_typography.scss
```diff
@@ -0,0 +1,5 @@
+// // New Transport Light
+$nta-light: "nta", Arial, sans-serif;
+$nta-light-tabular: "ntatabularnumbers", $nta-light;
+// // Helvetica Regular
+$helvetica-regular: "HelveticaNeue", "Helvetica Neue", "Arial", "Helvetica", sans-serif;
```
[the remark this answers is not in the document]
danacotoran: Yep, they are indeed!

app/assets/stylesheets/helpers/_core.scss
```diff
@@ -86,7 +86,7 @@ header.page-header div {
 
 /* Meta-data */
 .meta-data {
```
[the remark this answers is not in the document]
danacotoran: As discussed I'm not comfortable doing this as part of this PR as it doesn't quite fit with the scope and increases the testing effort required. I've made a head start on this on another branch.

app/assets/stylesheets/helpers/_core-print.scss
```diff
@@ -7,14 +7,14 @@ main > header {
   position: relative;
 
   h1 {
-    @include bold-24;
+    @include govuk-font(24, $weight: bold);
```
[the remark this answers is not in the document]
danacotoran: Hi @andysellick as discussed, while your suggestions are sensible as always, team consensus is that we should not introduce these types of changes as part of this piece of work. 

Firstly, because it does not fit the scope of this PR.
Secondly, because although removing/refactoring those styles seems like an easy task on the surface,  there is potentially a large amount of work in ensuring we don't break anything in the process (searching all the applications for the thing, removing the dependency for the thing before removing it from `static` – implementing this can be time consuming if _the thing_ is actually many classes/selectors).

Following on from this I've raised [an issue](https://github.com/alphagov/static/issues/2239) to document the points you've raised.

app/assets/stylesheets/helpers/_core.scss
```diff
@@ -8,27 +8,27 @@ body {
...
+  @include govuk-font(19);
 }
 
 h2, h3, h4 {
   font-weight: 400;
 }
 
 h2 {
-  @include core-24;
+  @include govuk-font(24);
```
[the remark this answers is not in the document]
danacotoran: Similar to what I've said [here](https://github.com/alphagov/static/pull/2238/files#r457936107), and as you've noted yourself, this is a lot of work and is not necessarily directly related with the typography changes in this PR (although admittedly there is some cross-over). 
I'd like us to get it done, however I don't think it should be done as part of this piece of work.

app/assets/stylesheets/helpers/_multi-step.scss
```diff
@@ -74,7 +77,7 @@
       padding-right: 9.5em;
 
       .answer {
-        @include bold-16;
+        @include govuk-font(16, $weight: bold);;
```
andysellick, line 80: Looks like an extra semi colon on this line.

app/assets/stylesheets/helpers/_print-base.scss
```diff
@@ -34,25 +34,25 @@ h4 {
...
 
 h2 {
-  @include bold-19;
+  @include govuk-font(19, $weight: bold);
   padding-top: 7pt;
 }
 
 h3 {
-  @include bold-16;
+  @include govuk-font(16, $weight: bold);;
```
andysellick, line 42: Extra semi colon.

app/assets/stylesheets/helpers/_print-base.scss
```diff
@@ -34,25 +34,25 @@ h4 {
...
 }
 
 h4 {
-  @include bold-14;
+  @include govuk-font(14, $weight: bold);;
```
andysellick, line 46: Extra semi colon.

app/assets/stylesheets/helpers/_text.scss
```diff
@@ -190,7 +190,7 @@ article {
   }
 
   th, td {
-    @include core-16($tabular-numbers: true);
```
andysellick, line 193: What did `tabular-numbers` do, and is it no longer needed?

app/assets/stylesheets/static.scss
```diff
@@ -1,6 +1,7 @@
 $govuk-compatibility-govuktemplate: true;
 $govuk-use-legacy-palette: false;
 
+@import "govuk_publishing_components/component_support";
```
andysellick, line 4: Why is this being added to this file?

app/assets/stylesheets/header-footer-only.scss
```diff
@@ -12,7 +12,6 @@ $govuk-use-legacy-palette: false;
 @import "css3";
 @import "device-pixels";
 @import "measurements";
-@import "typography";
```
andysellick, line 15: Does this file need `govuk_frontend_support` now that `typography` has been removed?

danacotoran: It already imports `govuk_frontend_support` at line 4 (unless you're implying it _doesn't_ need it?) 🤔

app/assets/stylesheets/helpers/_text.scss
danacotoran: I must have lost it in the process. 🤦🏻‍♀️ 
I believe what it does is request the tabular version of the Transport font.

Have updated to pass an additional `$tabular: true` param to the `govuk-font` mixin (as per [here](https://github.com/alphagov/govuk-frontend/blob/9dc2cfb064fb49e421d5a73be6a241506e1414b8/src/govuk/helpers/_typography.scss#L177))

app/assets/stylesheets/header-footer-only.scss
andysellick: Ah, sorry, missed that, thanks.

app/assets/stylesheets/static.scss
danacotoran: It's because many of the styles imported in this file have been updated to use `govuk-font`  instead of the deprecated `core-x` typography mixins. This imports the govuk mixins via `govuk_publishing_components`

andysellick: Is it in one of the @imports then? Just wanted to check it was needed, as there was no reference to the toolkit typography removed from here.

danacotoran: Oh yeah I see what you mean now. Yes so 4 of the `@import "helpers/something"` in this file used to call the deprecated mixin and are now using `govuk-font` instead.

There's an error if I try to remove the component support import.

andysellick: Ah, got it. Thanks for the explanation.

app/assets/stylesheets/styleguide/_font-stacks.scss
```diff
@@ -32,6 +32,6 @@ $NTA-Light:
        local("Helvetica");
 }
 
-$Helvetica-Regular:
+$helvetica-regular:
```
andysellick, line 35: Looks like this variable was removed from _organisations.scss - is it still needed?

app/assets/stylesheets/styleguide/_font-stacks.scss
```diff
@@ -19,7 +19,7 @@
        local("Helvetica");
 }
 
-$NTA-Light:
+$nta-light:
```
andysellick, line 22: This is being renamed but I can't see it anywhere else in the PR - is it needed?

app/assets/stylesheets/helpers/_text.scss
```diff
@@ -6,20 +6,20 @@
...
   h1 {
-    @include bold-24;
+    @include govuk-font(24, $weight: bold);
     margin-top: 1.5em;
     margin-bottom: 0.75em;
   }
 
   header h1 {
-    @include heading-27;
+    @include govuk-font(27, $weight: bold);
```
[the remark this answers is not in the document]
danacotoran: The [Design System How To Guide](https://design-system.service.gov.uk/get-started/updating-your-code/#mixins-and-variables) recommends that `govuk-font(27, $weight: bold)` is used to replace `heading-27`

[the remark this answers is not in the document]
andysellick: That's fine, I was raising this because it was handled one way here but another way here: https://github.com/alphagov/static/pull/2238/files#diff-1508b08fa30f83718afb89507ab05dfaR47

Can you explain the difference or do both these examples need to be consistent?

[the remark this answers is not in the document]
danacotoran: It was handled differently in that other scenario because there is a specific `govuk-heading` size that can replace `heading-48`. As 48px is a heading size we still actively use. 

My understanding is that we no longer officially use 27px headings, which is why the DS guidance recommends replacing it with `govuk-font(27, $weight: bold)` when migrating to the DS.

app/assets/stylesheets/helpers/_core.scss
```diff
@@ -44,17 +44,17 @@ header.page-header div {
   }
 
   h1 {
-    @include heading-48;
+    @extend %govuk-heading-xl;
     background-repeat: no-repeat;
```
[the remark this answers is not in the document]
danacotoran: that line was added 6 years ago... oh dear 
 https://github.com/alphagov/static/commit/6db12fce5ae9e0a11c483db474787883d1cae285

There is some really old stuff in this file (_9 years ago_ 😱).

app/assets/stylesheets/styleguide/_font-stacks.scss
danacotoran: I've removed this file altogether. Could not find a reference to it anywhere (I don't think it's actually been used in a long while?). The most recent changes to this file were from 8 years ago 😬

app/assets/stylesheets/styleguide/_font-stacks.scss
danacotoran: Nope, I think I renamed it in a find and replace.

app/assets/stylesheets/helpers/_text.scss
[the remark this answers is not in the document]
andysellick: Great, thanks for the detail.

app/assets/stylesheets/styleguide/_font-stacks.scss
andysellick: Good shout. Although it's worth pointing out that even if something's not been updated for a while it doesn't mean we're not still depending on it! Hence my earlier enthusiasm for removing much of this stuff...

PR #2248, opened 2020-08-04
[DO NOT MERGE] Spike using GA without cookies

andysellick:
Don't merge this, it's just an experiment.

- users who have not accepted cookies will still fire a GA pageview, but no cookies will be set
- when those users move to another page, another GA pageview, but the user id will be different (ie. no tracking of user movements, only page hits counted)
- users who accept cookies should still have the same behaviour, although as this is secondary to this spike I haven't fully tested this yet
- have had to circumvent some of the cookie consent checking, which is causing a test failure that has been temporarily disabled


andysellick:
Closing this PR as currently not needed. Also all the analytics code has now moved to `govuk_publishing_components`.

PR #2258, opened 2020-08-27
Fix crash on cookie setting

maxgds:
The banner code was trying to manipulate a cookie that doesn't
yet exist. This sets the default if none is present and removes
a redundant if statement.

The bug can be seen by going in an incognito window (so no cookies are set for the site) to http://www.gov.uk/help/cookies - the js crashes and makes it impossible to set cookie preferences which is the purpose of that page. This probably only affects this page as elsewhere the cookie defaults are set in the cookie banner that is excluded from appearing here.


owenatgov (approved):
👌

PR #2271, opened 2020-09-14
Reset humans.txt

richardTowers:
We temporarily updated www.gov.uk/humans.txt to say goodbye to Issy. It shouldn't be a permanent feature though, otherwise it looks like a bit of an overreaction.

This reverts the commits that added the message, but leaves the one that removes the trailing full stop from the first sentence (as that's a small visual improvement we can keep).


andysellick (approved):
😢

PR #2283, opened 2020-09-24
Make mobile menu toggle a button

andysellick:
- was formerly a link, which failed accessibility requirements because it behaved like a button
- uses existing JS for this (header-navigation.js, in govuk_publishing_components for reference, fact fans)
- copied and modified the styling from govuk-template (as it no longer applies) in order to preserve the look of the element exactly, CSS may therefore not be as nice as it could be, but this should all be replaced by the proper header at some point anyway

Screenshot 2020-09-24 at 11 11 26

Should be no visual changes.

Trello card: https://trello.com/c/2bL11sIv/414-mobile-menu-button-looks-like-a-link

---

Interesting note - this code for the menu toggle is in `static` (obviously). The JS that controls it (and also the search toggle for mobile, just above it) is in `govuk_publishing_components`. The menu that appears when you toggle this button appears to come from `whitehall` (https://github.com/alphagov/whitehall/blob/master/app/views/layouts/frontend.html.erb).

The sooner we get the header consolidated into one component the better.

### DIFF
--- a/app/assets/stylesheets/helpers/_header.scss
+++ b/app/assets/stylesheets/helpers/_header.scss
@@ -269,6 +269,36 @@
     }
   }
 
+  .mobile-menu-toggle {
+    @include govuk-font(16);
+    float: right;
+    padding: 6px 0 4px 0;
+    color: govuk-colour("white");
+    background: none;
+    border: 0;
+
+    @include govuk-media-query($from: tablet) {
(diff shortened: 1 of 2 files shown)


injms (approved):
🧪👨🏻‍🎤🙆🏻‍♂️

PR #2290, opened 2020-10-09
Update site search label to "Search GOV.UK"

danacotoran:
The site search box is labelled inconsistently throughout the site. 
This is a fail of WCAG SC 3.2.4: **Consistent Identification**: Components that have the same functionality within a set of Web pages are identified consistently. (Level AA)

We're changing the site search label to be "Search on GOV.UK" everywhere.

https://trello.com/c/RbD3om74


injms (approved):
Good fix until we can use the component 👍

PR #2293, opened 2020-10-13
Replace hard coded site search with search component

danacotoran:
Use [existing search component](https://components.publishing.service.gov.uk/component-guide/search) to replace the hard coded site search in the header.
This is a step towards eliminating dependency on old styles and towards the "componentisation" of our frontend apps.

Cleaned up a good amount of CSS that is no longer in use thanks to using the component.  

https://trello.com/c/RbD3om74

### DIFF
--- a/app/views/root/_base.html.erb
+++ b/app/views/root/_base.html.erb
@@ -18,7 +18,7 @@
     <div class="content govuk-clearfix">
       <%= render "govuk_publishing_components/components/search", {
         id: "site-search-text",
-        label_text: "Search",
+        label_text: "Search on GOV.UK",
         no_border: true,
         margin_bottom: 0
       } %>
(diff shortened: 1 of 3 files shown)


app/views/root/_base.html.erb
```diff
@@ -16,9 +16,12 @@
      # changing this, as the redirect adds some parameters to the search query. %>
   <form id="search" class="site-search govuk-clearfix" action="/search" method="get" role="search" aria-label="Sitewide">
     <div class="content govuk-clearfix">
-      <label for="site-search-text">Search on GOV.UK</label>
-      <input type="search" name="q" id="site-search-text" title="Search" class="js-search-focus">
-      <input class="submit" type="submit" value="Search" />
+      <%= render "govuk_publishing_components/components/search", {
+        id: "site-search-text",
+        label_text: "Search",
```
injms, line 21: Should this be 'Search on GOV.UK'?

danacotoran: Indeed it should 🤦🏻‍♀️

injms (approved):
🙌

Issue #2301, opened 2020-10-17
"Search on GOV.UK" label text is cut off on mobile (portrait) view.


issyl0:
Hello friends!

On my OnePlus 7 Pro (Google Chrome browser), I just noticed that the "Search on GOV.UK" search box label text (changed in #2290 as part of an accessibility review) is _slightly_ cut off in portrait view.

Thanks! 💖

Here's a screenshot, to help visualise.

![Screenshot_20201017-202747](https://user-images.githubusercontent.com/355033/96352164-1dce8500-10b9-11eb-9aba-28c61e214007.jpg)

issyl0:
Oooh, I just realised that #2293 might fix this as it resets the label text to just "Search". 🤔

injms:
A fix for this has been merged in, and will be deployed as soon as possible. Thank you @issyl0! 🙌

PR #2305, opened 2020-10-22
Fix search label cut off text

danacotoran:
The global site header search label text has been updated from **"Search"** to **"Search on GOV.UK"**.

This has lead to an issue where the label text is now appearing clipped on larger mobile adaptations as there simply is not enough space to accommodate it.
### Problem
Screenshot 2020-10-22 at 10 05 56

This updates the CSS so that the search field is hidden on larger mobiles – same as it currently is for smaller mobiles. The search form can be shown/hidden using the search toggle button.

### Solution
Screenshot 2020-10-22 at 10 10 19

----- 

Thanks @issyl0 for bringing this to our attention 👉🏼 👉🏼 👉🏼  #2301 

https://trello.com/c/RbD3om74

### DIFF
--- a/app/assets/stylesheets/helpers/_header.scss
+++ b/app/assets/stylesheets/helpers/_header.scss
@@ -9,6 +9,7 @@
   $govuk-header-link-hover: govuk-colour("white");
   $govuk-header-link-active: #1d8feb;
   $govuk-header-nav-item-border-color: #2e3133;
+  position: relative;
 
   #logo {
     background: none;
@@ -104,7 +105,10 @@
   .search-toggle {
     display: none;
 
-    @media screen and (max-width: 379px) {
+    @include govuk-media-query($until: tablet) {
+      position: absolute;
+      right: 0;
+      top: 0;
       background-color: $govuk-brand-colour;
       background-image: image-url("search-button.png");
       background-position: 0 50%;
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/helpers/_header.scss
```diff
@@ -104,15 +105,18 @@
   .search-toggle {
```
injms, line 105: What do you think about making the styles for `.search-toggle` mobile first so the `display: none` would kick in `($from: tablet)` instead of being overwritten?

danacotoran: Hi @injms that's a good point – @chris-gds has actually mentioned this too 👍. 

My rationale here is that the aim of this piece of work is to fix the issue by replacing the old media query style with the `govuk-media-query` equivalent.

This [other PR]( https://github.com/alphagov/static/pull/2293) removes a whole bunch of CSS and refactors the code so that it is mobile first, so I simply didn't want to repeat that on here in the interest of minimising potential conflicts/repetition (but I suppose there will be some conflicts either way, so I don't know... let me know what you think)

injms:
In Firefox, I'm seeing a slight overlap of the focus state:

image

danacotoran:
Closing this PR as https://github.com/alphagov/static/pull/2293 has been approved and also fixes the issue.

PR #2310, opened 2020-10-28
Add main navigation for accounts

alex-ju:
This snappy change adds main navigation for accounts to be used via [slimmer config](See https://github.com/alphagov/slimmer/pull/255).

Note this menu is not compatible with the search input. The search needs to be disabled when navigation is enabled otherwise they will visually clash.

This was tested locally in `finder-frontend` – since this is the primary usage – and will work for both `core_layout` and `header_footer_only` templates in static.

<table>
<tr><th>Signed in</th><th>Signed out</th></tr>
<tr><td valign="top">

signed-in-desktop

</td><td valign="top">

signed-out-desktop

</td></tr>

<tr><td valign="top">

signed-in-mobile

</td><td valign="top">

signed-out-mobile

</td></tr>
</table>

[Trello card](https://trello.com/c/vdZrgBRy)

### DIFF
--- a/app/views/root/_base.html.erb
+++ b/app/views/root/_base.html.erb
@@ -28,7 +28,7 @@
   <div id="accounts-signed-out" class="govuk-header__content gem-c-header__content" data-module="govuk-header">
     <button role="button" class="govuk-header__menu-button gem-c-header__menu-button govuk-js-header-toggle" aria-controls="navigation" aria-label="Show or hide Top Level Navigation">Menu</button>
     <nav class="gem-c-header__nav">
-      <ul id="navigation" class="govuk-header__navigation govuk-header__navigation--end" aria-label="Top Level Navigation">
+      <ul class="govuk-header__navigation govuk-header__navigation--end" aria-label="Top Level Navigation">
         <li class="govuk-header__navigation-item">
           <a class="govuk-header__link" href="<% Plek.new.website_root %>/transition-check/login">Sign in</a>
         </li>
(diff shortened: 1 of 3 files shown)


barrucadu (approved):
I think the e2e test failure is spurious, I've triggered a rebuild

app/views/root/_base.html.erb
```diff
@@ -24,6 +24,31 @@
...
+
+  <div id="accounts-signed-in" class="govuk-header__content gem-c-header__content" data-module="govuk-header">
+    <button role="button" class="govuk-header__menu-button gem-c-header__menu-button govuk-js-header-toggle" aria-controls="navigation" aria-label="Show or hide Top Level Navigation">Menu</button>
+    <nav class="gem-c-header__nav">
+      <ul id="navigation" class="govuk-header__navigation govuk-header__navigation--end" aria-label="Top Level Navigation">
+        <li class="govuk-header__navigation-item">
+          <a class="govuk-header__link" href="https://www.account.publishing.service.gov.uk/">Account</a>
+        </li>
+        <li class="govuk-header__navigation-item">
+          <a class="govuk-header__link" href="https://www.gov.uk/transition-check/logout">Sign out</a>
```
barrucadu, line 47: I've just had a thought; can we make these links different on staging?  www.gov.uk work for that

barrucadu: On staging we'd ideally have www.staging.publishing.service.gov.uk instead of www.gov.uk links and www.account.staging.publishing.service.gov.uk instead of www.account.publishing.service.gov.uk

barrucadu (approved):

PR #2320, opened 2020-11-04
Fix mobile menu for accounts

andysellick:
(hopefully) fixes the mobile menu for accounts in the header.

- the UL that is supposed to appear when the button is clicked should have an ID matching the aria-controls attribute of the button, for the JS to work
- I've used two different IDs just in case slimmer has a problem with duplicates, although some kind of magic seems to be occurring here to make only one chunk of this markup render at any one time

Screenshot 2020-11-04 at 10 39 20

Trello card: https://trello.com/c/ICDC0esK/368-accounts-snag-list-%F0%9F%8C%AD


barrucadu (approved):
Works on my machine :+1:

PR #2321, opened 2020-11-04
Add accounts domain to cross domain linker

andysellick:
So we get cross domain tracking if users have consented.

### DIFF
--- a/app/assets/javascripts/analytics/init.js.erb
+++ b/app/assets/javascripts/analytics/init.js.erb
@@ -43,6 +43,7 @@ var analyticsInit = function() {
       'access.service.gov.uk',
       'access.tax.service.gov.uk',
       'accounts.manage-apprenticeships.service.gov.uk',
+      'account.publishing.service.gov.uk',
       'add-driving-licence-check-code.service.gov.uk',
       'analyse-school-performance.service.gov.uk',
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/analytics/init.js.erb
```diff
@@ -43,6 +43,7 @@ var analyticsInit = function() {
       'access.service.gov.uk',
       'access.tax.service.gov.uk',
       'accounts.manage-apprenticeships.service.gov.uk',
+      'account.publishing.service.gov.uk',
```
barrucadu, line 46: The account manager is on www.account.publishing.service.gov.uk

barrucadu: Also this won't be sufficient for tracking sign ins, as we'll need to figure out how to get the _ga parameter passed through some redirects

andysellick: Yes. GA automatically applies subdomains based on these strings. So for example, if we just put `gov.uk` in, it would match everything in this list (but we don't, for good reasons).

andysellick: Let's have a separate conversation about redirects then, but the change in this PR still needs doing.

barrucadu (approved):

PR #2324, opened 2020-11-10
fix

fredericfran-gds:


fredericfran-gds:
This is no longer needed as this is fixed by: https://github.com/alphagov/static/pull/2327

PR #2327, opened 2020-11-12
Update Dockerfile to reflect it should only run in RAILS_ENV=production

kevindew:
This Dockerfile is only used in RAILS_ENV=production environments and doesn't have the system dependencies installed for other environments (and shouldn't have these unless they're used and thus tested). This change reflects this by changing the default RAILS_ENV to production and installing only the production gems.


fredericfran-gds (approved):
Thank you for fixing this for us

PR #2334, opened 2020-11-27
Bye bye Jon

sihugh:
Thanks to Keith and Joe for writing such lovely words about our friend.

### DIFF
--- a/public/humans.txt
+++ b/public/humans.txt
@@ -1,3 +1,13 @@
 GOV.UK is built by a team at the Government Digital Service (with thanks to Martha!) in London,
 Bristol and Manchester.  If you'd like to join us, see https://jobs.jobvite.com/gds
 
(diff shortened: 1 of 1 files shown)


alex-ju (approved):

selfthinker:
When this gets reverted, can we also update the job URL to point to https://gdscareers.gov.uk/?

alex-ju:
It seems like Steve already raised [a separate one](https://github.com/alphagov/static/pull/2335) to update the URL.

stevenjmesser:
> It seems like Steve already raised [a separate one](https://github.com/alphagov/static/pull/2335) to update the URL.

@selfthinker raises a good point on #2335 that linking to the campaign site would be better than the applications site. The former is owned by GDS whereas the latter is a SaaS product (with a strange convention for URLs IMO).

PR #2355, opened 2020-12-11
Prepare static for a multilingual eubusiness page

alex-ju:
- **Update the global-bar blocklist**
Update the `eubusiness` regex in the `global-bar-init.js` regex blocklist to match translation URLs as we prepare to rollout alphagov/collections#2134. By doing this we prevent the global bar banner from appearing on translation pages of the `/eubusiness` page (for example `/eubusiness.es`)

- **Update eubusiness heading copy**
It seems like the `scroll-tracker.js` script looks for hardcoded content at the bottom of the `/eubusiness` page, content that we update in alphagov/collections#2134.

[Trello card](https://trello.com/c/ZfMI7LNc)

### DIFF
--- a/app/assets/javascripts/analytics/scroll-tracker.js
+++ b/app/assets/javascripts/analytics/scroll-tracker.js
@@ -402,7 +402,7 @@
       ['Percent', 100]
     ],
     '/eubusiness': [
-      ['Heading', 'Additional help and support']
+      ['Heading', 'Webinars for EU-based businesses trading with the UK']
     ]
   };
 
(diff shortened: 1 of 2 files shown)


benthorner (approved):
Looks good to me - interesting to see how our scroll tracking works 👀.

app/assets/javascripts/analytics/scroll-tracker.js
```diff
@@ -402,7 +402,7 @@
       ['Percent', 100]
     ],
     '/eubusiness': [
-      ['Heading', 'Additional help and support']
+      ['Heading', 'Webinars for EU-based businesses trading with the UK']
     ]
```
hannako, line 406: To extend it to the translated pages, do we need to make the scroll tracking based on percentage rather than heading?

alex-ju: For posterity, we're going to switch to a percentage-based measurement in a follow-up PR

PR #2356, opened 2020-12-11
Fix anchor styles in survey script

alex-ju:
Improves the accessibility of these links and makes their style consistent with the rest of GOV.UK.

<table>
<tr><th>Before</th><th>After</th></tr>
<tr><td valign="top">

collections dev gov uk_eubusiness

</td><td valign="top">

collections dev gov uk_eubusiness (1)

</td></tr>
</table>

### DIFF
--- a/app/assets/javascripts/surveys.js
+++ b/app/assets/javascripts/surveys.js
@@ -14,7 +14,7 @@
     return (
       '<section id="user-satisfaction-survey" class="visible" aria-hidden="false">' +
       '  <div class="survey-wrapper">' +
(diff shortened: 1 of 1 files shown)


injms (approved):
This fixes the focus state visually 👍 

To me, the focus order of the 'close' and 'Take a short survey...' links seems odd - I'd expect the survey link to be focussed first and then the close link, but the opposite occurs. Not a blocker though.

alex-ju:
> To me, the focus order of the 'close' and 'Take a short survey...' links seems odd - I'd expect the survey link to be focussed first and then the close link, but the opposite occurs. Not a blocker though.

Indeed.. this is just a quick visual fix, but there are quite a few quirks with this 'component' and I hope we can review it at some point.

PR #2367, opened 2021-01-04
Allow stripping query string parameters from Analytics

kevindew:
Trello: https://trello.com/c/LiAOrW5F/1000-remove-or-redact-values-at-the-end-of-postcode-look-up-string-that-dont-match-postcode-format

This adds a new meta tag approach to stripping PII data from being sent
to analytics. This allows for particular query string parameters to be
redacted and for an application to specify them.

To use this an application should create a meta element of:

```
<meta name="govuk:static-analytics:strip-query-string-parameters"
content="param-1, param-2">
```

Where param-1, param-2 is a comma separated list of parameter names.
Formatting for this is inspired by `<meta name=keywords>`

Then when analytics data is sent to GA the strings are looked for a
match of the query string parameter and any found are replaced. To illustrate:
were a url of `https://www.gov.uk?param-1=secret` be sent to analytics,
and param-1 was flagged for stripping, then this was be replaced to be
`https://www.gov.uk?param-1=[param-1]`.

There are a few caveats of this approach that should be outlined:

It's just a regular expression that tries to match query strings and this
is done rather crudely. It just looks for a part of a string which is
`?parameter_name=<some value>` or `&paramater_name=<some value>`, so it is
possible that for some strings this could be a false positive. A more
thorough approach could be to parse URIs that are recognised in input
and break them down to constituent parts to identify query strings, but
it isn't clear this is needed.

This won't match a URL that is urlencoded - for example &amp; - this
feels like something that would occur if we were showing someone
publicly the URL they used - which seems a bit strange for PII.

As the means to specify query string parameters is a comma separated list,
this technique will not support any query string that use a comma in the
parameter name. This seemed pretty low risk to me, were we to want to support
this we may want to switch to a JSON encoded approach to defining parameter
names to strip.

An application of this can be seen in: https://github.com/alphagov/collections/pull/2202

### DIFF
--- a/app/assets/javascripts/analytics/pii.js
+++ b/app/assets/javascripts/analytics/pii.js
@@ -5,6 +5,7 @@
   var EMAIL_PATTERN = /[^\s=/?&]+(?:@|%40)[^\s=/?&]+/g
   var POSTCODE_PATTERN = /[A-PR-UWYZ][A-HJ-Z]?[0-9][0-9A-HJKMNPR-Y]?(?:[\s+]|%20)*[0-9](?!refund)[ABD-HJLNPQ-Z]{2,3}/gi
   var DATE_PATTERN = /\d{4}(-?)\d{2}(-?)\d{2}/g
+  var ESCAPE_REGEX_PATTERN = /[|\\{}()[\]^$+*?.]/g
 
   // specific URL parameters to be redacted from accounts URLs
   var RESET_PASSWORD_TOKEN_PATTERN = /reset_password_token=[a-zA-Z0-9-]+/g
@@ -19,9 +20,24 @@
     return ($('meta[name="govuk:static-analytics:strip-postcodes"]').length > 0)
   }
 
+  function queryStringParametersToStrip() {
+    var value = $('meta[name="govuk:static-analytics:strip-query-string-parameters"]').attr('content')
+    var parameters = []
+
+    if (value) {
+      var split = value.split(",")
+      for (var i = 0; i < split.length; i++) {
+        parameters.push(split[i].trim())
+      }
+    }
+
+    return parameters
+  }
+
   var pii = function () {
     this.stripDatePII = shouldStripDates()
     this.stripPostcodePII = shouldStripPostcodes()
+    this.queryStringParametersToStrip = queryStringParametersToStrip()
   }
 
   pii.prototype.stripPII = function (value) {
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/analytics/pii.js
```diff
@@ -75,6 +92,18 @@
     return array
   }
 
+  pii.prototype.stripQueryStringParameters = function (string) {
+    for (var i = 0; i < this.queryStringParametersToStrip.length; i++) {
+      var parameter = this.queryStringParametersToStrip[i]
+      var escaped = parameter.replace(ESCAPE_REGEX_PATTERN, '\\$&')
+      var regexp = new RegExp('((?:\\?|&)' + escaped + '=)(?:[^&#\\s]*)', 'g')
+      var replacement = '[' + parameter + ']'
```
andysellick, line 100: Doesn't look like this variable gets used anywhere?

spec/javascripts/analytics/pii.spec.js
```diff
@@ -138,9 +138,44 @@ describe("GOVUK.PII", function() {
     });
   });
 
+  describe('when there is a govuk:static-analytics:strip-query-string-parameters meta tag present', function() {
+    afterEach(function() {
+      resetHead();
+    });
```
andysellick, line 144: Nitpick: some of your lines end with `;` and some don't. Looks like the file before changes was inconsistent already though. Would you mind stripping them out?

andysellick (approved):
Looks good, I think. Just a couple of nitpicks.

app/assets/javascripts/analytics/pii.js
kevindew: Oops - thanks for catching.

spec/javascripts/analytics/pii.spec.js
kevindew: Yup, no problem. Old habits die hard for me with JS semi-colons 😁 . I miss the linter, but hopefully I'll get it installed to this app in the next week or two.

kevindew:
@andysellick thanks for the review. Is there anything useful I can do about keeping this in sync with the components gem?

andysellick:
@kevindew I'll copy this change across into the gem. I'm pretty close now to being able to make the full switch and retire this code in static. I might ping you for a review 😄

PR #2368, opened 2021-01-04
Update covid banner text

leenagupte:
site-wide-banner


kevindew (approved):
Looks good to me - I've not double checked the copy has been green lit though.

PR #2375, opened 2021-01-07
Set cache control on asset redirects

alex-ju:
### What
- Set `cache-control` on asset redirects
- Extend the asset redirects based on recent log analysis

### Why
Recent log analysis revealed that we have some uncached redirects and 404s on apple-touch-icons requests. This change aims to prevent such request from reaching our servers by setting a cache-control in the redirects header that will be picked up by our CDN.

[Trello card](https://trello.com/c/hNHb5JS8); [Related Trello card](https://trello.com/c/CN2bhhZi)

### DIFF
--- a/app/controllers/icon_redirects_controller.rb
+++ b/app/controllers/icon_redirects_controller.rb
@@ -1,4 +1,6 @@
 class IconRedirectsController < ApplicationController
+  before_action { expires_in(1.day, public: true) }
+
   def show
     redirect_to view_context.asset_path(request.path.to_s[1..-1]), status: :moved_permanently
   end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -27,4 +27,13 @@ Rails.application.routes.draw do
   get "/apple-touch-icon-76x76.png", to: "icon_redirects#apple_old_size_icon"
   get "/apple-touch-icon-60x60.png", to: "icon_redirects#apple_old_size_icon"
   get "/apple-touch-icon-precomposed.png", to: "icon_redirects#apple_old_size_icon"
+  get "/apple-touch-icon-114x114-precomposed.png", to: "icon_redirects#apple_old_size_icon"
+  get "/apple-touch-icon-120x120-precomposed.png", to: "icon_redirects#apple_old_size_icon"
+  get "/apple-touch-icon-144x144-precomposed.png", to: "icon_redirects#apple_old_size_icon"
+  get "/apple-touch-icon-152x152-precomposed.png", to: "icon_redirects#apple_old_size_icon"
+  get "/apple-touch-icon-176x176-precomposed.png", to: "icon_redirects#apple_old_size_icon"
+  get "/apple-touch-icon-180x180-precomposed.png", to: "icon_redirects#apple_old_size_icon"
+  get "/apple-touch-icon-57x57-precomposed.png", to: "icon_redirects#apple_old_size_icon"
+  get "/apple-touch-icon-72x72-precomposed.png", to: "icon_redirects#apple_old_size_icon"
+  get "/apple-touch-icon-76x76-precomposed.png", to: "icon_redirects#apple_old_size_icon"
 end
--- a/lib/tasks/publishing_api.rake
+++ b/lib/tasks/publishing_api.rake
@@ -80,6 +80,60 @@ ROUTES = [
     title: "Crest for Apple iOS bookmarks",
     description: "Used by iOS when saving GOV.UK as a shortcut icon.",
   },
+  {
+    base_path: "/apple-touch-icon-114x114-precomposed.png",
+    content_id: "c15ee6fd-fb3a-45ab-b9e1-0efe1a8a6c2e",
+    title: "Crest for Apple iOS bookmarks",
+    description: "Used by iOS when saving GOV.UK as a shortcut icon.",
+  },
+  {
+    base_path: "/apple-touch-icon-120x120-precomposed.png",
+    content_id: "2e676337-94f3-4f4f-8501-83a1c7813bf8",
+    title: "Crest for Apple iOS bookmarks",
+    description: "Used by iOS when saving GOV.UK as a shortcut icon.",
+  },
+  {
+    base_path: "/apple-touch-icon-144x144-precomposed.png",
+    content_id: "5fd9d11e-997a-4850-ad69-a43a3aec8a9f",
+    title: "Crest for Apple iOS bookmarks",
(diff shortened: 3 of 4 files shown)


app/controllers/icon_redirects_controller.rb
```diff
@@ -1,9 +1,11 @@
 class IconRedirectsController < ApplicationController
   def show
+    response.headers["Cache-Control"] = "max-age=315360000, public, immutable"
```
kevindew, line 3: I don't think think these are quite the right headers as these icons may change over time - the immutable indefinitely cacheable resource is the item that is redirected too.

Otherwise if the icon changes caches for www.gov.uk/apple-touch-icon-114x114-precomposed.png will indefinitely redirect to an old resource.

I'd suggest going for a more modest time like 1 hour? You kind of want it to be long enough to have an impact but be short enough we don't actually want to have to ever purge the cache when things change.

You can also set it in a nicer Rails fashion with something like `expires_in(1.hour, public: true)` if you put it in a before action you could specify once for the whole controller. For example:  `before_action { expires_in(1.hour, public: true) }`

chao-xian (dismissed):

kevindew (commented):
I think there's a bit of a snag in this. Hopefully I beat before you click merge.

app/controllers/icon_redirects_controller.rb
alex-ju: This is the header for the redirect, not the actual asset we redirect to. We're essentially making the redirect permanent. The assets themselves have a different header.

kevindew: Hmm I feel like one of us has our wires crossed. Apologies if that ends up being me!

This is the action for a request to https://www.gov.uk/favicon.ico right?

So currently when we send a request to https://www.gov.uk/favicon.ico we receive a redirect to https://www.gov.uk/assets/static/favicon-8d811b8c3badbc0b0e2f6e25d3660a96cc0cca7993e6f32e98785f205fc40907.ico and that has a cache header of "cache-control: no-cache"

With your change a request to https://www.gov.uk/favicon.ico would return a redirect to https://www.gov.uk/assets/static/favicon-8d811b8c3badbc0b0e2f6e25d3660a96cc0cca7993e6f32e98785f205fc40907.ico which should be cached for 10 years?

Then imagine if we change the favicon asset at some point - a new asset would be created with a new fingerprint id and that 10 year cache redirect would at best be going to the old item and at worse be going to a 404 (given we don't have reason to keep old asset versions forever).

kevindew: Sorry I replied before your edit seems like we were more in alignment than I realised!

alex-ju: Reduced to 1 day and used the more elegant `before_action` method. The actual assets we point to are set to 10 years – tried to be consistent with our current caching policy 🤷‍♂️

kevindew: The actual assets are fine to be 10 years because we'd never expect https://www.gov.uk/assets/static/favicon-8d811b8c3badbc0b0e2f6e25d3660a96cc0cca7993e6f32e98785f205fc40907.ico to return any other response than that version of the asset. Whereas we'd expect www.gov.uk/favico.ico to vary based on what the current favicon is.

app/controllers/icon_redirects_controller.rb
```diff
@@ -1,4 +1,9 @@
 class IconRedirectsController < ApplicationController
+
+  before_action {
+    expires_in(1.day, public: true)
+  }
```
kevindew, line 5:
```suggestion
  before_action { expires_in(1.day, public: true) }
```

Minor nit pick, https://rubystyle.guide/#single-line-blocks

app/controllers/icon_redirects_controller.rb
alex-ju: To me `https://www.gov.uk/assets/static/favicon-8d811b8c3badbc0b0e2f6e25d3660a96cc0cca7993e6f32e98785f205fc40907.ico` returns `max-age=315360000, public, immutable`

config/routes.rb
```diff
@@ -27,4 +27,13 @@
   get "/apple-touch-icon-76x76.png", to: "icon_redirects#apple_old_size_icon"
   get "/apple-touch-icon-60x60.png", to: "icon_redirects#apple_old_size_icon"
   get "/apple-touch-icon-precomposed.png", to: "icon_redirects#apple_old_size_icon"
```
kevindew, line 29: I notice some of these return 404s:

```
➜  ~ curl -sI https://www.gov.uk/apple-touch-precomposed.png | grep HTTP
HTTP/2 404
➜  ~ curl -sI https://www.gov.uk/apple-touch-icon-76x76.png | grep HTTP
HTTP/2 404
➜  ~ curl -sI https://www.gov.uk/favicon.ico | grep HTTP
HTTP/2 301
```

Which makes me wonder if there's a part of puppet that needs to be kept in sync with this. I'll just have a look and comment further with my findings. I wouldn't want you to add these and find out that no-one can access them.

alex-ju: 🤔

app/controllers/icon_redirects_controller.rb
alex-ju: Not having rubocop in the main rake task bites me every time.

config/routes.rb
kevindew: In a rather grim TIL moment I've discovered these assets are put into the Publishing API 🤯  via: https://github.com/alphagov/static/blob/fc25d7873c7aec702805abd7c5e20792e8989d5e/lib/tasks/publishing_api.rake#L48-L83 meaning we have content items like: https://www.gov.uk/api/content/favicon.ico

It looks like you'll need to add the new items to this list too and run the rake tasks when they're published. There's surely a nicer way these could have been done 🤔 

Btw the first Item I checked https://www.gov.uk/apple-touch-precomposed.png I had the URL wrong it's fine for https://www.gov.uk/apple-touch-icon-precomposed.png

alex-ju: I spotted this a few days ago, but couldn't make sense of the _why_ – I'll add them there too.

alex-ju: Thanks for the support on this – I added https://github.com/alphagov/static/pull/2375/commits/d20c1c23d019a1fe53f2fae4b9060b11da26458c to prepare the new assets to be published via Publishing API

thomasleese (approved):
Having these in the Publishing API is a bit unpleasant... but it's how they've been done so far!

PR #2376, opened 2021-01-07
Use analytics from gem

andysellick:
## What

Switch to using the analytics code in the components gem, recently copied from static.

Interesting thing to note. When this branch was deployed for testing to integration, and then staging, the GA property for the main GOV.UK analytics was not the expected value. Instead of ending with `-1`, on integration it ended with `-22` and on staging with `-20`. These GA properties are real and collect data, so it looks like this is a deliberate thing, but I can't see where in the code this change might be occurring.

## Why

Accounts needed the analytics code in the gem but we didn't want to disturb the whole of GOV.UK, so it was copied across and used solely by that app. However, our intent has always been to put it into the gem, so this completes both that and removes the brittleness of having to maintain two copies of the same bit of code.

## Changes

If this PR works correctly, nothing at all should change. All existing tracking should remain fully functional and as before.

Trello cards:

- https://trello.com/c/p3UtkNtE/420-resolve-the-two-copies-of-analytics-code
- https://trello.com/c/epZ12grR/110-migrate-analytics-code-into-the-gem


danacotoran (approved):
Insofar as I understand this, it seems to work as expected! 
From testing this locally with GA Debug vs testing on the live site, I could not spot any errors, or differences in how the tracking works.

PR #2380, opened 2021-01-11
Adding relevant sass for search component

chris-gds:
## What?

Fix to resolve visual issue on `government-frontend` app within search input, adding support file directly to static so the search component pulls in correct mixin to calculate the height for an input field.

## Why?

There was an issue highlighted with search [on this page ](https://www.gov.uk/guidance/access-the-public-register-for-environmental-information)which was related to a [search PR](https://github.com/alphagov/govuk_publishing_components/pull/1793)

A [PR](https://github.com/alphagov/government-frontend/pull/1950) was raised to fix this issue directly on `government-frontend` however it was highlighted this isn't quite the correct fix.   This PR adds the relevant file directly to static so the header + search element has the sass that it needs.

## Visuals

Before
Screenshot 2020-12-22 at 18 00 57

After
Screenshot 2020-12-22 at 18 01 15

## Anything else 

There was some conversation about this previously on [this PR](https://github.com/alphagov/static/pull/2049) and [this PR](https://github.com/alphagov/static/pull/2238) - raising as a draft.

### DIFF
--- a/app/assets/stylesheets/header-footer-only.scss
+++ b/app/assets/stylesheets/header-footer-only.scss
@@ -2,6 +2,8 @@ $govuk-compatibility-govuktemplate: true;
 $govuk-use-legacy-palette: false;
 
 @import "govuk_publishing_components/govuk_frontend_support";
+// The search component requires this helper mixin to render height #PR-2380
+@import "govuk_publishing_components/components/helpers/px-to-em";
 @import "govuk_publishing_components/components/_cookie-banner";
 // collections uses the action-link component and relies upon this line, see
 // https://github.com/alphagov/collections/pull/1754


andysellick (dismissed):
@chris-gds there's a reason why `component_support` wasn't included already, which is that it pulls in all of govuk-frontend, which increases the size of the CSS across GOV.UK (in fact this is the reason the components gem has this and the separate `govuk_frontend_support` file in the first place - govuk_frontend_support includes only those govuk-frontend imports that don't add CSS weight i.e. only mixins and variables).

Is it possible you could achieve the required effect by doing this?

`@import "govuk_publishing_components/components/helpers/px-to-em";`

andysellick:
(note to self - I should probably document this)

chris-gds:
> @chris-gds there's a reason why `component_support` wasn't included already, which is that it pulls in all of govuk-frontend, which increases the size of the CSS across GOV.UK

Thanks for taking a look at the draft so quickly, I did hesitate thinking this would impact performance.  Updated with the suggested solution as static is the unusual aspect of this.

andysellick (approved):

PR #2385, opened 2021-01-13
[DO NOT MERGE] Revert "Use analytics from gem"

andysellick:
Reverts alphagov/static#2376

I'm creating this just in case... I hope I won't need it.


danacotoran (approved):
Good foresight. I hope this isn't needed 😬

andysellick:
Not needed, deployment of analytics appears to be working correctly 😅

PR #2390, opened 2021-01-14
Hyphenate the word "sitewide" on search form aria-label

danacotoran:
VoiceOver reads out the "Sitewide" `aria-label` on the search form in a strange way. It seems that writing it as "Site-wide" helps VoiceOver read it out more clearly.

### Before
https://user-images.githubusercontent.com/7116819/104621026-8eb91500-5687-11eb-800e-59d546c5e8f9.mov

### After
https://user-images.githubusercontent.com/7116819/104621049-95478c80-5687-11eb-89aa-07d07e4afe08.mov


injms (approved):
Nice catch - thanks for fixing.

Super use of video in the pull request as well 👍

PR #2391, opened 2021-01-14
Use correct component references

andysellick:
Update component references following this change to correct component file names: https://github.com/alphagov/govuk_publishing_components/pull/1848


maxgds (approved):

maxgds (approved):
Sorry about confusion on my part between js and component call in previous review.

PR #2397, opened 2021-01-19
Correct outdated Gds API Adapters usage

kevindew:
The PublishingApiV2 namespace has been removed. I also simplified the Publishing API call since we can use the GdsApi.publishing_api method to access the adapter.


alex-ju (approved):
Thanks for this fix, Kevin!
Tested in integration with the task changes in https://github.com/alphagov/static/pull/2375 and works well.

PR #2398, opened 2021-01-19
Error template markup adjustments

danacotoran:
Update the error pages layout to more closely match the Design system [pattern for error pages](https://design-system.service.gov.uk/patterns/page-not-found-pages/default/index.html) and amend some spacing issues.

There is now more space between the page header and the content, which were previously too close together on mobile. Additionally, the content is now inside a `two-thirds` column instead of running full-width.

https://trello.com/c/Mmv4gF5o

Before | After
------------ | -------------
Screenshot 2021-01-20 at 10 39 39 | Screenshot 2021-01-20 at 10 42 35

### Before
Screenshot 2021-01-20 at 10 44 00

### After
Screenshot 2021-01-20 at 10 43 43

### DIFF
--- a/app/views/root/_error_page.html.erb
+++ b/app/views/root/_error_page.html.erb
@@ -12,29 +12,26 @@
 <% content_for :body_classes, "mainstream error" %>
 
 <% content_for :wrapper_content do %>
-<main id="content" role="main" class="group">
-  <header class="page-header group">
-    <div>
-      <h1><%= heading %></h1>
-    </div>
-  </header>
-  <div class="article-container">
-    <article role="article" class="group">
-      <div class="inner">
+  <main id="content" role="main" class="govuk-main-wrapper govuk-main-wrapper--l">
+    <div class="govuk-grid-row">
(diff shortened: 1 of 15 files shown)


app/views/root/_error_page.html.erb
```diff
@@ -12,29 +12,23 @@
...
-      <h1><%= heading %></h1>
-    </div>
-  </header>
-  <div class="article-container">
-    <article role="article" class="group">
-      <div class="inner">
+  <main id="content" role="main" class="govuk-main-wrapper govuk-main-wrapper--l">
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-two-thirds">
+        <h1 class="govuk-heading-l"><%= heading %></h1>
```
injms, line 18: Could a [page title](https://components.publishing.service.gov.uk/component-guide/title) component be used here?

danacotoran: good shout @injms, I've updated it to use the component as suggested.

injms (approved):
Super - thank you 👍

PR #2408, opened 2021-02-02
Replace hard coded account nav with layout header

danacotoran:
During development the account nav markup was copied over from the layout header component instead of using the component itself. This was probably due to time constraints. 

Screenshot 2021-02-02 at 18 30 08

The component audits are complaining that we are including the assets for `layout_header`, when we are not using the component.

To fix this problem, this removes the hard coded account navigation partial and uses `layout_header` instead. This is technically duplicating the header, so this change is dependent on some changes in Slimmer which will conditionally reveal this layout header only when the account nav should be present.

Whereas it made sense to have the hard coded account navigation within the `:inside_header` content block (since obviously the header navigation links should be inside the header), the `layout_header` has to go after the existing header (in the `:after_header` content block). 
This is because [Slimmer will always conditionally hide one of the headers and reveal the other](https://github.com/alphagov/slimmer/pull/263), therefore they can not be nested within each other. 

### Dependent on these PRs
- https://github.com/alphagov/govuk_publishing_components/pull/1902 
- https://github.com/alphagov/slimmer/pull/263 

🙅🏻‍♀️ **Do not merge this in before the PRs above have been deployed** 🙅🏻‍♀️

### Trello 
https://trello.com/c/9Kjy1hjb


andysellick (commented):
I can't get this running locally due to an unrelated problem with my local setup. The change looks good, my only thought is we discussed why the code has moved from the `unless` statement to the `content_for :after_header`, can you record the explanation here as well?

danacotoran:
Thanks @andysellick and my bad, I completely missed your comment.
I've added a (probably not very well worded) explanation to the PR description, I hope it makes sense.

andysellick (approved):

PR #2411, opened 2021-02-09
Add layouts that use the gem layout component

injms:
## What

Adds `gem_layout` and `gem_layout_full_width` as *optional* templates that can be requested by Slimmer. The plan is to swap the use of `core_layout` for `gem_layout` in each frontend app - and that both templates will be running side by side until this is complete.

Don't use this layout until the frontend app you're using has Slimmer v15.5.0 or later.

This can be viewed when running static locally at http://static.dev.gov.uk/templates/gem_layout.html.erb

## Why

The current `core_layout` uses [`govuk_template`][1], which is deprecated. `govuk_template` relies on [`govuk_frontend_toolkit`][2] for styles, which is also deprecated.

 The `gem_layout` template uses the [public layout component][3] from the GOV.UK Publishing Components gem, which is maintained and regularly updated. The gem relies on the [GOV.UK Design System][govuk-frontend], which is maintained and regularly updated.

Core layout contains styles that can interfere with components, for which hacks have to be written; for example [focus link overrides][4] and [specific overrides in the cookie banner][5]. These can be removed when the `core_layout` template stops being used.

Removing the use of `govuk_template` will mean that the compatibility mode (`$govuk-compatibility-govuktemplate`) can be turned off.

Because the `gem_layout` template only uses components which follow the [component principles][6] and [component conventions][7] there are no side effects - the styles and JavaScript won't affect anything outside of the components used in the template itself.

## Appearance

Note - both the menu and the search appear at the same time. Inside each app Slimmer takes care of showing and hiding those parts of the header, so only one or none of those will appear.

On desktop / large screens:

image

and with emergency banner:

image

On mobile / smaller screens:
image

and with emergency banner:

image

[1]: https://github.com/alphagov/govuk_template
[2]: https://github.com/alphagov/govuk_frontend_toolkit
[3]: https://components.publishing.service.gov.uk/component-guide/layout_for_public

[4]: https://github.com/alphagov/govuk_publishing_components/blob/94732d047a7375bd9c7350e8aeb26fc287aca4dd/app/assets/stylesheets/govuk_publishing_components/components/mixins/_govuk-template-link-focus-override.scss
[5]: https://github.com/alphagov/govuk_publishing_components/blob/220193a9d0f546195cf145c4af25d515385324e6/app/assets/stylesheets/govuk_publishing_components/components/_cookie-banner.scss#L56 

[6]: https://github.com/alphagov/govuk_publishing_components/blob/master/docs/component_principles.md
[7]: https://github.com/alphagov/govuk_publishing_components/blob/master/docs/component_conventions.md
[govuk-frontend]: https://github.com/alphagov/govuk-frontend/


app/views/root/gem_layout.html.erb
```diff
@@ -0,0 +1,3 @@
+<%= render partial: "gem_base", locals: {
+  full_width: false,
```
alex-ju, line 2: I guess we can omit this here, as it defaults to false in layout_header, and if nil is passed in `gem_base` it will still be false

README.md
```diff
@@ -48,7 +50,7 @@ Restarting the app should pick up the changes.
 For example, to see changes made to static templates which
 are wrapped around feedback pages, run `bowl feedback
```
alex-ju, line 51: not-related to this change, but I'm not sure bowl is still a thing 😬

alex-ju (commented):
🚢

app/views/root/gem_layout.html.erb
injms: Omitting it throws an error, since `full_width` isn't declared in gem_base - to get around this I've set `full_width` to default to false so it doesn't need to be set.

alex-ju: ah, right – these are the partial locals, not the component call – my bad

alex-ju (approved):
The above 🚢  was meant to be an approved review :shipit:

PR #2412, opened 2021-02-09
Remove wrapper and homepage layouts

alex-ju:
Remove `wrapper` and `homepage` layouts as they are no longer used.

In an effort to converge to a single layout for all application:
 - [`licence-finder` was updated to use `core-layout`, replacing the deprecated `wrapper`](https://github.com/alphagov/licence-finder/pull/914)
 - [`frontend` was updated to use `core-layout` in entirety, replacing the remaining `homepage`](https://github.com/alphagov/frontend/pull/2616) 

See below the two apps tested with this branch of `static`

<table>
<tr><th>frontend</th><th>licence-finder</th></tr>
<tr><td valign="top">

frontend dev gov uk

</td><td valign="top">

licence-finder dev gov uk

</td></tr>
</table>

[Trello card](https://trello.com/c/Dpe7MWQp)


app/views/root/homepage.html.erb
```diff
@@ -1,13 +0,0 @@
...
- # after the govuk_template. govuk_template as it stands doesn't have a
- # content_for which could otherwise be used.
-
- # This is only wanted for the homepage and can't be put in to the frontend
- # template on frontend due to the way slimmer would not copy over the
- # comment
-%>
-<%= render partial: 'base', locals: { hide_nav: true } %>
-<%= render template: 'layouts/govuk_template' %>
-<!-- Thanks Martha -->
```
andysellick, line 13: I had a quick look to see where this comment comes out now but I can't see it - do we no longer have this? Was there a decision made to remove it from GOV.UK entirely?

andysellick (approved):
Looks good, just one question.

app/views/root/homepage.html.erb
alex-ju: That is a good point, Andy. I will move it in the `core-layout` template in this PR so we don't forget about it.

andysellick: I just checked and I can't see it on the live homepage 😱

Issue #2182, opened 2020-05-18
Emergency banner uses non-standard prefix


alex-ju:
[emergency-banner](https://github.com/alphagov/static/blob/master/app/assets/stylesheets/helpers/_govuk-emergency-banner.scss) uses a [non-standard](https://github.com/alphagov/govuk_publishing_components/blob/master/docs/component_conventions.md) `govuk-` prefix

PR #2417, opened 2021-02-15
Update emergency banner

injms:
## What

 * Renamed stylesheet file and classes to remove `govuk` namespace; the `govuk` namespace is used by the GOV.UK Frontend / Design System, and using it here could cause confusion. Fixes https://github.com/alphagov/static/issues/2182.
 * Updated layout to use GOV.UK Frontend layout classes; BEM'd up the class naming scheme to be clearer that these are part of the emergency banner, and updated the styles to match.
 * Updated tests to take new class names into account
 * Update layout to take into account the global banner

## Why

The emergency banner used deprecated variables that are going to be remove - such as `$red` - and needed to be updated to avoid it breaking when these are eventually deleted. I took the opportunity to update the layout without changing the layout.

## Visual differences

There's one small visual difference when you look at _only_ the template itself:

Before on just the core_layout template:
static-banner-before

After on just the core_layout template:

static-banner-after-national-emergency

The width has changed from being hardcoded in static's stylesheet to coming from `govuk-width-container` GOV.UK Frontend, which isn't included by default in this template, hence why the banner is not contained to a set width. This is fine when used in a frontend app - all of which have the GOV.UK Frontend layout styles.

Homepage:

| | large screen size | small screen size|
| - | - | - |
| Before | image | image |
|After | Screenshot_2021-02-15 Welcome to GOV UK | Screenshot_2021-02-15 Welcome to GOV UK |

Non-homepage page:

| | large screen size | small screen size|
| - | - | - |
| Before | image | image |
| After | Screenshot_2021-02-15 Find your local council - GOV UK | Screenshot_2021-02-15 Find your local council - GOV UK(1) |

With JavaScript disabled - which causes the global banner to not appear:

Screenshot 2021-02-15 at 15 35 57

Screenshot 2021-02-15 at 15 37 00


alex-ju (approved):
Excellent! This also fixes #2182.

injms:
I've tested this again on integration using the [emergency banner guidance](https://docs.publishing.service.gov.uk/manual/emergency-publishing.html) to ensure that it works properly. The pages listed to test work properly on all banner types.

PR #2426, opened 2021-02-26
Split the font property into separate properties for consistency

Nooshu:
We've used separate font properties for other areas of GOV.UK. This PR updates to use the same method for consistency, and removed "GDS Transport" from the font stack since we aren't currently using v2 of the font on GOV.UK.


alex-ju (approved):
Excellent! Will get this to production soon.

PR #2428, opened 2021-03-04
Remove dependency on `govuk_frontend_toolkit`

alex-ju:
### What
Remove dependency on `govuk_frontend_toolkit` by replacing variables and mixins with `govuk-frontend` equivalents.

### Why
`govuk_frontend_toolkit` is a deprecated dependency; the way `static` uses it is tangled, unclear and untested

### Note
I checked the [JS imports](https://github.com/alphagov/govuk_frontend_toolkit/tree/master/javascripts/govuk) before wiping them out. Part of them are covered by `govuk-frontend`, part migrated to `govuk_publishing_components`, part not relevant/used anymore. Most importantly `application.js` (the last one importing `govuk_toolkit`) is only used as a fallback in `static` (I know, 🙃) as all our frontend application are using `header-footer-only.js` which doesn't rely on scripts from toolkit anymore (recently removed while refactoring).


app/assets/stylesheets/helpers/_organisations.scss
```diff
@@ -34,124 +34,124 @@
...
 
-@mixin organisation-logo-border($colour) {
-  border-color: $colour;
+@mixin organisation-logo-border($organisation) {
+  border-color: govuk-organisation-colour($organisation);
 }
 
 .attorney-generals-office {
-  @include organisation-logo-border($attorney-generals-office);
+  @include organisation-logo-border('attorney-generals-office');
```
andysellick, line 50: This looks good, but do we need these organisation colours at all anymore? Where are they used? I know some components in the gem use them, but that code isn't reliant on this.

andysellick (approved):
Amazing. Kill it with fire. Great stuff.

app/assets/stylesheets/helpers/_organisations.scss
alex-ju: I also suspect we can remove this, but I'll need to check a few apps before doing that.

PR #2430, opened 2021-03-08
Allow showing individual global bar links

sihugh:
It's possible that we might want to show only one of the Brexit/Covid links that are on the global bar.

This PR allows this and makes the margin underneath the upper link dependent on whether there is anything below it.

## Just Coronavirus

localhost_3090_student-finance (3)

## Just Brexit

localhost_3090_student-finance (1)

## Both links (the margin survives)

localhost_3090_student-finance (2)


hannako (approved):
🌉

PR #2433, opened 2021-03-15
Uncouple components from template

alex-ju:
### What
Uncouples the elements we want to preserve (emergency banner, global banner, survey banner) from what we want to replace (the `govuk_template` bits).

### Why
To easy the process of replacing `govuk_template ` and to be able to test and preview components in isolation.


alex-ju:
@injms hopefully I'm not stepping on toes with this PR – happy to close this in favour of an alternative approach from the accessibility team.

injms (approved):
🙌

PR #2437, opened 2021-03-18
Update header search button text

injms:
## What
The button text for the search button in the header has been changed from "Search" to the more specific "Search GOV.UK".

## Why

The default text for the button in the sarch component is "Search" - this means that multiple forms can have the same submit text. This can cause confusion when navigating out of context - for example, a screenreader will pull out all of the interactive elements on a page and list them without the surrounding information.

### DIFF
--- a/app/views/root/_base.html.erb
+++ b/app/views/root/_base.html.erb
@@ -14,15 +14,19 @@
 
 <% content_for :inside_header do %>
   <button class="search-toggle js-header-toggle" data-search-toggle-for="search">Show or hide search</button>
-  <% # The /search page redirects to a finder if keywords are included. Be careful about
-     # changing this, as the redirect adds some parameters to the search query. %>
+  <%
+    # The /search page redirects to a finder if keywords are included. Be
+    # careful about changing this, as the redirect adds some parameters to the
+    # search query.
+  %>
   <form id="search" class="site-search govuk-clearfix" action="/search" method="get" role="search" aria-label="Site-wide">
     <div class="content govuk-clearfix">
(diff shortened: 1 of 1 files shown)


maxgds (approved):
I think this looks good enough to go out. 

@selfthinker could you check it in Dragon - I'm interested to know if introducing "gov.uk" as part of the wording might cause challenges as it doesn't have a widely known way to be pronounced? If it'll cause problems we might want to iterate on this change.

injms:
I'm going to merge this in because there currently is another use of "gov.uk" in the search, and this was a recommendation from DAC - both of which makes me think that this is okay. If this turns out to be a problem we'll have to change it in multiple places anyway.

selfthinker:
> @selfthinker could you check it in Dragon - I'm interested to know if introducing "gov.uk" as part of the wording might cause challenges as it doesn't have a widely known way to be pronounced? If it'll cause problems we might want to iterate on this change.

Sorry for the late response.
I just checked and there was nothing unexpected happening.

Because you specifically mentioned that GOV.UK "doesn't have a widely known way to be pronounced", I wonder if this is something that is better tested in user testing rather than compatibility testing. Because I won't know which ways people might pronounce it.
I have tried "goff you kay" as well as "goff dot you kay". Both were working fine. There might be more ways to say it but those are the ways I could think of.

Also, people don't need to pronounce "GOV.UK" at all in order to activate the search box. I wonder how many people do it anyway. But you can just say "click search" and Dragon will offer you both the search box as well as the button.

And lastly, "Search on GOV.UK" is already in use, because that is the label that is used on the homepage.

PR #2440, opened 2021-03-19
Update error pages to use public layout template

injms:
## What

 * The base error page template has been updated to use the public layout component
 * Application stylesheet has been updated to include only the styles needed  for the public layout component, global bar, and emergency banner
 * Application JavaScript has been updated to only include the modules needed for the public layout component
 * Error page stylesheet has only the extra imports needed - in this case, only the the title component was needed
 * Minor tweak to make the apostrophe in the "We're" on some of the error pages look nicer
 * Fixes a minor bug where the heading styles weren't being included, so the `h1` on the error page was using the default browser font size for `h1`s.

## Why
Adding a template to static that uses the public layout component meant updating these error pages first, since they use the same default `application.scss` and `application.js` that the public layout component uses.

The default stylesheet and JavaScript file now contains only the things needed for the public layout component.

## Visual differences

Note - the page footer will look different as it the public layout component uses the [footer component](https://components.publishing.service.gov.uk/component-guide/layout_footer). These changes are outside the scope of this pull request and are best raised as issues on the [govuk_publishing_components repo](https://github.com/alphagov/govuk_publishing_components/issues?q=is%3Aissue+is%3Aopen)

| Page | Before | After |
| - | - | - |
| 400 | before - firefox - 400-html-erb | firefox-400-html-erb |
| 401 | before - firefox - 401-html-erb | firefox-401-html-erb |
| 403 | before - firefox - 403-html-erb | firefox-403-html-erb |
| 404 | before - firefox - 404-html-erb | firefox-404-html-erb |
| 405 | before - firefox - 405-html-erb | firefox-405-html-erb |
| 406 | before - firefox - 406-html-erb | firefox-406-html-erb |
| 410 | before - firefox - 410-html-erb | firefox-410-html-erb |
| 422 | before - firefox - 422-html-erb | firefox-422-html-erb |
| 429 | before - firefox - 429-html-erb | firefox-429-html-erb |
| 500 | before - firefox - 500-html-erb | firefox-500-html-erb |
| 501 | before - firefox - 501-html-erb | firefox-501-html-erb |
| 502 | before - firefox - 502-html-erb | firefox-502-html-erb |
| 503 | before - firefox - 503-html-erb | firefox-503-html-erb |
| 504 | before - firefox - 504-html-erb | firefox-504-html-erb |

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,4 +63,4 @@ development VM it's possible to increase the `proxy_read_timeout` value in
 
 Tests can run in browser at `/specs`
 
-Or in terminal to run only the jasmine tests you can use `RAILS_ENV=test bundle exec rake jasmine:ci`
+Or in terminal to run only the jasmine tests you can use `RAILS_ENV=test bundle exec rake jasmine:ci` or in Docker `govuk-docker run -e RAILS_ENV=test static-lite bundle exec rake jasmine:ci`
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -1,1 +1,17 @@
-//= require header-footer-only
+//= require govuk_publishing_components/dependencies
+
+//= require govuk_publishing_components/lib/cookie-functions
+//= require govuk_publishing_components/lib/header-navigation
+//= require govuk_publishing_components/lib/track-click
+
+//= require analytics
+
+//= require govuk_publishing_components/components/cookie-banner
+//= require govuk_publishing_components/components/layout-header
+//= require govuk_publishing_components/components/feedback
+
+//= require modules/global-bar
+//= require modules/cross-domain-tracking
+
+//= require global-bar-init
+//= require surveys
--- a/app/assets/stylesheets/error-page.scss
+++ b/app/assets/stylesheets/error-page.scss
@@ -1,9 +1,2 @@
-$govuk-compatibility-govuktemplate: true;
-$govuk-use-legacy-palette: false;
-
 @import "govuk_publishing_components/govuk_frontend_support";
-@import "govuk_publishing_components/component_support";
-@import "govuk_publishing_components/components/cookie-banner";
-@import "govuk_publishing_components/components/button";
-@import "govuk_publishing_components/components/feedback";
-@import "govuk_publishing_components/components/action-link";
+@import "govuk_publishing_components/components/title";
--- a/app/views/root/400.html.erb
+++ b/app/views/root/400.html.erb
@@ -1,5 +1,5 @@
 <%= render partial: "error_page", locals: {
     heading: "Sorry, there is a problem",
     intro: '<p class="govuk-body">Go back and change the information you entered.</p>',
-    status_code: 400
+    status_code: 400,
 } %>
--- a/app/assets/stylesheets/application-print.scss
+++ b/app/assets/stylesheets/application-print.scss
@@ -1,5 +1,14 @@
-// PRINT STYLESHEET COMPILER
-// OUTPUT
-// The following files and inline CSS will form the output of print.css
-@import "header-footer-only-print";
-@import "helpers/core-print";
+// is-print is used by govuk_frontend_toolkit to alter the printed font-stack
+$is-print: true;
+
+@import "govuk_publishing_components/govuk_frontend_support";
+@import "govuk_publishing_components/component_support";
+
+@import "govuk_publishing_components/components/print/button";
+@import "govuk_publishing_components/components/print/feedback";
+@import "govuk_publishing_components/components/print/layout-footer";
+@import "govuk_publishing_components/components/print/layout-header";
+@import "govuk_publishing_components/components/print/search";
+@import "govuk_publishing_components/components/print/skip-link";
+
+@import "helpers/print-base";
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -1,9 +1,16 @@
-$govuk-compatibility-govuktemplate: true;
-$govuk-use-legacy-palette: false;
-
+@import "govuk_publishing_components/govuk_frontend_support";
 @import "govuk_publishing_components/component_support";
-@import "header-footer-only";
 
-// styles
-@import "helpers/organisations";
-@import "helpers/wrapper";
+@import "govuk_publishing_components/components/action-link";
+@import "govuk_publishing_components/components/button";
+@import "govuk_publishing_components/components/cookie-banner";
+@import "govuk_publishing_components/components/feedback";
+@import "govuk_publishing_components/components/input";
+@import "govuk_publishing_components/components/layout-footer";
+@import "govuk_publishing_components/components/layout-for-public";
+@import "govuk_publishing_components/components/layout-header";
+@import "govuk_publishing_components/components/search";
+@import "govuk_publishing_components/components/skip-link";
+
+@import "components/emergency-banner";
+@import "components/global-bar";
--- a/app/assets/stylesheets/error-page-print.scss
+++ b/app/assets/stylesheets/error-page-print.scss
@@ -1,10 +1,6 @@
-$govuk-compatibility-govuktemplate: true;
-$govuk-use-legacy-palette: false;
-
 // is-print is used by govuk_frontend_toolkit to alter the printed font-stack
 $is-print: true;
 
 @import "govuk_publishing_components/govuk_frontend_support";
-@import "govuk_publishing_components/component_support";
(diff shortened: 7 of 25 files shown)


app/assets/stylesheets/error-page.scss
```diff
@@ -1,9 +1,3 @@
-$govuk-compatibility-govuktemplate: true;
-$govuk-use-legacy-palette: false;
-
-@import "govuk_publishing_components/govuk_frontend_support";
 @import "govuk_publishing_components/component_support";
-@import "govuk_publishing_components/components/cookie-banner";
-@import "govuk_publishing_components/components/button";
-@import "govuk_publishing_components/components/feedback";
-@import "govuk_publishing_components/components/action-link";
+
```
maxgds, line 2: I think we might still need the cookie banner support - if the first page you land on on gov.uk fails for some reason you currently still get the banner served, so we probably need to keep it the same.

injms: Good spot, thank you!

injms: I had a check and the cookie banner works, phew. 

The cookie banner styles come from [`application.scss`](https://github.com/alphagov/static/blob/5f423d27fb90e90010604404dbd113034f47e1a6/app/assets/javascripts/application.js#L10):
https://github.com/alphagov/static/blob/5f423d27fb90e90010604404dbd113034f47e1a6/app/assets/javascripts/application.js#L10

The `error-page.scss` should only contain components and styles that are needed on the error pages but aren't in the application stylesheet.

app/assets/javascripts/application.js
```diff
@@ -1 +1,21 @@
-//= require header-footer-only
+/* global $ */
+
+//= require govuk_publishing_components/dependencies
+
+//= require govuk_publishing_components/lib
+//= require govuk_publishing_components/lib/track-click
```
andysellick, line 6: Confusingly, `lib` is both a directory and a file, `lib.js`. I'd thought that this would include the whole directory, but that would need `require_tree`, so I think this is including `lib.js`. However, `lib.js` then includes everything in the `lib` directory, plus the polyfills.

Since `lib` the directory contains `track-click`, line 6 isn't needed?

app/assets/javascripts/application.js
```diff
@@ -1 +1,21 @@
-//= require header-footer-only
+/* global $ */
+
+//= require govuk_publishing_components/dependencies
+
+//= require govuk_publishing_components/lib
```
andysellick, line 5: Is everything in `lib` needed for the error pages? Based on a quick glance, you might only need `track-click` and `cookie-functions`?

app/assets/javascripts/application.js
```diff
@@ -1 +1,21 @@
...
+
+//= require analytics
+
+//= require govuk_publishing_components/components/cookie-banner
+//= require govuk_publishing_components/components/layout-header
+//= require govuk_publishing_components/components/feedback
+//= require govuk_publishing_components/modules
+
+//= require modules/global-bar
+//= require modules/cross-domain-tracking
```
andysellick, line 16: Probably out of the scope of this PR, I'm left wondering if this script is needed, or whether it should be migrated into the gem with the rest of the analytics code?

app/assets/javascripts/application.js
```diff
@@ -1 +1,21 @@
...
+
+//= require global-bar-init
```
andysellick, line 18: Interestingly `global-bar-init` includes the cookie functions code from the gem, which is already pulled in by `analytics.js` in the gem, which is included here on line 8. Again, possibly out scope for this PR, but I wonder if there's an optimisation to made here, or whether rails is smart enough to not duplicate this code?

andysellick (commented):
I haven't done a full review, just writing down some of the thoughts we had when we talked about this.

app/assets/javascripts/application.js
injms: From a bit of trial and error, I think that the only required bits are:

 - `cookie-functions`
 - `header-navigation`
 - `track-click`

📉 👍

app/assets/javascripts/application.js
injms: I'd like to leave that to another PR - I've added it to an [issue noting it though][1].

[1]: https://github.com/alphagov/static/issues/2443

app/assets/javascripts/application.js
injms: I don't know if there's any de-duping or tree-shaking going on in Rails's asset pipeline. This is the kind of thing that Webpack would help with - we could include dependencies as many times as we want and Webpack would ensure that they're only in the compiled code once.

README.md
```diff
@@ -63,4 +63,4 @@ development VM it's possible to increase the `proxy_read_timeout` value in
 
 Tests can run in browser at `/specs`
 
-Or in terminal to run only the jasmine tests you can use `RAILS_ENV=test bundle exec rake jasmine:ci`
+Or in terminal to run only the jasmine tests you can use `RAILS_ENV=test bundle exec rake jasmine:ci` or in Docker `govuk-docker run -e RAILS_ENV=test static-lite bundle exec rake jasmine:ci`
```
owenatgov, line 66: If the image (?) being run is `static-lite` could this also be run from the directory itself using...

```
govuk-docker-run -e RAILS_ENV=test bundle exec rake jasmine:ci
```

This doesn't really matter, its more to do with how govuk-docker works. Feel free to ignore if you don't think it's worth it.

app/views/root/400.html.erb
```diff
@@ -1,5 +1,5 @@
 <%= render partial: "error_page", locals: {
     heading: "Sorry, there is a problem",
     intro: '<p class="govuk-body">Go back and change the information you entered.</p>',
-    status_code: 400
+    status_code: 400,
```
owenatgov, line 4: I was wondering why there were so many changed files in this PR 😂 Nice one 👍

app/views/root/400.html.erb
```diff
@@ -1,5 +1,5 @@
 <%= render partial: "error_page", locals: {
     heading: "Sorry, there is a problem",
```
owenatgov, line 2: Maybe out of scope for this PR but there's a lot of hard coded content in these templates. Could we translate it? I don't know if this content would ever actually be translated, I'm picturing a situation where we implement something to set locale based on user info or whatever. It's more for good practice. At the very least it'll handle a few instances of repeated content. I don't even know if static supports translations.

owenatgov (commented):
Some finicky comments. Happy with this otherwise 👍

app/views/root/400.html.erb
injms: I agree, there's a lot of content that could be translated here. I can't see `rails-i18n` or any localisation files in Static, so I think it'd need to be set up from scratch. With that in mind, setting up translations is outside of the scope of this pull request / should be a thing that is worthy of its own pull request.

I've raised this an an issue so we don't forget: https://github.com/alphagov/static/issues/2448

app/views/root/400.html.erb
injms: I do love me a dangling comma. Better diffs ftw!

README.md
injms: Hmm, I get an error when I run the shortcut version - I think the the shortcut runs:

```sh
govuk-docker run -e static-lite RAILS_ENV=test bundle exec rake jasmine:ci
```

which throws because `-e` should be before `static-lite` - so `-e` gets interpreted as an executable file:

```sh
Error response from daemon: OCI runtime create failed: container_linux.go:370: starting container process caused: exec: "-e": executable file not found in $PATH: unknown
```

I'd like to leave this as is rather than spend time figuring out a way of making the shortcut work correctly.

app/views/root/400.html.erb
owenatgov: That's perfectly reasonable. Is there a way we can at least make some of this content non-repeatable, maybe in a little partial or something? It's not a huge deal but _just in case_ it has to change suddenly we can update it once.

Not a blocker for merging. Just blasting opinions.

owenatgov (approved):
I am content with this. One final comment but otherwise I reckon this can go out (once it's ready to go out)

app/views/root/400.html.erb
injms: I've had a go at this and I think it ends up convoluting things - the template ends up calling a partial, which calls another partial - and then the middle partial only sometimes overwrites the variables inside. It feels like premature optimisation that'd need to be undone when translations are used.

I think your first solution of using translations would be the best, simplest, and most future proof method to solve this - and that's outside the scope of this pull request.

PR #2455, opened 2021-04-09
Hide Brexit global bar

bilbof:
We are adding a new global banner.
This hides the Brexit link to reduce the size of the global banner.

https://docs.publishing.service.gov.uk/manual/emergency-publishing.html#2-hide-the-brexit-link-from-the-global-bar

### DIFF
--- a/app/views/components/_global_bar.html.erb
+++ b/app/views/components/_global_bar.html.erb
@@ -11,7 +11,7 @@
   coronavirus_href = "/coronavirus"
   coronavirus_subtext = "Rules, guidance and support"
 
-  show_transition_link = true
+  show_transition_link = false
   transition_title = "Brexit"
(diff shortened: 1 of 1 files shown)


brucebolt (approved):

app/views/components/_global_bar.html.erb
```diff
@@ -11,7 +11,7 @@
   coronavirus_href = "/coronavirus"
   coronavirus_subtext = "Rules, guidance and support"
 
-  show_transition_link = true
+  show_transition_link = false
```
leenagupte, line 14: should the COVID banner be hidden as well?

bilbof: Good question, we checked this with SMT and only the brexit banner should be removed.

PR #2462, opened 2021-04-16
WIP - add new Explore megamenu

maxf:
The Explore team are working on a new drop-down megamenu off the govuk-header bar.
This branch adds it.


alex-ju (commented):
Aware this is a draft PR, but since it's visible I'd like to highlight a few things that might help us save some time in the long run:
- we are in the process of migrating away from `core_layout` and `header_footer_only` to the new, Design System-based `gem_layout` (so this new component should be introduced as an alternative navigation in the [public layout](https://components.publishing.service.gov.uk/component-guide/layout_for_public))
- `govuk_template` is a deprecated dependency and we plan to remove it as soon as possible – the Design System team is preparing to drop compatibility mode in the next major release (this means we will be removing `govuk_template` in the very near future)
- we are migrating away from jQuery to plain/vanilla JS, so we try to avoid introducing new jQuery code
- any new UI components we introduce would need to follow the [agreed principles and conventions](https://github.com/alphagov/govuk_publishing_components/blob/master/docs/develop-component.md)

Hopefully, this might help the team steer this stream of work in the right direction. Feel free to raise any issues/concerns in the `govuk-frontenders` meeting if things don't make sense.

chao-xian:
This work has since been superseded. The header is now in the gem.

PR #2476, opened 2021-05-11
WIP - Navigation stuff

DilwoarH:


app/views/components/header/_header.html.erb
```diff
@@ -0,0 +1,41 @@
+<header role="banner" id="global-header" class="<%= yield(:header_class) %>">
+  <div class="header-wrapper">
+    <div class="header-global">
+      <div class="header-logo">
+        <a href="<%= content_for?(:homepage_url) ? yield(:homepage_url) : "https://www.gov.uk/" %>"
```
injms, line 58: In this context, is the `:homepage_url` and `https://www.gov.uk/` the same thing?

injms: Hmm, seems to be used in the previous header - so maybe we should leave this be.

DilwoarH:
superseded by  https://github.com/alphagov/static/pull/2520

PR #2486, opened 2021-06-01
Add super navigation bar

DilwoarH:
## What

Create the base super navigation bar component. This will introduce the three links: Topics, Departments, Government activities.

It should be hidden via a flag so that it doesn't get deployed to production.

## Why

This will help us to split up the work for the menu bar iteration and make code review easier.

## Next steps

The next step after this PR is merged is to create the drop-down menu. It is important that we do this iteratively as this will make things easier to understand and the iterations are not blocking each other. The PR is the biggest iteration as this sets the foundations of the navigation bar.

## Other things to consider

I have tried to follow the [GOVUK Design system](https://design-system.service.gov.uk/components/header/) standards as much as possible but as this design is completely new - there had to be some iterations/overrides made.

## Trello

https://trello.com/c/jGzHyK2B/335-create-base-super-navigation-component

---

### Desktop

Screenshot 2021-06-01 at 01 05 13

### Mobile collapsed

Screenshot 2021-06-01 at 01 05 26

### Mobile expanded

Screenshot 2021-06-01 at 01 05 38

### No JS

*BROKEN*

Screenshot 2021-06-01 at 01 05 26

### Desktop

Screenshot 2021-06-01 at 01 09 16

### Mobile collapsed

Screenshot 2021-06-01 at 01 09 44

### Mobile expanded

Screenshot 2021-06-01 at 01 09 57

### No JS

Screenshot 2021-06-01 at 01 10 20

## Accessibility Tree

Jun-01-2021 01-13-41


chao-xian (commented):
Looks ok so far but I've got one comment, and will let @injms take a look.

app/views/components/super-navigation-header/_super-navigation-header.html.erb
```diff
@@ -1,3 +1,7 @@
+<%
+  navigation_links = YAML.load(File.read(File.join(File.dirname(__FILE__),"/navigation-links.yml")))
```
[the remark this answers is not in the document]
DilwoarH: Happy for this to happen, I was unsure about what use of controllers was in this app as there wasn't much use already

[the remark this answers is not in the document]
chao-xian: Let's split this into a separate piece of work to use the translations API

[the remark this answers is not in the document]
injms: If the rails-i18n gem is used, then would this be a case of using `t()` methods?

app/assets/stylesheets/components/_super-navigation-header.scss
```diff
@@ -0,0 +1,90 @@
...
+
+.govuk-header__super-navigation-item-link-text--search {
+  @include govuk-media-query($from: $breakpoint) {
+    display: block;
+    text-indent: -5000px;
+    overflow: hidden;
+  }
+}
+
+.govuk-header__super-navigation-menu-button {
```
injms, line 83: Above the 640px breakpoint, do you think that the menu button text looks a bit mismatched with the other text?

image

app/assets/stylesheets/components/_super-navigation-header.scss
```diff
@@ -0,0 +1,90 @@
...
+  }
+}
+
+.govuk-header__super-navigation-item-link-text--search {
+  @include govuk-media-query($from: $breakpoint) {
+    display: block;
+    text-indent: -5000px;
+    overflow: hidden;
+  }
+}
```
injms, line 81: I think that this would be better off using either the mixin or the styles from [`govuk-visually-hidden`](https://github.com/alphagov/govuk-frontend/blob/4b5ed0ada3b64722ae2074dfa0db5d4d3a45cb79/src/govuk/helpers/_visually-hidden.scss#L16) as it's a better way of making things available to screenreaders.

app/views/components/super-navigation-header/_header-logo.html.erb
```diff
@@ -0,0 +1,33 @@
+<div class="govuk-header__logo govuk-header__super-navigation-header-logo">
+  <a href="<%= content_for?(:homepage_url) ? yield(:homepage_url) : "https://www.gov.uk/" %>"
+      title="<%= content_for?(:logo_link_title) ? yield(:logo_link_title) : "Go to the GOV.UK homepage" %>"
+      class="content govuk-header__link govuk-header__link--homepage"
+      id="logo"
```
injms, line 5: Once the smokey test has been updated, we can get rid of this `id="logo"` as it's not used for anything.

app/assets/stylesheets/components/_super-navigation-header.scss
```diff
@@ -0,0 +1,90 @@
...
+
+.govuk-header__super-navigation-item-link {
+  padding: govuk-spacing(3) + 2;
+  width: 100%;
+
+  @include govuk-media-query($from: $breakpoint) {
+    width: auto;
+  }
+}
```
injms, line 46: The new accordion component from the gem had some feedback from the Design System team that the focus state was overly large:

> The focus state taking up the entire button is larger than necessary and is likely to be distracting for users

https://github.com/alphagov/govuk-design-system-backlog/issues/1#issuecomment-839802842

I wonder if the same thing could apply for the focus state and the clickable area for this menu when seen on small and medium screen sizes?

image

app/assets/stylesheets/components/_super-navigation-header.scss
```diff
@@ -0,0 +1,90 @@
+$breakpoint: 769px;
+
+// Remove bottom borders
+.govuk-header__super-navigation,
```
injms, line 4: The `govuk` prefix is the namespace for the Design System - anything in Static should be using the prefix `app-`.
```suggestion
.app-header__super-navigation,
```

app/views/components/super-navigation-header/_header-content.html.erb
```diff
@@ -0,0 +1,21 @@
...
+          <a class="govuk-header__link govuk-header__super-navigation-item-link" href="<%= link["href"] %>">
+            <%= link["label"] %>
+          </a>
+        </li>
+      <% end %>
+      <li class="govuk-header__navigation-item govuk-header__super-navigation-item govuk-header__super-navigation-item--search">
+          <a class="govuk-header__link govuk-header__super-navigation-item-link govuk-header__super-navigation-item-link--search" href="/search">
+            <span class="govuk-header__super-navigation-item-link-text--search">Search</span>
+          </a>
+        </li>
```
injms, line 18: There's some space characters between each list item that's causing each link to have a slither of space between each of them - and frustatingly using `<%-` doesn't seem to be working here.

I mention it because I think that this looks to be on purpose, but the CSS says otherwise; and the space between "Government activites" and "Search" looks as if it's broken because of the gap on the bottom right corner:

image

chao-xian:
@injms thank you for your review!

app/assets/stylesheets/components/_super-navigation-header.scss
DilwoarH: Fixed

app/assets/stylesheets/components/_super-navigation-header.scss
DilwoarH: Tried that - didn't work because it has absolute positioning

app/assets/stylesheets/components/_super-navigation-header.scss
DilwoarH: It's part of the mobile styling, the clickable area is much larger

app/assets/stylesheets/components/_super-navigation-header.scss
DilwoarH: Updated

app/views/components/super-navigation-header/_header-content.html.erb
DilwoarH: Fixed

DilwoarH:
@injms @chao-xian - ready for another review

injms (approved):
This looks good to me 🚢 🚀  Nice work!

DilwoarH:
> This looks good to me 🚢 🚀 Nice work!

Thank you ! ⭐

PR #2495, opened 2021-06-05
Add routing constraints to restrict icon formats

pixeltrix:
Previously the icon redirect routes would accept any format and be recognised. Since this path is passed directly into `asset_path` to find the destination digest path for the redirect it seems advisable to only allow the expected types.

The paths could've been set to `format: false` but this leaves them being recognised as the default type (text/html) which could have unintended side effects.

### DIFF
--- a/config/initializers/mime_types.rb
+++ b/config/initializers/mime_types.rb
@@ -2,3 +2,5 @@
 
 # Add new mime types for use in respond_to blocks:
 # Mime::Type.register "text/richtext", :rtf
+
+Mime::Type.register "image/x-icon", :ico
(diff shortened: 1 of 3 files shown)


pixeltrix:
I'm not normally looking at GOV.UK repos on a Saturday but I was looking at what you're doing for `/favicon.ico` as I can't get CloudFront to compress an `image/x-icon` correctly. Whilst looking at the implementation I noticed an issue with the defined routes and thought best to contribute back any changes I'd be making.

I think the Fastly config prevents the additional file extension but if you request `/favicon.ico.txt` locally you end up with a `Sprockets::Rails::Helper::AssetNotFound` exception. I'm pretty sure it's not exploitable but I thought it best that a user controlled path doesn't get sent into that code since we've had path traversal exploits in Sprockets before: https://blog.heroku.com/rails-asset-pipeline-vulnerability - though that one needed `assets.compile = true` in production.

PR #2513, opened 2021-06-16
Update Account links in the `gem_layout` template

injms:
## What

Update the Account sign in / sign out / account links in the header section of the `gem_layout` template to match the links in `core_layout` template.

## Why

The links weren't updated when `core_layout` was updated.

## Visual changes

None.


hannako (approved):
:LGTM:

PR #2514, opened 2021-06-16
Add and update survey styles to gem layout template

injms:
## What

Adds to the `gem_layout` template, as the styles were missing but the ability to show surveys was not.

## Why

The survey banner was being shown without any styles and looked quite broken.

## Visual differences
Before:

image

After:

image


richardTowers (approved):
Looks sensible to me.

PR #2515, opened 2021-06-16
Add explore_header templates

chao-xian:
This will allow us to switch between the current header menu and the newer design that we want to AB test with. This means that frontend apps using the AB Test gem can request a different template via Slimmer.

See https://github.com/alphagov/government-frontend/pull/2156 for example

Trello https://trello.com/c/LUhrajXF/323-prepare-the-new-menu-a-b-test-for-sections-of-govuk


chao-xian:
This now needs to be amended to accommodate https://github.com/alphagov/frontend/pull/2775

injms (approved):

PR #2524, opened 2021-06-21
Add account pages to global bar BlockList

danacotoran:
The Accounts team are moving the account homepage to the `frontend` app, with the plan of transitioning other parts of the account manager to be rendered by `frontend`, as we retire the `govuk-account-manager-prototype` app. 

Seeing as the `frontend` app uses `static`, the global bar (currently Coronavirus bar) is automatically added to the new `/account/home` page. The design for the account manager excludes this additional bar currently.

This adds paths beginning with `/account/` (assuming that any future account-related pages will live under `/account`) to the `urlBlockList` in the global bar script.

### Before (bar)
Screenshot 2021-06-21 at 16 52 48

### After (no bar)
Screenshot 2021-06-21 at 16 56 02

------
https://trello.com/c/lIDQNATi/813-implement-the-govuk-account-home-page

### DIFF
--- a/app/assets/javascripts/global-bar-init.js
+++ b/app/assets/javascripts/global-bar-init.js
@@ -30,7 +30,8 @@ var globalBarInit = {
       '^/coronavirus/.*$',
       '^/brexit(.cy)?$',
       '^/transition-check/.*$',
-      '^/eubusiness(\\..*)?$'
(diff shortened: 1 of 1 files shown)


huwd (approved):
makes sense to me!

PR #2527, opened 2021-06-23
Add gem layout template variation for accounts

danacotoran:
The Accounts team are [in the process of](https://github.com/alphagov/frontend/compare/msw/account-home-page) moving over some of the account pages to the `frontend` app. 

Since `frontend` app uses `static`, and certain aspects of the account manager template differ greatly from the default `gem_layout`, this warrants the introduction of a new template for use on accounts related pages (thanks @injms for suggesting this approach).
Some main differences are:
- the Account home link in the header should link to the account homepage (as opposed to linking to the GOVUK homepage)
- the Account product name should be present in the header
- there should be an account nav in the header
- the feedback component at the bottom of the page should be hidden as the account has its own version of a feedback component/link (the **Help improve GOV.UK accounts** box at the bottom of [this page](https://www.account.publishing.service.gov.uk/sign-in))

----

**Important**: depends on https://github.com/alphagov/govuk_publishing_components/pull/2160

-----

https://trello.com/c/lIDQNATi/813-implement-the-govuk-account-home-page

### DIFF
--- a/docs/slimmer_templates.md
+++ b/docs/slimmer_templates.md
@@ -10,6 +10,12 @@ This is the same as the `gem_layout` template, except that this layout **doesn't
 
 Use this layout if you want to have full width content - such as the blue welcome bar on the GOV.UK homepage.
 
+## `gem_layout_account`
+
+This is a variation on the `gem_layout` template intended to be used on GOV.UK Account pages. This layout **doesn't** include the feedback component as the account pages use a different one from the rest of GOV.UK.
+
+This also includes the Account product name in the layout header and changes the logo link to the account homepage link.
+
 ## `core_layout` (default)
 
 This template contains styles for the black header bar, the footer and core layout classes. By default it will centre your content to the same width as the GOV.UK header. It provides classes for grid layouts using the column mixins from the frontend toolkit.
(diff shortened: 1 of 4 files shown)


docs/slimmer_templates.md
```diff
@@ -10,6 +10,12 @@ This is the same as the `gem_layout` template, except that this layout **doesn't
 
 Use this layout if you want to have full width content - such as the blue welcome bar on the GOV.UK homepage.
 
+## `gem_layout_account`
+
+This is a variation on the `gem_layout` template intended to be used on GOV.UK Account pages. This layout **doesn't** include the feedback component as the account pages use a different one from the rest of GOV.UK.
```
andysellick, line 15: You've probably already had this conversation, but I wonder if now the accounts stuff is going to be part of `frontend`, why not use the existing feedback mechanism instead of having a whole separate one just for accounts?

andysellick (approved):
Seems to make sense 👍

docs/slimmer_templates.md
danacotoran: Yes you're completely right: for now we're trying to have this new account home page in `frontend` match the manage/security section layout as closely as possible, but in the longer term we definitely need to look at integrating account feedback into the existing feedback system.

PR #2540, opened 2021-07-06
Add explore header templates for public_layout in gem

chao-xian:
For the apps that are on the gem based public_layout template (currently Frontend and Collections) we need to be able to specify which header we want to show the user. Accompanying PR here for the Explore Super Menu being added to the gem is https://github.com/alphagov/govuk_publishing_components/pull/2179/

Trello https://trello.com/c/fYNA3mnu/390-prep-ab-test-for-frontend-and-collections

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -8,6 +8,7 @@
 
 //= require govuk_publishing_components/components/cookie-banner
 //= require govuk_publishing_components/components/layout-header
+//= require govuk_publishing_components/components/layout-super-navigation-header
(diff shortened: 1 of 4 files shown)


alex-ju (approved):
This looks good, pending the mentioned change in `govuk_publishing_components`. I'm slightly worried about the number of gem_layouts we ended up with, but at least they're sharing most of the code, for now.

chao-xian:
@alex-ju thank you - yes I agree but hopefully we can get rid of the "explore" variants later on and have the new header in the current gem layout.

PR #2546, opened 2021-07-08
Remove Explore Super Menu Header local component

chao-xian:
## What

Removes local component for new header and switches to version in `govuk_publishing_components`.

## Why

The component was ported to `govuk_publishing_components` in https://github.com/alphagov/govuk_publishing_components/pull/2179 and the local one isn't being used in testing.

## Visual changes

None


injms (approved):
🦾 🚀

PR #2572, opened 2021-08-10
Remove redundant JavaScript import

alex-ju:
### What
Remove redundant `govuk_publishing_components/lib/header-navigation` import

### Why
`header-navigation` is already required and executed by all public-facing applications as part of `govuk_publishing_components/lib`. Given `header-navigation` is not an actual module, but a IIFE relict from `govuk_template`, this duplication results in two toggle-like events being attached to the same button, which causes the search button to open and close itself in one click

Note

- [ ] requires [`frontend` to be brought in line with the other apps](https://github.com/alphagov/frontend/pull/2901) using the `gem_layout` (namely `collections`, `email-alert-frontend`, `feedback`, `info-frontend`, `licence-finder` and `smart-answers`).

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -1,7 +1,6 @@
 //= require govuk_publishing_components/dependencies
 
 //= require govuk_publishing_components/lib/cookie-functions
-//= require govuk_publishing_components/lib/header-navigation
 
 //= require analytics
 


kevindew:
Nice work catching this. The frontend app I looked at for this was frontend which [doesn't seem to have an import of govuk_publishing_components/lib](https://github.com/alphagov/frontend/blob/main/app/assets/javascripts/application.js) - so I did wonder if we might be trading fixing this in one app to break it another?

alex-ju:
> Nice work catching this. The frontend app I looked at for this was frontend which [doesn't seem to have an import of govuk_publishing_components/lib](https://github.com/alphagov/frontend/blob/main/app/assets/javascripts/application.js) - so I did wonder if we might be trading fixing this in one app to break it another?

Indeed – I prepared https://github.com/alphagov/frontend/pull/2901 to bring `frontend` in line with the other apps.

kevindew (approved):
Cool - so long as you're happy all the frontend apps are covered 👍

PR #2577, opened 2021-08-18
Bump govuk_publishing_components

kevindew:
Trello: https://trello.com/c/pATsfsuD/524-convert-finder-frontend-module-initialisers-in-livesearchjs-to-not-use-jquery

I've expedited this release because https://github.com/alphagov/finder-frontend/pull/2617 requires this change and I'd forgotten static needs to be also updated.


kevindew:
Actually, I got this wrong. This is needed to be expedited after all, so I'll close it and let dependabot do it's thing.

PR #2588, opened 2021-09-03
Add `gem_layout_header_navigation` template

alex-ju:
### What

Add `gem_layout_header_navigation` template

### Why

A set of applications [1][2][3] require a header navigation (sometimes referred to as 'proposition menu') which is only available in `core_layout` at the moment.

### Checklist

- [ ] Dependent on https://github.com/alphagov/govuk_publishing_components/pull/2292

[1] https://github.com/alphagov/whitehall/pull/6240
[2] https://github.com/alphagov/finder-frontend/pull/2616
[3] https://github.com/alphagov/government-frontend/pull/2206

### DIFF
--- a/app/controllers/root_controller.rb
+++ b/app/controllers/root_controller.rb
@@ -28,6 +28,7 @@ class RootController < ApplicationController
     gem_layout_explore_header
     gem_layout_full_width
     gem_layout_full_width_explore_header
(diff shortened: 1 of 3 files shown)


alex-ju:
Converting this to a draft as the current header (and the associated navigation) may not be around for long.

PR #2590, opened 2021-09-09
Set the navigation header as default

injms:
## What

Updates both the gem layout and the core layout templates to use the new navigation header as the default.

Turns the navigation header off for the account manager template.

Adds a layout with the previous header.

Card: https://trello.com/c/3Vu0IPhG

## Why
...

## Visual differences

Gem layout:

| Before | After |
| - | - |
| image | image |

Layout with previous header:

### DIFF
--- a/app/controllers/root_controller.rb
+++ b/app/controllers/root_controller.rb
@@ -24,14 +24,15 @@ class RootController < ApplicationController
     504
     campaign
     gem_layout
-    gem_layout_old_header
     gem_layout_account_manager
     gem_layout_explore_header
     gem_layout_full_width
-    gem_layout_full_width_old_header
     gem_layout_full_width_explore_header
+    gem_layout_full_width_old_header
     gem_layout_no_feedback_form
     gem_layout_no_footer_navigation
+    gem_layout_old_header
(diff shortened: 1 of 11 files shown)


chao-xian:
We'll need a `header_footer_only` template that has the old header too for `/transition-check` - just discovered this in finder-frontend running this branch against it.

chao-xian:
I've split out some of this here: https://github.com/alphagov/static/pull/2595

app/controllers/root_controller.rb
```diff
@@ -32,6 +32,8 @@ class RootController < ApplicationController
     gem_layout_full_width_explore_header
     gem_layout_no_feedback_form
     gem_layout_no_footer_navigation
+    gem_layout_old_header
```
chao-xian, line 35: I think my changes [here](https://github.com/alphagov/static/pull/2595/commits/0525c73d4b95a1948c9428c8edb800fb0000a3a3)  mean this would become duplicates?

injms: Duplicate removed and the templates alphabetised 👍

chao-xian (approved):

PR #2595, opened 2021-09-17
Add templates with old style header

chao-xian:
These are being added as an interim step before we set the Explore Super Menu Header to be the default. We need the old header to ensure we keep Account links in place.

### DIFF
--- a/app/views/root/gem_layout_full_width_old_header.html.erb
+++ b/app/views/root/gem_layout_full_width_old_header.html.erb
@@ -1,0 +1,4 @@
+<%= render partial: 'gem_base', locals: {
+  full_width: true,
+  show_explore_header: false,
+} %>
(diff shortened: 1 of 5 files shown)


app/views/root/gem_layout_full_width_old_header.html.erb
```diff
@@ -0,0 +1,4 @@
+<%= render partial: 'gem_base', locals: {
+  full_width: true,
+  show_explore_header: false,
```
injms, line 3: Just so we're aware - setting any parameter to `false` that uses `||=` to assign in `gem_base` template will get overwritten by the default.

So when `show_explore_header` gets set to true by default in the `gem_base` template, the template needs to be updated so it doesn't overwrite this parameter when `false` is passed in.

injms (approved):
🔀 👍

app/views/root/gem_layout_full_width_old_header.html.erb
chao-xian: 👍

PR #2649, opened 2021-11-10
Add print stylesheet for navigation header

injms:
## What

Adds the print stylesheet for the navigation header

## Why

The navigation header has a print stylesheet, so it should be used. Currently  the print stylesheet stops the navigation header from being printed as it takes up too much space and isn't that useful.

## Visual differences

| Browser | Before | After |
| - | ------ | ----- 
| Firefox |  |  |
| Chrome |  |  | 
| Safari |  |  |
| Internet Explorer 11 | Screenshot 2021-11-10 at 12 05 30 | Screenshot 2021-11-10 at 12 05 03 |

### DIFF
--- a/app/assets/stylesheets/application-print.scss
+++ b/app/assets/stylesheets/application-print.scss
@@ -5,5 +5,6 @@ $is-print: true;
 @import "govuk_publishing_components/component_support";
 
 @import "govuk_publishing_components/components/print/button";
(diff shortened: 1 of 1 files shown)


owenatgov (approved):
🖨️

chao-xian (approved):

PR #2658, opened 2021-11-23
Clear the template caches when the emergency banner is updated

bilbof:
Rather than requiring a separate [script](https://github.com/alphagov/govuk-puppet/blob/178ff3c0d1b024a5e0c81f384b07d0dfb65f418d/modules/govuk_jenkins/templates/jobs/clear_template_cache.yaml.erb#L18) run by Jenkins to clear the caches, Static can clear it's own cache using the same [ActionPack page caching API](https://github.com/rails/actionpack-page_caching): 

This ensures that this process is somewhat tested (as compared to the Jenkins script, which is hard to test).

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -22,6 +22,8 @@ module Static
     # -- all .rb files in that directory are automatically loaded after loading
     # the framework and any gems in your application.
 
+    config.eager_load_paths += %W[#{config.root}/lib]
+
     # Using a sass css compressor causes a scss file to be processed twice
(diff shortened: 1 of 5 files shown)


bilbof:
The ERB files could be generated and placed in the `/public` directory directly, rather than involving the ActionPack page cache API. Generating the ERB files statically would enable us to move away from serving these files with Nginx though any alternative such as using S3 must consider local development. This may make testing the ERB templates using Minitest trickier.

bilbof:
This would need to be run on all nodes, so not suitable for a Rake task.

PR #2659, opened 2021-11-24
Clear the template caches every 5 minutes

bilbof:
Rather than requiring a separate [script][1] run by Jenkins to clear the caches, Static can clear its own cache using
the same ActionPack page caching [API](https://github.com/rails/actionpack-page_caching).

This ensures that this process is tested (as compared to the Jenkins script, which is hard to test).

It also ensures that the template caches expire - this design is similar to the 'sweepers' recommended in the ActionPack page cache docs.

We may still need to cycle all Static pods when the emergency banner is added/removed once we are running Static in Kubernetes. We could run a sidecar container that clears Static caches, but that's out of scope for this work.

Longer term we probably want to look at replacing Static or at least making the serving less dynamic - there's no need for a Ruby webserver to be involved since it's Nginx that mostly is serving these files.

A corresponding change to run `bundle exec whenever` in govuk-app-deployment is required to update the crontab.

[1]: https://github.com/alphagov/govuk-puppet/blob/178ff3c0d1b024a5e0c81f384b07d0dfb65f418d/modules/govuk_jenkins/templates/jobs/clear_template_cache.yaml.erb#L18

To be merged once https://github.com/alphagov/govuk-rfcs/pull/144 is approved.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -14,6 +14,7 @@ gem "rack_strip_client_ip"
 gem "redis"
 gem "sassc-rails"
 gem "uglifier"
+gem "whenever"
 
 group :development do
   gem "better_errors"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -82,6 +82,7 @@ GEM
       regexp_parser (>= 1.5, < 3.0)
       xpath (~> 3.2)
     childprocess (3.0.0)
+    chronic (0.10.2)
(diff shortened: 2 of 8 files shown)


bilbof:
This spike was superseded by https://github.com/alphagov/govuk-puppet/pull/11326.

PR #2663, opened 2021-11-29
Remove unused layouts and `govuk_template`

alex-ju:
### What
As all frontend applications are now using `gem_layout` or a variation of this, we can now remove unused layouts (`core_layout` and `header_footer_only`) and 'non-layout templates' (`campaign` and `proposition_menu`) and make the `gem_layout` default.

This also means we can remove all the assets that are only being used within these templates – and more importantly drop the dependency on [govuk_template](https://github.com/alphagov/govuk_template). However, a few assets from `govuk_template` (`apple-touch-icon*`) need to be ported.

I have also tested this change manually in integration by accessing a range of pages rendered by various frontend applications while keeping an eye on logs.

### Why
To keep our codebase clean and prevent deprecated layouts from being used.


alex-ju:
Rebased to resolve a doc conflict after https://github.com/alphagov/static/pull/2662

app/views/root/chromeless.html.erb
```diff
@@ -1,20 +1 @@
...
-  </div>
-<% end %>
-
-<% content_for :footer_support_links do %>
-  <%= render :partial => 'footer_support_links' %>
-<% end %>
-<% content_for :body_end do %>
-  <%= render :partial => 'javascript', :locals => { :js_file => 'header-footer-only' } %>
-<% end %>
+<%= render partial: "gem_base" %>
```
injms, line 1: Is showing both the header and footer wanted here? Prior to this change both were left out for this particular template.

docs/slimmer_templates.md
```diff
@@ -1,6 +1,6 @@
 # Slimmer Templates
 
-## `gem_layout`
+## `gem_layout` (default)
```
injms, line 3: 🎉

app/views/root/chromeless.html.erb
injms: Super - thanks for explaining. Makes sense; and if there is any special work I agree it'd best be done outside of this pull request.

injms (approved):
Super work - thank you!

PR #2682, opened 2022-01-07
Add missing textarea styles for the feedback component

alex-ju:
Missed in https://github.com/alphagov/static/pull/2680

<table>
<tr><th>Before</th><th>After</th></tr>
<tr><td valign="top">

Screenshot 2022-01-07 at 11 31 55

</td><td valign="top">

Screenshot 2022-01-07 at 11 34 44

</td></tr>
</table


andysellick (approved):
Makes sense, but it seems like this only happens for pages rendered by some applications - where is this problem occurring?

alex-ju:
> Makes sense, but it seems like this only happens for pages rendered by some applications - where is this problem occurring?

When the application (e.g. frontend) imports the styles there's no problem, but some app (e.g. collections) don't have textareas – hence the problem.

PR #2683, opened 2022-01-10
Remove jQuery from cross domain tracking

andysellick:
## What
Removes jQuery from the cross domain tracking script, and updates the code to use the GOVUK Frontend standard for module structure i.e. from a `start` to an `init` structure.

## Why
We're removing jQuery from GOV.UK.

## Visual changes
None.

### DIFF
--- a/app/assets/javascripts/modules/cross-domain-tracking.js
+++ b/app/assets/javascripts/modules/cross-domain-tracking.js
@@ -1,47 +1,55 @@
+window.GOVUK = window.GOVUK || {}
 window.GOVUK.Modules = window.GOVUK.Modules || {};
 
-(function (Modules, global) {
-  'use strict'
+(function (Modules) {
+  function CrossDomainTracking ($module) {
+    this.$module = $module
+    Modules.crossDomainLinkedTrackers = Modules.crossDomainLinkedTrackers || []
+  }
 
-  var $ = global.$
-
-  Modules.crossDomainLinkedTrackers = []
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/modules/cross-domain-tracking.js
```diff
@@ -1,47 +1,55 @@
...
-  var $ = global.$
+(function (Modules) {
+  function CrossDomainTracking ($module) {
+    this.$module = $module
+    Modules.crossDomainLinkedTrackers = Modules.crossDomainLinkedTrackers || []
+  }
 
-  Modules.crossDomainLinkedTrackers = []
+  CrossDomainTracking.prototype.init = function () {
+    if (this.isTrackable(this.$module)) {
```
alex-ju, line 11: non-blocker: alternatively, we could have used `bind` to preserve the instance (and be able to call `this.$module` from) within the `isTrackable` function (similarly for `addLinkedTrackerDomain`)

alex-ju (approved):
Looks good!

app/assets/javascripts/modules/cross-domain-tracking.js
andysellick: Both `isTrackable` and `addLinkedTrackerDomain` are called later in the code and pass different elements. I considered trying something in order to not have to pass things, but decided that was more in the realm of a refactor than the scope of this PR.

PR #2685, opened 2022-01-12
Remove unneeded navigation header files

injms:
## What

Removes unneeded CSS and JavaScript files that were used for the navigation header before it was moved to the components gem[^1].

## Why

Because they're not needed here and could cause confusion.

## Visual changes

None.

[^1]: https://components.publishing.service.gov.uk/component-guide/layout_super_navigation_header


andysellick (approved):
👏

PR #2687, opened 2022-01-13
Remove jQuery from globalbar

andysellick:
## What
Removes jQuery from the global bar script, and updates the code to use the GOVUK Frontend standard for module structure i.e. from a `start` to an `init` structure.

## Why
We're removing jQuery from GOV.UK.

## Visual changes
None.

### DIFF
--- a/app/assets/javascripts/modules/global-bar.js
+++ b/app/assets/javascripts/modules/global-bar.js
@@ -13,8 +13,12 @@
 
   Modules.GlobalBar = function () {
     this.start = function ($el) {
+      this.$module = $el[0]
       var GLOBAL_BAR_SEEN_COOKIE = 'global_bar_seen'
-      var alwaysOn = $el.data('global-bar-permanent')
+      var alwaysOn = this.$module.getAttribute('data-global-bar-permanent')
+      if (alwaysOn === 'false') {
+        alwaysOn = false // in this situation we need to convert string to boolean
+      }
       var cookieCategory = GOVUK.getCookieCategory(GLOBAL_BAR_SEEN_COOKIE)
       var cookieConsent = GOVUK.getConsentCookie()[cookieCategory]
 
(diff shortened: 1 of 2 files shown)


spec/javascripts/modules/global-bar.spec.js
```diff
@@ -107,23 +114,30 @@ describe('Global bar module', function () {
...
     it('tracks when clicking on a link marked with js-call-to-action', function () {
-      globalBar.start(element)
+      globalBar = new GOVUK.Modules.GlobalBar(element[0])
+      globalBar.init()
 
-      $(element).find('.js-call-to-action').click()
+      var link = $(element).find('.js-call-to-action')[0]
+      link.addEventListener('click', function (e) {
+        e.preventDefault()
+      })
```
kevindew, line 132: It wasn't clear to me why we need this event listener when we didn't before - perhaps that's something for the commit message?

kevindew (commented):
Looks great @andysellick - just a couple of small comments

spec/javascripts/modules/global-bar.spec.js
andysellick: That's an excellent question. Honestly I don't know. I've just tested the original code to see the difference and I think it's something to do with the difference between calling `click` on a jQuery object compared with a JS DOM node. 

If you run this test in the browser (with the original code, not my changes) it leaves the element in question in the viewport. Calling `$('.js-call-to-action').click()` manually in the console triggers the relevant code but doesn't do anything else (expected). Calling `document.querySelector('.js-call-to-action').click()` (the vanilla JS equivalent) triggers the code but also navigates to the URL set in the href attribute for that link. So it seems that there's definitely a difference, but I'm not sure why this happens.

Reading the code my understanding is that this link should behave as a normal link - the only thing that the JS should do is fire a tracking event when it is clicked, so the logical thing for the test to do is to artificially prevent that navigation from happening. I don't know how the original code allowed that to be tested without causing a navigation to happen, but it seems the difference between the jQuery and native JS `click` events may account for it.

app/assets/javascripts/modules/global-bar.js
```diff
@@ -8,84 +8,104 @@
...
       }
+    })
 
-      function handleCallToActionClick () {
-        var $link = $(this)
-        var url = $link.attr('href')
-        track(url)
+    // if the element is visible
+    if (this.$module.offsetParent !== null) {
+      if (!alwaysOn) {
```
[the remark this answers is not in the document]
andysellick: Good point, changed.

app/assets/javascripts/modules/global-bar.js
```diff
@@ -8,84 +8,104 @@
...
-        var cookieVersion = currentCookieVersion
+    function handleCallToActionClick (target) {
+      var url = target.getAttribute('href')
+      track(url)
+    }
 
-        if (currentCookie) {
-          cookieVersion = currentCookie.version
-        }
+    function hide (evt) {
```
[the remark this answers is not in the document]
andysellick: Reasonable, have changed.

spec/javascripts/modules/global-bar.spec.js
kevindew: Ah yes, I gave it a spin locally and the jQuery equivalent only triggers the event handlers and not the default action. I didn't realise that's how jQuery `.click` worked I thought it was the same and you needed to do `.trigger` for just event handlers.

Though looking [at the docs](https://api.jquery.com/click/#click-handler) they make it clear it's a shorthand for `.trigger('click')`

kevindew (approved):

PR #2693, opened 2022-01-24
Fix intermittent test failures in surveys

andysellick:
## What
Fix two intermittently failing tests and re-enable and fix two disabled (and also intermittently failing) tests.

### Failures
Two of the survey JavaScript tests have been failing intermittently. This seems to have been caused by the state of the `surveyExpanded` attribute within the surveys code. This attribute is set correctly or incorrectly by other tests, so the failure would occur depending on the order the tests are run.

Specifically, this occurs in the `if` statement on line 244, where matching the `if` instead of the `else` can fire the wrong tracking event. The solution is to manually set this attribute to the expected value in the failing tests.

Issue noted in issue https://github.com/alphagov/govuk_publishing_components/issues/2524

### Disabled tests
Once enabled, these tests were also failing intermittently, again apparently due to race conditions caused by other tests. Explicitly setting the starting conditions for these tests seems to have solved the problem.

## Why
Test failures are bad. Also I want to remove jQuery from `static` and resolving them will make that easier.

## Visual changes
None.

### DIFF
--- a/spec/javascripts/surveys.spec.js
+++ b/spec/javascripts/surveys.spec.js
@@ -52,6 +52,7 @@ describe('Surveys', function () {
       GOVUK.cookie(surveys.surveySeenCookieName(survey), null)
     })
     $block.remove()
+    $('#global-bar').remove()
   })
 
   describe('init', function () {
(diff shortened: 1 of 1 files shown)


injms (approved):
Super - looks like some good detective work in figuring this out 🕵️

PR #2711, opened 2022-02-21
Conditionally add the GTM component

hannako:
We want to test out our configuration of GTM and implementation of Ga4 tracking. We are going to do this in integration only at this stage. The env vars required to render the GTM component have been made available to government frontend on integration only, see https://github.com/alphagov/govuk-puppet/pull/11550

Trello: https://trello.com/c/0yzp3M2T/785-ga4-conditionally-add-the-gtm-component-to-static-templates

### DIFF
--- a/app/views/root/_gem_base.html.erb
+++ b/app/views/root/_gem_base.html.erb
@@ -32,13 +32,15 @@
   global_bar << global_banner if global_banner
 %>
 
-<% if ENV["GOOGLE_TAG_MANAGER_ID"] %>
-   <%= render "govuk_publishing_components/components/google_tag_manager_script", {
-     gtm_id: ENV["GOOGLE_TAG_MANAGER_ID"],
-     gtm_auth: ENV["GOOGLE_TAG_MANAGER_AUTH"],
-     gtm_preview: ENV["GOOGLE_TAG_MANAGER_PREVIEW"]
-   } %>
- <% end %>
+<% content_for :head do %>
+  <% if ENV["GOOGLE_TAG_MANAGER_ID"] %>
+     <%= render "govuk_publishing_components/components/google_tag_manager_script", {
+       gtm_id: ENV["GOOGLE_TAG_MANAGER_ID"],
+       gtm_auth: ENV["GOOGLE_TAG_MANAGER_AUTH"],
+       gtm_preview: ENV["GOOGLE_TAG_MANAGER_PREVIEW"]
+     } %>
+  <% end %>
+<% end %>
 
 <%= render "govuk_publishing_components/components/layout_for_public", {
   account_nav_location: account_nav_location,
--- a/app/views/root/_gem_base.html.erb
+++ b/app/views/root/_gem_base.html.erb
@@ -32,6 +32,14 @@
   global_bar << global_banner if global_banner
 %>
 
+<% if ENV["GOOGLE_TAG_MANAGER_ID"] %>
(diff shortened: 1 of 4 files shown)


app/views/root/_gem_base.html.erb
```diff
@@ -32,6 +32,14 @@
   global_bar << global_banner if global_banner
 %>
 
+<% if ENV["GOOGLE_TAG_MANAGER_ID"] %>
+   <%= render "govuk_publishing_components/components/google_tag_manager_script", {
+     gtm_id: ENV["GOOGLE_TAG_MANAGER_ID"],
+     gtm_auth: ENV["GOOGLE_TAG_MANAGER_AUTH"],
+     gtm_preview: ENV["GOOGLE_TAG_MANAGER_PREVIEW"]
+   } %>
+ <% end %>
```
injms, line 41: Putting the component here means that it's placed before the opening `html` tag:


For context, we're aiming to reduce our reliance on `static` for templates - the plan is for the templates to be used directly by a frontend application rather than from `static`. 

With this in mind, I'd suggest that this should be added as an option for the public layout component in the components gem, and either the environment variables used in the component if they can be access there or passed into the component parameters. 

Happy to talk this through if there are other pressures that I'm not aware of.

injms, line 41: Putting the component here means that it's placed before the opening `html` tag:


For context, we're aiming to reduce our reliance on `static` for templates - the plan is for the templates to be used directly by a frontend application rather than from `static`. 

With this in mind, I'd suggest that this should be added as an option for the public layout component in the components gem, and either the environment variables used in the component if they can be access there or passed into the component parameters. 

Happy to talk this through if there are other pressures that I'm not aware of.

injms (dismissed):

app/views/root/_gem_base.html.erb
hannako: Thanks @injms for your review. Can i just confirm that you mean [here](https://github.com/alphagov/govuk_publishing_components/blob/master/app/views/govuk_publishing_components/components/_layout_for_public.html.erb)?

alex-ju:
```suggestion
<% content_for :head do %>
  <% if ENV["GOOGLE_TAG_MANAGER_ID"] %>
     <%= render "govuk_publishing_components/components/google_tag_manager_script", {
       gtm_id: ENV["GOOGLE_TAG_MANAGER_ID"],
       gtm_auth: ENV["GOOGLE_TAG_MANAGER_AUTH"],
       gtm_preview: ENV["GOOGLE_TAG_MANAGER_PREVIEW"]
     } %>
  <% end %>
<% end %>
```

I am suggesting we pass this script up to the layout component. I'd like to keep these details decoupled from the public layout and using env variables at the gem level doesn't sound like good practice. What do you think, @injms? Would you be ok with this approach?

alex-ju: This way we can carry the GTM experiment with minimal impact on the existing component (as yield in `<head>` is already available).

injms: Passing it into the `:head` makes sense for this experiment - happy with that as an approach for this, but if it becomes more permanent I think it should be looked at again.

hannako: OK great, thank you for @injms. I will dismiss your review where you requested the change, if you would mind approving please 😄

alex-ju (approved):

injms (approved):

PR #2719, opened 2022-03-03
Remove coronavirus banner

leenagupte:
Trello: https://trello.com/c/BpOxInF5

# What's changed and why?

Investigate how much work is required to remove the coronavirus banner.

# Findings

1. As there are no other banners, `show_global_bar` needs to be set to `false` otherwise a black bar is displayed after the GOV.UK header:

|show_global_bar = true|show_global_bar = false|
|:-----------------------|:-----------------------|
|global_banner_left_on|global_banner_turned_off|

2. After the global banner is turned off, the "Tell us what you think" survey sometimes makes a reappearance:

|Before|After|
|:------|:----|
|Screenshot 2022-03-03 at 14 00 16|Screenshot 2022-03-03 at 13 59 40global_banner_turned_off|

### DIFF
--- a/app/assets/stylesheets/components/_global-bar.scss
+++ b/app/assets/stylesheets/components/_global-bar.scss
@@ -50,10 +50,6 @@ $covid-grey: #262828;
 .global-bar-title,
 .global-bar-text {
   color: govuk-colour("white");
-
-  & + .global-bar-covid-wrapper {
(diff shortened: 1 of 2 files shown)


gclssvglx (approved):
Nice one @leenagupte - great stuff!

PR #2727, opened 2022-03-11
Bill's leaving today

sihugh:
We're waiving the usual 5 year rule because of Bill's exceptional impact here.


hannako (approved):
❤️ 😭

andysellick (approved):
My instinct was to reject this PR, I might be in denial about Bill leaving...

richardTowers (approved):
100%

We haven't been as consistent in doing these as we should have been, so apologies to any recent leavers we've missed. Definitely well deserved in this case!

PR #2758, opened 2022-04-25
Updating Dockerfile

nsabri1:
Update Static Dockerfile - to align with multi-build strategy.

Added new appUser UID 1001 and GID 1001

/app/public directory created and `chown` o `appuser` - this directory is used and created on runtime.
Prior to adding chown app would fail due to persimmon error as we are now using nonRoot containers.
`6"error":"Errno::EACCES\nPermission denied @ dir_s_mkdir - /app/public/templates\n/usr/local/lib/ruby/2.7.0/fileutils.rb:250:in `mkdir'\n/usr/local/lib/ruby/2.7.0/fileutils.rb:250:in `fu_mkdir'\n/usr/local/lib/ruby/2.7.0/fileutils.rb:228:in `block (2 levels) in mkdir_p'\n/usr/local/lib/ruby/2.7.0/`


Dockerfile
```diff
@@ -1,25 +1,45 @@
...
+RUN apt-get update -qy && \
+    apt-get upgrade -y && \
+    apt-get install -y nodejs
+ 
+RUN groupadd -g 1001 appuser && \
+    useradd appuser -u 1001 -g 1001 --home /app
+ 
+COPY --from=builder /usr/local/bundle/ /usr/local/bundle/
+COPY --from=builder /app /app
+RUN mkdir -p /app/public/templates && chown -R 1001:1001 /app/public
```
sengi, line 39: Does the app need to write to all of `/app/public` or just `/app/public/templates`?

nsabri1: Good shout - it does not - updated with git amend

sengi (approved):

PR #2759, opened 2022-04-26
Ruby 2 7 6

gclssvglx:
Update to Ruby 2.7.6

[Trello](https://trello.com/c/47FPs84V/390-ruby-276-update-static)


Gemfile
```diff
@@ -1,5 +1,7 @@
 source "https://rubygems.org"
 
+ruby File.read(".ruby-version").chomp
```
KludgeKML, line 3: I'm in two minds about this. On the one hand, it's DRYer, but on the other hand since we're using explicit versions for everything else in the gem file it feels like it breaks the norm.

(Not to say that because we don't do it elsewhere we shouldn't start here, just that explicit ruby version in the gemfile is the accepted idiom).

KludgeKML (approved):
LGTM

PR #2763, opened 2022-04-27
Migrate to use jasmine-browser-runner

chao-xian:
Migrating from jasmine gem which is deprecated and no supported in Rails 7.

This is a blocker to a Rails 7 upgrade

[Trello](https://trello.com/c/3J1dAhN8/973-migrate-static-to-use-jasmine-browse-runner-to-unblock-security-update)

PR #2771, opened 2022-05-04
Revert "Migrate to use jasmine-browser-runner"

chao-xian:
Reverts alphagov/static#2763

Main is broken for some odd reason with intermittent failures :(


edwardkerry (approved):
looks reverted to me

PR #2777, opened 2022-05-16
Change error page frontend

andysellick:
## What
Stop using the `title` component from `govuk_publishing_components` on the error pages. Because:

- error page had its own CSS and print CSS that was only needed for one component, title
- title can be achieved using govuk-frontend styles which are already available on this page
- without this we don't need these extra stylesheets, simpler

## Why
The inclusion of the title component in the error page styles, which aren't included elsewhere in `static` was also causing some chaos in some new features of the component auditing. This change makes both these things simpler.

## Visual changes
None (checked print appearance as well, no difference).


injms (approved):
Good find!

PR #2783, opened 2022-05-23
Re-apply use of jasmine-browser-runner (re-reverts #2771)

chao-xian:
Re-applies alphagov/static#2763 plus a fix to the specs that had a run order bug.

(Reverts alphagov/static#2771 which itself is a revert of alphagov/static#2763 )

Description from original PR:

Migrating from jasmine gem which is deprecated and no supported in Rails 7.

This is a blocker to a Rails 7 upgrade

[Trello](https://trello.com/c/3J1dAhN8/973-migrate-static-to-use-jasmine-browse-runner-to-unblock-security-update)


gclssvglx (approved):
Fantastic work! Making the cookie state consistent between tests definitely seems to have resolved the failing test issue. Nice one @chao-xian 👍

PR #2788, opened 2022-05-25
Use emergency banner component from govuk_publishing_components

patrickpatrickpatrick:
## What

Move the [`emergency_banner`](https://docs.publishing.service.gov.uk/repos/static/emergency-banner.html) component from [`static`](https://github.com/alphagov/static) to the [`govuk_publishing_components`](https://github.com/alphagov/govuk_publishing_components/) gem. 

### What Has Been Done

Delete the templates, documentation, stylesheets (these have been moved to the gem). Change calls to the emergency_banner partial to be calls to the emergency_banner component in  [`govuk_publishing_components`](https://github.com/alphagov/govuk_publishing_components/). Remove default link_text logic (if no link_text provided, set to 'More information') and associated tests as this is now handled by the component. As part of reimplementing emergency banner to be an isolated component, a new gem layout has been created with a homepage parameter. When set this will set the homepage parameter in emergency banner to be true and will apply the correct homepage styling to the emergency banner (if there is an emergency banner to be displayed).

### What Has Not Been Done

Rewrite `emergency_banner` to use `content_store` instead of Redis. There's a lot of work involved in this step which is not in the remit of merely moving this component over to `govuk_publishing_components`.

## Why

Part of an ongoing effort to refactor static and retire slimmer as outlined [here](https://trello.com/c/zLiC5VwG/7-objective-retire-static-and-slimmer).

## Testing On Integration

Used the pages linked to in the [developer documentation](https://docs.publishing.service.gov.uk/manual/emergency-publishing.html#some-example-pages-to-check).

### Notable Death

With description and link set but no link_text set.

Screenshot 2022-06-09 at 11 58 06

Screenshot 2022-06-09 at 11 58 54

Screenshot 2022-06-09 at 11 59 01

Screenshot 2022-06-09 at 11 59 16

### National Emergency

With description, link and link_text set.

Screenshot 2022-06-09 at 12 03 31

Screenshot 2022-06-09 at 12 03 38

Screenshot 2022-06-09 at 12 03 47

Screenshot 2022-06-09 at 12 03 53

### Local Emergency

No description or link set.

Screenshot 2022-06-09 at 12 09 06

Screenshot 2022-06-09 at 12 09 13
 
Screenshot 2022-06-09 at 12 09 19

Screenshot 2022-06-09 at 12 09 28


patrickpatrickpatrick:
https://github.com/alphagov/govuk_publishing_components/pull/2795 this will have to merged first for this to pass all tests and work

patrickpatrickpatrick:
have tested these changes in integration (with static app and frontend app and the branch of govuk_publishing_components) and they work as expected

chao-xian:
@injms this looks ok to me, how is this looking to you?

hannalaakso (approved):
We've now checked that the banner works in integration and staging so I think this is good to go 🙌 

I created a [tech debt ticket for checking the rake task for deploying the banner locally](https://trello.com/c/Pd1DRgUU/665-check-if-rake-task-in-static-to-deploy-emergency-banner-locally-works) (we don't think there should be a problem with it and it also isn't a high priority).

hannalaakso:
Just waiting on https://github.com/alphagov/frontend/pull/3260 now.

injms (dismissed):
Pausing this for the moment.

hannalaakso (approved):
I previously approved this so re-approving now. There's been a force push since but this was to accommodate the changes from https://github.com/alphagov/govuk_publishing_components/pull/2973. Now we're just waiting on that PR to be ready to be merged.

PR #2806, opened 2022-06-21
Update to the newest components gem

andysellick:
## What
Bump to the latest version of the components gem.

## Why
We want to test some of the code on integration and I can't wait for dependabot.

## Visual changes
None.


gclssvglx (approved):
Looks great - thanks @andysellick

PR #2808, opened 2022-06-24
Account templates/partials cleanup

danacotoran:
The account home has been removed, therefore the only page using an account template is now the **Manage your emails** page in `email-alert-frontend` https://github.com/alphagov/email-alert-frontend/blob/d76ca950331ff46173d1d429cc59243ed7d2d87a/app/controllers/subscriptions_management_controller.rb#L97 
There is no need to have any of the other templates any longer.

The templates left have identical content at the moment, and are `gem_layout_account_manager` and `gem_layout_account_manager_no_nav`. 

The `gem_layout_account_manager_no_nav` one can be removed once we've updated `subscriptions_management_controller.rb` in `email-alert-frontend` to reference `gem_layout_account_manager` instead (the `no_nav` part is no longer relevant as now there is no nav on the account by default).


alex9smith (approved):
LGTM. I've done a quick check through `email-alert-frontend` and `frontend` for any other uses of the `gem_layout_account_manager_*` templates and couldn't find any.

PR #2835, opened 2022-07-27
Test that gem_layout renders feedback component

robinjam:
[Trello card](https://trello.com/c/XDLC291t/2948-write-test-for-static-to-go-hand-in-hand-with-smokey-3)

This imports a test from Smokey[^1][^2] that confirms that the feedback component is present & functional.

The Smokey test does not currently meet the eligibility criteria for inclusion in the repository[^3], but by adding this test to Static we can meet the "Second critical functional check" criterion.

n.b. I'm new to GDS, so please do look over it critically and let me know if I've made any mistakes 🙏

[^1]: https://github.com/alphagov/smokey/blob/main/features/apps/static.feature
[^2]: https://github.com/alphagov/smokey/blob/main/features/step_definitions/feedback_steps.rb
[^3]: https://github.com/alphagov/smokey/blob/main/docs/writing-tests.md

### DIFF
--- a/test/integration/templates/gem_layout_test.rb
+++ b/test/integration/templates/gem_layout_test.rb
@@ -1,0 +1,34 @@
+require "integration_test_helper"
+
+class GemLayoutTest < ActionDispatch::IntegrationTest
+  setup do
+    Capybara.current_driver = Capybara.javascript_driver
+    visit "/templates/gem_layout.html.erb"
+  end
+
+  should "allow user to report a problem with the page" do
+    click_on "Report a problem with this page"
+    assert page.has_content?("Help us improve GOV.UK")
+    assert page.has_field?("What went wrong?")
+
+    # Regression test for scenario where wrong URL is set
+    url_input = page.find("form[action='/contact/govuk/problem_reports'] input[name=url]", visible: false)
+    assert_equal page.current_url, url_input.value
+  end
+
+  should "allow user to report that the page is not useful" do
+    click_on "No" # No, this page is not useful
+    assert page.has_content?("Help us improve GOV.UK")
+    assert page.has_field?("Email address")
+
+    # Regression test for scenario where wrong URL is set
+    url_input = page.find("form[action='/contact/govuk/email-survey-signup'] input[name='email_survey_signup[survey_source]']", visible: false)
+    full_path = URI(page.current_url).request_uri
+    assert_equal full_path, url_input.value
+  end
+
+  should "allow user to report that the page is useful" do
+    click_on "Yes" # Yes, this page is useful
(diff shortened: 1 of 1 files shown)


test/integration/templates/gem_layout_test.rb
```diff
@@ -0,0 +1,29 @@
...
+    Capybara.current_driver = Capybara.javascript_driver
+
+    visit "/templates/gem_layout.html.erb"
+
+    click_on "Report a problem with this page"
+    assert page.has_content?("Help us improve GOV.UK")
+    assert page.has_field?("What went wrong?")
+    # Regression test for scenario where wrong URL is set
+    url_input = page.find("form[action='/contact/govuk/problem_reports'] input[name=url]", visible: false)
+    assert_equal page.current_url, url_input.value
```
ChrisBAshton, line 14: This could be in its own `should` block, e.g. `should "allow user to report a problem with the page"`

test/integration/templates/gem_layout_test.rb
```diff
@@ -0,0 +1,29 @@
...
+    click_on "Close"
+
+    click_on "No" # No, this page is not useful
+    assert page.has_content?("Help us improve GOV.UK")
+    assert page.has_field?("Email address")
+    # Regression test for scenario where wrong URL is set
+    url_input = page.find("form[action='/contact/govuk/email-survey-signup'] input[name='email_survey_signup[survey_source]']", visible: false)
+    full_path = URI(page.current_url).request_uri
+    assert_equal full_path, url_input.value
```
ChrisBAshton, line 23: This could be in its own `should` block, e.g. `should "allow user to report the page is not useful"`

test/integration/templates/gem_layout_test.rb
```diff
@@ -0,0 +1,29 @@
...
+    click_on "Close"
+
+    click_on "Yes" # Yes, this page is useful
```
ChrisBAshton, line 26: This could be part of the first block, which you might rename to `should "allow user to report the page is useful`.

ChrisBAshton (commented):
Looks good! ⭐  A couple of minor comments below.

I wasn't expecting an interactive test - checking that a particular thing is rendered would probably be sufficient - but I see no harm in putting the more detailed test in here (perhaps even gives us an opportunity to simplify the corresponding Smokey later on!).

I'll give @kevindew a chance to review too, as he has the most recent context on these tests.

robinjam:
Thanks for the review & the great suggestions @ChrisBAshton 🙏

ChrisBAshton (approved):
LGTM 👍 nice one!

PR #2839, opened 2022-08-03
Explicitly allow redirects to other hosts within gov.uk domain

AgaDufrat:
In rails 7 redirects to other hosts are no longer permitted by default
and ActionController::Redirecting::UnsafeRedirectError was raised.

However in all those cases we are redirect to other GOV.UK domains such as
assets.gov.uk

Tested in staging


leenagupte (approved):
👍

PR #2843, opened 2022-08-12
Heed the `GA_*_ID` env vars at runtime.

sengi:
The Google Analytics IDs were being burned into the `analytics.js` Rails asset at build time (i.e. `rails assets:precompile`), which was causing all environments to use the same property IDs when deployed as containers. In the containerised world we want to compile the assets once, at build time, and upload them to a central service for serving static files (Amazon S3).

While the account number is the same (UA-26179049), each environment has a distinct property number (integration=22, staging=20, prod=1).

Avoid substituting environment variables into JavaScript files via ERB at build time. Instead, move these up into HTML templates so that changes to `GA_*_ID` take effect at container startup rather than requiring a rebuild.

#### Testing

Deployed to the integration EKS cluster and observed GA collect requests are now being sent with the expected web property ID (`UA-26179049-22`) for integration, instead of `UA-26179049-1` as before.

`POST | https://www.google-analytics.com/j/collect?v=1&_v=j96&aip=1&a=1642051629&t=pageview&_s=1&dl=https://www.eks.integration.govuk.digital/&dp=/&ul=en-gb&de=UTF-8&dt=Welcome to GOV.UK&sd=30-bit&sr=1920x1200&vp=1145x767&je=0&_u=SCEACAABBAAAAC~&cid=1100976322.1652793698&tid=UA-26179049-22&_gid=1606836569.1660302003&_slc=1&cd15=200&cd16=unknown&cd95=1100976322.1652793698&cd11=1&cd2=homepage&cd3=other&cd4=f3bbdec2-0e62-4520-a7fd-6ffd5d36e03a&cd9=<OT1056>&cd12=not withdrawn&cd17=homepage&cd20=frontend&cd30=none&cd31=none&cd32=none&cd56=other&cd57=other&cd58=other&cd59=other&cd39=false&cd89=frontend&cd26=0&cd27=0&cd23=unknown&cd38=false&cd100=true&z=739931572`

Didn't see any new errors in the JS console. Only tested on latest Firefox, Safari and Edge (all macOS). Can't easily inspect the GA collection requests on Chrome because it does some strange internal redirect, but there were no errors there either.

### DIFF
--- a/app/assets/javascripts/analytics.js.erb
+++ b/app/assets/javascripts/analytics.js.erb
@@ -136,9 +136,7 @@ var linkedDomains = [
 ]
 
 window.GOVUK.analyticsVars = window.GOVUK.analyticsVars || {}
-window.GOVUK.analyticsVars.gaProperty = '<%= Rails.application.config.ga_universal_id %>'
 window.GOVUK.analyticsVars.primaryLinkedDomains = ['account.gov.uk']
-window.GOVUK.analyticsVars.gaPropertyCrossDomain = '<%= Rails.application.config.ga_secondary_id %>'
 window.GOVUK.analyticsVars.linkedDomains = linkedDomains
 
 if (typeof window.GOVUK.analyticsInit !== 'undefined') {
--- a/app/views/root/_gem_base.html.erb
+++ b/app/views/root/_gem_base.html.erb
@@ -31,6 +31,12 @@
 %>
 
 <% content_for :head do %>
+  <script>
+    window.GOVUK = window.GOVUK || {};
+    window.GOVUK.analyticsVars = window.GOVUK.analyticsVars || {};
+    window.GOVUK.analyticsVars.gaProperty = '<%= Rails.application.config.ga_universal_id %>';
+    window.GOVUK.analyticsVars.gaPropertyCrossDomain = '<%= Rails.application.config.ga_secondary_id %>';
+  </script>
   <% if ENV["GOOGLE_TAG_MANAGER_ID"] %>
      <%= render "govuk_publishing_components/components/google_tag_manager_script", {
        gtm_id: ENV["GOOGLE_TAG_MANAGER_ID"],


kevindew:
To add some additional context to this - the use of erb in assets is basically incompatible with containerising applications. As we need the application to have one static set of assets that do not change based on environment variables. (erb based assets are essentially rather dangerous, it'd be great if they could be forbidden)

This app is one of the very few (and, hopefully, sole) case of an erb asset - in the anlaytics file. And to make this compatible with replatforming we need to make it a plain JS file.

tagging @andysellick and @gclssvglx.

app/assets/javascripts/analytics.js.erb
```diff
@@ -136,9 +136,7 @@ var linkedDomains = [
 ]
```
kevindew, line 136: We've also a conditional block at the top of this file: 
https://github.com/alphagov/static/blob/dea9d162882005b2205575a368d1b771744effcb/app/assets/javascripts/analytics.js.erb#L3-L5

that we'll need to remove otherwise this conditional won't work in replatforming and this will leave a lingering .erb file to trip someone up later. 

I wonder if the ga4-analytics can be a separate file where the ENV var conditional is in the HTML? @andysellick what do you think?

app/assets/javascripts/analytics.js.erb
```diff
@@ -136,9 +136,7 @@ var linkedDomains = [
...
 
 window.GOVUK.analyticsVars = window.GOVUK.analyticsVars || {}
-window.GOVUK.analyticsVars.gaProperty = '<%= Rails.application.config.ga_universal_id %>'
 window.GOVUK.analyticsVars.primaryLinkedDomains = ['account.gov.uk']
-window.GOVUK.analyticsVars.gaPropertyCrossDomain = '<%= Rails.application.config.ga_secondary_id %>'
 window.GOVUK.analyticsVars.linkedDomains = linkedDomains
```
kevindew, line 140: I wonder if we risk any problems if these variables aren't set for whatever reason and need to check that gaProperty is set or not before initialising analytics. If it fails gracefully all the better.

app/views/root/_gem_base.html.erb
```diff
@@ -31,6 +31,12 @@
 %>
 
 <% content_for :head do %>
+  <script>
+    window.GOVUK = window.GOVUK || {};
+    window.GOVUK.analyticsVars = window.GOVUK.analyticsVars || {};
+    window.GOVUK.analyticsVars.gaProperty = '<%= Rails.application.config.ga_universal_id %>';
+    window.GOVUK.analyticsVars.gaPropertyCrossDomain = '<%= Rails.application.config.ga_secondary_id %>';
+  </script>
```
kevindew, line 39: It's probably worth seeking some other opinions on preferences of how these properties are set since it'll be on every page of GOV.UK

I somewhat wondered if we might want them to be meta elements `e.g <meta name="govuk:ga_universal_id" content="<%= Rails.application.config.ga_universal_id %>">` is perhaps a little cleaner and then get that data by JS, rather than mix things up. This would also mean we don't need to fix this later as an example of `unsafe-inline` JS, if we ever manage to get a content security policy implemented.

I feel like @andysellick is probably closer to this code than me though so be good to get his thoughts.

app/assets/javascripts/analytics.js.erb
sengi: Oh good question. I think we're good though:

https://sourcegraph.com/github.com/alphagov/govuk_publishing_components/-/blob/app/assets/javascripts/govuk_publishing_components/analytics/init.js?L8:16

app/views/root/_gem_base.html.erb
injms: Similar to Kevin, I'd prefer this to not be set as an inline script - a meta tag or a data attribute on the script itself would be more future proof.

sengi: Great idea, SGTM! (you can probably tell I haven't done any frontend stuff professionally for over a decade 😅)

Stand by for updated implementation :)

app/assets/javascripts/analytics.js.erb
andysellick: We're currently working to remove this logic entirely, hopefully within the next few weeks. We were using this approach to get the new analytics up and running quickly on integration but we're now looking at doing it 'properly' i.e. integrating it with the cookie consent mechanism and doing it entirely in JS (no .erb required).

app/views/root/_gem_base.html.erb
andysellick: I agree with the suggestion although I think it'll be slightly more complicated to implement and we're looking to retire this analytics code in a year's time so if it turns out to be problematic then maybe don't worry about it too much?

andysellick (commented):
Happy with this change as it is - as I said, an inline script isn't ideal but we're looking to retire this code in a year's time, so I'm not worried about that.

What I think would be helpful is if all of this could be documented somewhere (if it isn't already). Specifically I'd like to be able to understand how these GA property IDs are retrieved and used in different environments, and where they come from (looks like environment variables set somewhere? If so why does `production.rb` fall back to a hard coded value that matches production, when it presumably should get that correct value from the environment variable? and similar questions)

gclssvglx (commented):
Happy with this change - the approach makes total sense.

app/views/root/_gem_base.html.erb
sengi: If you're ok with the inline script approach then happy to go with that 👍 (I haven't had time to redo it with `meta`, though I completely agree that would be the way to go if we weren't planning on deleting this soon-ish.)

sengi:
> Happy with this change as it is - as I said, an inline script isn't ideal but we're looking to retire this code in a year's time, so I'm not worried about that.
> 
> What I think would be helpful is if all of this could be documented somewhere (if it isn't already). Specifically I'd like to be able to understand how these GA property IDs are retrieved and used in different environments, and where they come from (looks like environment variables set somewhere? If so why does `production.rb` fall back to a hard coded value that matches production, when it presumably should get that correct value from the environment variable? and similar questions)

The reason for the surprising behaviour (which this PR fixes) is that the erb templates are evaluated **at build time** (i.e. `rails assets:precompile` rather than at runtime. Hence (as Kevin pointed out) we should avoid using templates to substitute (or "interpolate", in Ruby parlance) any configuration values into Rails assets. If we stick to this principle then we'll avoid this type of problem.

The reason this wasn't already broken (i.e. why we were getting away with it, or in other words the thing that was hiding the bug) was that in the old EC2/Puppet system we actually build the application on every machine when we roll out a release (which isn't best practice, but it's how Rails apps typically used to be deployed back when GOV.UK was new). We can't get away with doing that any more because container-based hosting requires (by design) that we promote the same build product through the environments to production rather than rebuilding each time.

So once this is merged, there should be nothing to document — these analytics IDs will work the same way as any other configuration in our Rails apps (environment variable FOO -> Rails config/production.rb -> Rails.application.config.foo, so the standard 12-factor stuff).

sengi:
Looks like this PR is obsoleted by the GA4 launch 😀🎉 Thanks, everyone!

PR #2854, opened 2022-08-30
Prevent GOVUK_ASSET_HOST from setting the asset host

theseanything:
This environment env var isn't being used in our infrastructure to set asset host as ASSET_HOST is always set and takes precedence. Neither is it used in our Heroku preview apps nor for local development (govuk-docker and ./startup.sh). This cleans up the codebase and removes confusion around which environment variables should be set.


kevindew (approved):
Nice one. Great to get rid of the extra complexity.

PR #2868, opened 2022-09-08
Use GA4 init code

andysellick:
## What

- runs the new GA4 init code, which starts the analytics code only after cookie consent
- note that this will not run anywhere but integration because the code it refers to is only included on integration

This will depend on a new version of the components gem that includes https://github.com/alphagov/govuk_publishing_components/pull/2915, until then this will silently do nothing. See that PR for details of how this works.

## Why

Part of the process to move away from using the GTM component and integrating all of the new analytics code properly into the cookie consent mechanism.

## Visual changes
None.

Trello card: https://trello.com/c/vrMj18et/353-integrate-gtm-with-cookie-consent-mechanism


JamesCGDS (dismissed):
All good 👍

AshGDS (approved):
All good :+1:

PR #2869, opened 2022-09-08
Use a wider div to wrap banner text

1pretz1:
Currently, `govuk-grid-column-two-thirds` knocks the 'II' onto a second line for the Queens death emergency banner. Using three thirds keeps all the text on a single line for larger screens.

## Before 

Screenshot 2022-09-08 at 19 41 05

## After

Screenshot 2022-09-08 at 19 40 43
Screenshot 2022-09-08 at 19 40 52

### DIFF
--- a/app/views/components/_emergency_banner.html.erb
+++ b/app/views/components/_emergency_banner.html.erb
@@ -1,7 +1,7 @@
 <div class="emergency-banner emergency-banner--<%= campaign_class %> dont-print" role="banner" data-nosnippet>
   <div class="govuk-width-container">
     <div class="govuk-grid-row">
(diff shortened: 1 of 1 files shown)


dgm34 (approved):

sihugh (approved):

myakura:
Even if this widens the space that the grid spans, there will be chances where "II" pushes to next line. What about inserting `&nbsp;` between the name and "II" ?

1pretz1:
Thanks for the suggestion @myakura, that is very true. `&nbsp;` would make sense but can't currently implement that with the way the banner text is populated.

PR #2874, opened 2022-09-09
Avoid 2 banner roles on the same page

maxf:
The top banner also has role=banner, so we're
marking this one with an aria-labelled-by attribute instead
and using `section`, which has the same semantics as `<div role="region">`


patrickpatrickpatrick (approved):

maxf:
Rendering differences:

Screenshot 2022-09-09 at 16 05 59

New markup in browser

Screenshot 2022-09-09 at 16 04 54

PR #2876, opened 2022-09-12
New Template with No Emergency Banner

patrickpatrickpatrick:
## What

A new template which does not display the emergency banner, even when it is active.

## Why

For the Queen topical page, it was requested that the emergency be removed as it creates a circular journey and pushes content down. This is because the emergency banner has a link to the page that it appears on. The addition of this template will prevent this happening in the future.

### DIFF
--- a/app/controllers/root_controller.rb
+++ b/app/controllers/root_controller.rb
@@ -29,6 +29,7 @@ class RootController < ApplicationController
     gem_layout_full_width
     gem_layout_full_width_explore_header
     gem_layout_full_width_no_footer_navigation
+    gem_layout_no_emergency_banner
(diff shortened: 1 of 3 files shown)


chao-xian:
Further context is that when we have the Emergency Banner up that points to a page within GOV.UK we'd likely not want the banner on it there as well. This allows us to do that in other places later on too.

patrickpatrickpatrick:
> Further context is that when we have the Emergency Banner up that points to a page within GOV.UK we'd likely not want the banner on it there as well. This allows us to do that in other places later on too.

rewrote the commit message to make it clearer of all use cases of this template

chao-xian (approved):

PR #2886, opened 2022-09-22
Look for gtag env var

andysellick:
## What
Allow the use of Google Tag in place of Google Tag Manager.

- modify the GA4 code to check for the existence of a gtag environment variable alongside the existing code to check for a GTM env var
- if the gtag env var is found, this will be used in preference to GTM by the code called in the components gem from window.GOVUK.analyticsGA4.init()
- otherwise will default back to GTM for all environments, or simply do nothing if no GTM or gtag env vars are found

## Why
We want to test Gtag on integration and this allows us to do it without breaking things on other environments. Related to: https://github.com/alphagov/govuk_publishing_components/pull/2979

## Visual changes
None.

Trello card: https://trello.com/c/Oo0T8ZCa/178-spike-into-back-out-plan-from-gtm-to-gtag

### DIFF
--- a/app/assets/javascripts/analytics.js.erb
+++ b/app/assets/javascripts/analytics.js.erb
@@ -145,14 +145,22 @@ if (typeof window.GOVUK.analyticsInit !== 'undefined') {
   window.GOVUK.analyticsInit()
 }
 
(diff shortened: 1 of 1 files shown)


AshGDS (approved):
All good :+1:

PR #2891, opened 2022-09-26
Get components gem version

andysellick:
## What
Get the version of the components gem and store it in the `window.GOVUK.analyticsGA4.vars` object, for later reference in the analytics code in the components gem.

- needed for the GA4 analytics code
- this was being created in the components gem, but that meant including a .js.erb file there, which we're trying to move away from
- instead, include it here, so we reduce the number of .js.erb files and deal with this later

Related change: https://github.com/alphagov/govuk_publishing_components/pull/2984

## Why
Using `.js.erb` files is apparently deprecated, so we're trying to avoid using them. This removes the `.js.erb` file in the gem (well, renames it back to `.js`) and moves the code here, so at least it's all in one place and we can come up with a single solution to this later.

## Visual changes
None.

Trello card: https://trello.com/c/Yu6Fw4ob/220-tech-debt-to-be-addressed


AshGDS (approved):
All good :+1:

JamesCGDS (approved):
All good 👍  I'll update [PR #2892](https://github.com/alphagov/static/pull/2892) accordingly

PR #2892, opened 2022-09-26
Update analyticsGA4 naming convention

JamesCGDS:
## What
This PR updates the GA4 variable names to coincide with the naming convention updates in the `govuk_publishing_components` gem (PR #2987).

## Why
The `govuk_publishing_components` gem has updated its GA4 naming conventions and this needs to be reflected in this application so that non-existent variables aren't being targeted.

## Visual Changes
N/A


andysellick (approved):
Looks good 👍

PR #2904, opened 2022-10-14
Add apprenticeships.gov.uk to linkedDomains

andyhd:
Added apprenticeships.gov.uk and find-employer-schemes.education.gov.uk to linkedDomains to enable cross-domain Google Analytics tracking.


maxf (commented):
How do new domains get added? Who decides?

andyhd:
@maxf that would be a question for @zilnhoj

andysellick (approved):

PR #2916, opened 2022-10-26
Remove unused gem

KludgeKML:
Doesn't look as though this gem is currently used, so very minor change!

https://trello.com/c/b9PRFqgU/231-applications-use-govuk-content-schema-test-helpers-which-is-an-archived-gem

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -32,7 +32,6 @@ end
 
 group :test do
   gem "climate_control"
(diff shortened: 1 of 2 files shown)


hannako (approved):
Yep https://github.com/alphagov/static/search?q=GovukContentSchemaTestHelpers gives nada

✂️

PR #2931, opened 2022-11-15
Add option to use a temporary directory for page cache

samsimpson1:
This is required as we're using read-only filesystems in replatforming and by default rails attempts to write cached pages to a read-only part of the filesystem


sengi (commented):
LGTM. I've requested a courtesy review from @AgaDufrat on #navigation-and-presentation-govuk (Aga, feel free to reassign if you want, of course)

AgaDufrat (approved):
LGTM

PR #2939, opened 2022-11-28
Remove image_optim gem

chao-xian:
This was added way back in 2015 for developers to compress images locally. There are other ways of doing that locally and this adds an unnecessary dependency to the app.


kevindew:
Do you want to remove https://github.com/alphagov/static/blob/main/docs/image-optimisation.md too?

PR #2959, opened 2022-12-14
Remove dependency on .js.erb and puppet for loading our analytics

AshGDS:
Hi @andysellick / @JamesCGDS - would you be able to approve this if all is OK? Thanks :+1:

This removes all the code which initialised our analytics from static. This is no longer used as we can't use puppet with our new infrastructure. Therefore our analytics is now initialised using https://github.com/alphagov/govuk_publishing_components/pull/3145


app/assets/javascripts/application.js
```diff
@@ -3,7 +3,7 @@
 //= require govuk_publishing_components/lib/trigger-event
 //= require govuk_publishing_components/lib/cookie-functions
 
-//= require analytics
+//= require load-analytics
```
andysellick, line 6: This looks well thought out but I've only given it a quick glance. My initial question is did you consider moving some or all of this into `govuk_publishing_components`? We're aiming to retire `static` eventually and putting this together with the rest of the analytics code seems like a good idea. Fine if there's a good reason why not, I'll find some time to do a proper review soon...

AshGDS: @andysellick Nope I didn't consider `govuk_publishing_components`... but yeah I guess the only reason it was in `static` was because that was where the environment variables lived which we don't need anymore. 

Should we pause reviewing this for now and i'll move it into `govuk_publishing_components`?

andysellick: I think it would be worth investigating, yes. As I said, there might be a reason to keep it in `static` (or at least part of it) but worth checking.

andysellick: Thanks

andysellick: I've had another thought overnight - it might also be worth splitting out this code to separate the UA stuff and the GA4 stuff - maybe have one init script in the existing `analytics/` dir and one in the `analytics-ga4/` dir. Again, I've not looked into the code to check the feasibility of this, so it's not essential if it proves tricky.

AshGDS:
I'll see if it's feasible in `govuk_publishing_components` so you don't need to worry about reviewing this for now @andysellick @JamesCGDS - thanks

AshGDS:
As it's feasible to achieve this in `govuk_publishing_components`, this branch is just removing the code we no longer need. https://github.com/alphagov/govuk_publishing_components/pull/3145

AshGDS:
@andysellick I've pushed the fix for failing the Jasmine tests in `static`. It's failing on GitHub as we haven't published the `load-analytics.js` change in the components gem yet so it can't see it. 

To test it you'll need to have your static pointed to your local publishing components, with the `load-analytics` branch checked out in your local publishing components repo.

Thanks :+1:

andysellick:
@AshGDS latest changes look good, well done figuring this out. I guess in order to get this PR to be mergable we'll need to get your gem changes into a release, get it in static and then rebase this PR? I think you said it was safe to do that, as the code can still be initialised by static, it just ignores the new stuff you've written?

AshGDS:
Thanks @andysellick - the branch has the latest publishing components now :+1:

PR #2988, opened 2023-01-17
Remove workaround for bad `mail` gem release.

sengi:
Remove the version constraint that we were using to avoid the bad release of the `mail` gem, now that https://www.github.com/mikel/mail/issues/1489 is fixed.

Update to `2.8.0.1`, which fixes the permissions issue.

Generated with `gsed -i '/gem "mail"/d' && bundle update mail`.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,7 +10,6 @@ gem "gds-api-adapters"
 gem "govuk_app_config"
 gem "govuk_personalisation"
 gem "govuk_publishing_components"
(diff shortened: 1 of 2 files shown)


chao-xian:
FYI https://github.com/alphagov/static/pull/2981/files

chao-xian:
Surprised this isn't a conflict. Closing

sengi:
Thanks! Yeah very strange that this showed up as mergeable at all.

PR #3039, opened 2023-04-14
Deploy Global Banner for Emergency Alerts Test

patrickpatrickpatrick:
## What

Show the global bar with a message about upcoming Emergency Alerts test. Change the Emergency Bar background colour to light blue (so as not to confuse it with the black Emergency Banner). Inline the message text so the bar does not take two lines and push content to far down the page.

## Why

In advance of the Emergency Alerts test, we want to raise awareness of them happening. This will be removed after the Emergency Alerts test takes place.

Screenshot 2023-04-14 at 13 23 25

[Relevant Trello Card](https://trello.com/c/WtRR3aGm/1775-urgent-emergency-alerts-site-wide-banner-m)

### DIFF
--- a/app/assets/stylesheets/components/_global-bar.scss
+++ b/app/assets/stylesheets/components/_global-bar.scss
@@ -8,7 +8,7 @@ $covid-grey: #262828;
 
 .global-bar {
   @include govuk-font(19);
-  background-color: govuk-colour("black");
+  background-color: #d9e7f2;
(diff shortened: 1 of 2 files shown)


chao-xian (approved):
LGTM but not sure if the span affects anything else. Might want @matthillco to take a look too, but I don't want that to be a blocker - changes here look pretty minor.

PR #3053, opened 2023-05-05
Add gem_layout_full_width_browse_header template

patrickpatrickpatrick:
## What

Add a new template for Browse page headers on the collection application. Update the gem_base with new attributes that have been added to the public_layout.

## Why

There are instances of pages on collections that have a coloured background in the heading section. This meant that the the blue bar (which is rendered in the public_layout) had to be rendered by collections otherwise it would result in a gap between the blue bar and coloured section. This solution seemed fine until we had a recent deployment of the global bar. As the global bar includes a blue bar, it meant that the Browse pages had two blue bars. They therefore needed to be removed while the global bar was present and then added back again when the global bar was removed. 

To avoid this needing to happen the next time the bar is deployed, this new template has been implemented. It means that if a page has a coloured background heading it can be specified in the template and then it will be applied to the blue bar. This means the blue bar can be rendered in public_layout and so applications don't need to render the blue bar if they have a heading section with a background colour.

[Relevant Trello Card](https://trello.com/c/pT7Sh8Yn/1793-update-layoutforpublic-template-so-that-global-bar-is-independent-of-requiring-changes-in-the-browse-template-l)

### DIFF
--- a/app/controllers/root_controller.rb
+++ b/app/controllers/root_controller.rb
@@ -27,6 +27,7 @@ class RootController < ApplicationController
     gem_layout_account_manager
     gem_layout_explore_header
     gem_layout_full_width
+    gem_layout_full_width_browse_header
     gem_layout_full_width_explore_header
     gem_layout_full_width_no_footer_navigation
     gem_layout_homepage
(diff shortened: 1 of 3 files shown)


patrickpatrickpatrick:
https://github.com/alphagov/govuk_publishing_components/pull/3380 depends on this being merged in first

jon-kirwan (approved):
All looks good to me 👍

PR #3072, opened 2023-06-02
Update playbook for global banner.

sengi:
- We no longer run Varnish.
- Update the troubleshooting steps now that Jenkins is no longer involved and Static's template cache no longer persists across rollouts.


chao-xian:
@sengi instead of links to search results can't we link to those URLs you've listed anyway?

chao-xian:
Ohhh it wasn't clear that these pages are showing up in the Dev Docs. So we'd have a cyclical situation if dead pages pointed to the two pages you listed in the description of this PR. OK this PR makes sense as it is then.

chao-xian (approved):

PR #3116, opened 2023-07-28
Make `emergency_banner` rake task more robust

ChrisBAshton:
## What

The rake task will now fail fast if:

- Too many arguments are provided
- The `campaign_class` argument has an unrecognised value
- The `link` argument is not a valid URI

## Why

Previously, inputting a typo would make Origin return 500 errors! By running the rake task with an unescaped comma in the input, as below:

```
kubectl -n apps exec deploy/static -- rake "emergency_banner:deploy['notable-death','Henry Fielding dies','English novelist and dramatist known for his earthy humour and satire dies, age 47','https://en.wikipedia.org/wiki/Henry_Fielding','More information']"
```

The bad use of comma would give the following incorrect entry in redis:

```
irb(main):001:0> Redis.new.hgetall("emergency_banner")
=>
{"short_description"=>
  "'English novelist and dramatist known for his earthy humour and satire dies",
 "heading"=>"'Henry Fielding dies'",
 "campaign_class"=>"'notable-death'",
 "link"=>"age 47'",
```

This resulted in an "access denied" error for https://assets.staging.publishing.service.gov.uk/templates/gem_layout_homepage.html.erb, and subsequently a 500 error for all pages on Origin (e.g. https://www-origin.staging.publishing.service.gov.uk/). The immediate user impact would be minimal, as our CDN automatically falls back to the mirrors, but even so - that's not a good thing!

Therefore we decided to add some safeguards to make it harder for a developer to accidentally break the site.

Trello: https://trello.com/c/ocrS67MT/3222-make-emergency-banner-rake-task-more-robust

### DIFF
--- a/lib/emergency_banner/deploy.rb
+++ b/lib/emergency_banner/deploy.rb
@@ -1,6 +1,13 @@
+require "uri"
+
 module EmergencyBanner
   class Deploy
     def run(campaign_class, heading, short_description = "", link = "", link_text = "")
+      if link != ""
+        uri = URI.parse(link)
+        raise ArgumentError, "Invalid URL provided: #{link}" unless uri.is_a?(URI::HTTP) && !uri.host.nil?
+      end
+
       redis = Redis.new
       redis.hmset(
         :emergency_banner,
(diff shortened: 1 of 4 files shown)


hannako (approved):
Yikes! Thank you!

MuriloDalRi (approved):

PR #3134, opened 2023-08-25
Add GA4 tracking to survey banner

AshGDS:
## What

- Adds GA4 tracking to the survey banner. More specifically, it adds a `data-ga4-auto` to send the presence of the banner, `data-ga4-event` for the 'Close' button, and `data-ga4-link` for the survey link.
- As the survey is inserted into the DOM after the modules.js code runs, we can't simply add data-module= to the elements. Instead we manually initialise the GA4 tracking on each element. As JavaScript inserts the banner HTML, we only need to add the tracking attributes here and not in any HTML templates.
- The survey banner is inserted after the GA4 pageview runs, which is why we are using `data-ga4-auto` instead.

## Why

https://trello.com/c/udF07C9O/662-banners-survey-banner-element-visibility-navigation-close-user-interaction-and-form-events

## How to test

- Run a local version of static with this branch checked out
- In `surveys.js` go to the function `isSurveyToBeDisplayed` and set it to just return true, e.g.:
```JavaScript
 isSurveyToBeDisplayed: function (survey) { return true }
```
- In `gem_base.html.erb`, add the HTML element somewhere: 
```HTML
<div id="user-satisfaction-survey-container"></div>
```

### DIFF
--- a/app/assets/javascripts/surveys.js
+++ b/app/assets/javascripts/surveys.js
@@ -15,10 +15,22 @@
   var templateBase = function (children) {
     return (
       '<section id="user-satisfaction-survey" class="visible" aria-hidden="false">' +
-      '  <div class="survey-wrapper govuk-width-container">' +
-      '    <a class="govuk-link survey-close-button" href="#user-survey-cancel" aria-labelledby="survey-title user-survey-cancel" id="user-survey-cancel" role="button">Close</a>' +
+      '  <div class="survey-wrapper govuk-width-container" data-module="ga4-auto-tracker" ' +
+              'data-ga4-auto=\'' + JSON.stringify({ event_data: { event_name: 'element_visible', type: 'survey banner' } }) +
+              '\'>' +
+      '    <a class="govuk-link survey-close-button" ' +
+              'href="#user-survey-cancel" ' +
+              'aria-labelledby="survey-title user-survey-cancel" ' +
+              'id="user-survey-cancel" ' +
+              'role="button" ' +
+              'data-module="ga4-event-tracker" ' +
+              'data-ga4-event=\'' + JSON.stringify({ event_name: 'select_content', type: 'survey banner', action: 'closed' }) +
+            '\'>Close</a>' +
       '    <h2 class="survey-title" id="survey-title">{{title}}</h2>' +
-           children +
+          '<div data-module="ga4-link-tracker" data-ga4-track-links-only ' +
+                'data-ga4-link=\'' + JSON.stringify({ event_name: 'navigation', type: 'survey banner', index: 1, index_total: 1 }) + '\'>' +
+            children +
+          '</div>' +
       '  </div>' +
       '</section>'
     )
@@ -197,12 +209,14 @@
     displayURLSurvey: function (survey, surveyContainer) {
       var urlSurveyTemplate = userSurveys.getUrlSurveyTemplate()
       surveyContainer.innerHTML = urlSurveyTemplate.render(survey)
+      window.GOVUK.modules.start(surveyContainer) // Initialises our GA4 modules
       userSurveys.setURLSurveyEventHandlers(survey)
     },
 
     displayEmailSurvey: function (survey, surveyContainer) {
       var emailSurveyTemplate = userSurveys.getEmailSurveyTemplate()
       surveyContainer.innerHTML = emailSurveyTemplate.render(survey)
+      window.GOVUK.modules.start(surveyContainer) // Initialises our GA4 modules
       userSurveys.setEmailSurveyEventHandlers(survey)
     },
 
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/surveys.js
```diff
@@ -194,15 +202,29 @@
       userSurveys.trackEvent(survey.identifier, 'banner_shown', 'Banner has been shown')
     },
 
+    initialiseModuleOnElement: function (element, Module) {
+      if (Module) {
+        var module = new Module(element)
+        module.init()
+        return module
```
andysellick, line 212: Why is `module` returned here? I can't see it being used anywhere.

app/assets/javascripts/surveys.js
```diff
@@ -194,15 +202,29 @@
...
+      }
+    },
+
     displayURLSurvey: function (survey, surveyContainer) {
       var urlSurveyTemplate = userSurveys.getUrlSurveyTemplate()
       surveyContainer.innerHTML = urlSurveyTemplate.render(survey)
+      this.initialiseModuleOnElement(surveyContainer, window.GOVUK.Modules.Ga4LinkTracker)
+      this.initialiseModuleOnElement(surveyContainer, window.GOVUK.Modules.Ga4EventTracker)
```
andysellick, line 220: What's the event tracker used for? I'm guessing something to do with the form, but I don't see a `data-ga4-event` anywhere.

app/assets/javascripts/surveys.js
```diff
@@ -194,15 +202,29 @@
...
+      this.initialiseModuleOnElement(surveyContainer.querySelector('[data-ga4-auto]'), window.GOVUK.Modules.Ga4AutoTracker)
       userSurveys.setURLSurveyEventHandlers(survey)
     },
 
     displayEmailSurvey: function (survey, surveyContainer) {
       var emailSurveyTemplate = userSurveys.getEmailSurveyTemplate()
       surveyContainer.innerHTML = emailSurveyTemplate.render(survey)
+      this.initialiseModuleOnElement(surveyContainer, window.GOVUK.Modules.Ga4LinkTracker)
+      this.initialiseModuleOnElement(surveyContainer, window.GOVUK.Modules.Ga4EventTracker)
+      this.initialiseModuleOnElement(surveyContainer.querySelector('[data-ga4-auto]'), window.GOVUK.Modules.Ga4AutoTracker)
```
andysellick, line 230: Looks like these three lines are repeated in both this function and `displayURLSurvey` - would it be sensible to move them into a separate function and call that?

app/assets/javascripts/surveys.js
AshGDS: It's on line 26, for the `Close` button on the banner

app/assets/javascripts/surveys.js
AshGDS: Previously I thought I needed it as I thought I needed to call a function on one of the modules after initialising, but this wasn't the case. I left in `return module` to make the function generic just so it can be used by others in the future, but I don't mind if you want it removed :+1:

app/assets/javascripts/surveys.js
AshGDS: Will do :+1: The call to the GA4LinkTracker is different across both places, so I've had to account for that!

AshGDS:
Thanks @andysellick, I've made the requested changes so it should be ready for another review. I also missed a step while explaining how to display the survey banner locally, so I've updated the PR description if that's useful.

app/assets/javascripts/surveys.js
andysellick: I think let's take it out unless we need it right now - it can always be added in again if required later.

app/assets/javascripts/surveys.js
andysellick, line 1: I've just thought of something - you might be able to save yourself a bit of effort here by letting the survey code initialise itself and then call `window.GOVUK.modules.start()`. It's fine to call that for a second time (after it gets called automatically on page load) because modules can't be initialised twice (thanks to the `-module-started='true'` attributes that get appended when the module is initialised on an element).

Sorry, should have thought of this sooner...

app/assets/javascripts/surveys.js
AshGDS: As it acts like a button as it closes the banner, we want to track it like a button - if we track it as a link, we would get values `url`, `link_path_parts`, `method`, and `link_domain` being sent to the dataLayer.

app/assets/javascripts/surveys.js
AshGDS: @andysellick No worries - will try this now - good to know! :+1:

app/assets/javascripts/surveys.js
andysellick: Ah of course, thanks for the explanation.

AshGDS:
Thanks @andysellick - should be good to review again :+1:

andysellick (approved):
👏

PR #3158, opened 2023-09-26
Add GA4 form tracker and section value to survey banner

AshGDS:
## What
- Adds the `section` value to the GA4 object for the survey banner events, with the value of `section` being the survey's heading text.
- Adds the GA4 form tracker to the email survey banner to track submitting the form.
- Adds a `form_complete` event using `Ga4AutoTracker` when the survey is sent to the user. This adds a `<span>` to the DOM with GA4 auto on it, rather than adding GA4 code to `surveys.js`. I thought this would be better as it prevents us adding a bunch of GA4 code in static which checks if GA4 modules exists, if cookie consent exists, and then more code to push the event. Instead we add a `<span>` with the auto tracker on it, then run `window.GOVUK.Modules.start()`.

## Why
- I missed out some tracking requirements in https://trello.com/c/udF07C9O/662-banners-survey-banner-element-visibility-navigation-close-user-interaction-and-form-events

## How to test

- Run a local version of static with this branch checked out
- In `surveys.js` go to the function `isSurveyToBeDisplayed` and set it to just return true, e.g.:
```JavaScript
 isSurveyToBeDisplayed: function (survey) { return true }
```
- In `gem_base.html.erb`, add the HTML element somewhere: 
```HTML
<div id="user-satisfaction-survey-container"></div>
```

### DIFF
--- a/app/assets/javascripts/surveys.js
+++ b/app/assets/javascripts/surveys.js
@@ -327,7 +327,8 @@
       // The module is then started - the module will send data-ga4-auto object to the dataLayer.
       // Doing it this way prevents us from calling and having to maintaining GA4 JS/cookie consent in this repo.
       var emailSurveyHeading = document.getElementById('survey-title').textContent.trim()
-      var emailSurveyText = error ? document.getElementById('email-survey-post-failure') : document.getElementById('#email-survey-post-success')
+
+      var emailSurveyText = error ? document.getElementById('email-survey-post-failure') : document.getElementById('email-survey-post-success')
       emailSurveyText = emailSurveyText.textContent.trim()
       var span = document.createElement('span')
       span.setAttribute('data-module', 'ga4-auto-tracker')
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/surveys.js
```diff
@@ -320,6 +322,27 @@
...
+      span.setAttribute('data-ga4-auto', JSON.stringify({
+        event_name: 'form_complete',
+        action: 'complete',
+        type: 'survey banner',
+        text: emailSurveyText,
+        section: emailSurveyHeading,
+        tool_name: emailSurveyHeading
+      }))
+      emailSurveyElement.appendChild(span)
+      window.GOVUK.modules.start(emailSurveyElement)
```
andysellick, line 344: This looks good but I'm wondering if we should think about our modules in this context a bit more. This isn't a criticism of the work you've done here, more something maybe for us to think about in general.

It feels slightly clumsy to create an element, attach attributes to it, then call start modules on it, when perhaps we could be calling the module directly and just pass it what it needs? Maybe a bit more like we do when we run tests...

```Javascript
var attributes = { 'ga4': 'attributes' }
var module = new GOVUK.Modules.Ga4AutoTracker({ attributes: attributes })
module.init()
```

andysellick (approved):

app/assets/javascripts/surveys.js
AshGDS: Thanks @andysellick - that's a good idea, I think we should implement that :+1:

PR #3170, opened 2023-10-12
Implement One Login "cross service header"

jon-kirwan:
## What

https://trello.com/c/9o9pvE3a/2146-implement-one-login-cross-service-header-on-the-email-subscriptions-service

## Why

> - To help us here in One Login improve the component
> - To provide email subscriptions service users with the navigation improvements we’re already confident about following our testing

## Anything else

- https://github.com/alphagov/email-alert-frontend/pull/1634
- https://github.com/alphagov/govuk_publishing_components/pull/3659

### DIFF
--- a/app/views/root/_gem_base.html.erb
+++ b/app/views/root/_gem_base.html.erb
@@ -8,7 +8,10 @@
   omit_global_banner ||= nil
   omit_account_navigation ||= nil
   product_name ||= nil
+  navigation_items ||= []
+  one_login_navigation_items ||= {}
   show_explore_header = show_explore_header === false ? false : true
+  show_cross_service_header ||= false
   show_account_layout ||= false
   account_nav_location ||= nil
   footer_meta ||= nil
@@ -37,6 +40,7 @@
 
 <%= render "govuk_publishing_components/components/layout_for_public", {
   account_nav_location: account_nav_location,
+  **(defined?(blue_bar) ? {blue_bar: blue_bar,} : {}),
   cookie_banner_data: {
     ga4_tracking: true,
   },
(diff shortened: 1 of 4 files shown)


app/views/root/_gem_base.html.erb
```diff
@@ -41,7 +40,7 @@
 
 <%= render "govuk_publishing_components/components/layout_for_public", {
   account_nav_location: account_nav_location,
-  blue_bar: blue_bar,
+  **(defined?(blue_bar) ? {blue_bar: blue_bar,} : {}),
```
MartinJJones, line 43: Not too familiar with this syntax, appears to be the double splat operator after looking into it. Just not 100% sure why it is needed here over the existing approach of `blue_bar ||= false` and any potential downsides of using it, for example what would happen if the value of blue_bar was:

```ERB
{
  enabled: true,
  size: 'l',
}`
```

MartinJJones (commented):
Changes look good to me overall, I just have one question regarding the approach for adding `blue_bar` to the template.

app/views/root/_gem_base.html.erb
jon-kirwan: Thanks @MartinJJones. Here is some more info:

- if `blue_bar` is passed in from `gem_base` it will cause design issues for some pages, for example, https://www.gov.uk/world/ukraine (which uses the `gem_layout` template) will have the blue bar removed
- I moved the `blue_bar` key/value pair to the `gem_layout_account_manager` template (with a default value of `false`) to be passed to `gem_base`
- `blue_bar` then requires a default value for all templates that don’t pass anything
- I tried setting this up in the usual way of `blue_bar ||= false`. However, I found some problems:
  - i’m not sure what the default value should be and i'm not sure that `true` is appropriate for some templates
  - assigning a default value of something like `nil` or `""` led to the blue bar being applied incorrectly due to logic in the public layout component (which checks `local_assigns` to see if a key/value pair is present
  - finally, passing a value of `false` assigns the default value https://ruby-doc.org/3.2.2/syntax/assignment_rdoc.html#label-Abbreviated+Assignment
- it seemed better to detect if a `blue_bar` key is being passed; assign the value if so, and do nothing if no key/value pair is present
- the `**` operator was one way I discovered to conditionally insert / or ignore a key/value pair .There's a few options discussed [here](https://stackoverflow.com/questions/5750770/conditional-key-value-in-a-ruby-hash)
- the [logic in the public layout component](https://github.com/alphagov/govuk_publishing_components/blob/main/app/views/govuk_publishing_components/components/_layout_for_public.html.erb#L6) correctly assigns the correct value if no `blue_bar` key is available.

jon-kirwan: Passing something similar to the example shouldn't cause an error (I think, it will apply it and attempt to pass it to the component, for example: `blue_bar: { enabled: true, size: 'l' }`). However, in the public layout component, blue_bar expects a true or false value. Anything that differs from this might cause an issue. Nevertheless, I don't anticipate encountering that problem unless changes are made in both places.

```
{
  enabled: true,
  size: 'l',
}
```

MartinJJones (commented):
Thanks for answering any questions I had, also tested on Integration by visiting a page on GOV.UK, viewing the Service Manual and using the Email Subscription Service.

Happy to approve this one once the changes required for testing are reverted in the Gemfiles

app/views/root/_gem_base.html.erb
jon-kirwan: > Yeah, agree with you on this one as well, there is a GitHub issue open relating to this - [alphagov/govuk_publishing_components#2472](https://github.com/alphagov/govuk_publishing_components/issues/2472)

Good find! Thanks for sharing the link.

MartinJJones (approved):

PR #3221, opened 2024-01-10
Remove emergency banner publishing rake tasks

brucebolt:
In https://github.com/alphagov/whitehall/pull/8709, we added a user interface to deploy and remove the emergency banner.

The documentation was updated in https://github.com/alphagov/govuk-developer-docs/pull/4385 to remove reference to these rake tasks.

Therefore they can be removed as they are no longer being used.


hannako (approved):
🥳 Top banana!

PR #3224, opened 2024-01-12
Update apple touch icons with the new versions

MartinJJones:
## What
Update the apple-touch-icons included in static

## Visual Changes

Te


hannalaakso (approved):
This looks like it should work 👍  Although as discussed with Martin, I'm still seeing the old icon when testing this via Browserstack with "Add to home screen" in iOS Safari. But it's possible that the old asset is being cached. We're going to test this again when it goes on integration which should be a more reliable indicator. 

For reference, we're removing the "GOV.UK" text from the icons to align with the GOV.UK Design System who are removing it in this context (see [this PR description](https://github.com/alphagov/govuk-frontend/pull/4354) for details).

MartinJJones:
> This looks like it should work 👍 Although as discussed with Martin, I'm still seeing the old icon when testing this via Browserstack with "Add to home screen" in iOS Safari. But it's possible that the old asset is being cached. We're going to test this again when it goes on integration which should be a more reliable indicator.

@hannalaakso just wanted to confirm that we are now seeing the apple touch icons on Integration, I've included some screenshots from Browserstack below:

| Device | Before | After |
| --- | --- | --- |
| iOS | iOS - before | iOS - after |
| Android | android - before | android - after |

However, the reason we are now seeing the icons on Integration is that we are currently using the test branch of the publishing components gem which includes 4.8 of the Design System, which is where the new icons are coming from.

The apple touch icons included in static appear to be used as a fallback to help avoid 404 errors for the favicon asset itself. I created a new Trello card to look at this further in the future, to review if the icons and routes are still required - https://trello.com/c/AqCgrXqK/42-review-and-improve-apple-touch-icon-redirects-used-in-static

hannalaakso (approved):
Thanks @MartinJJones  👍

PR #3339, opened 2024-05-31
Add GA4 tracking to global bar

AshGDS:
## What / Why
- Adds basic GA4 tracking to the global bar
- This tracking was missed in our GA4 team implementation work, most likely because this banner isn't documented in our component guide lists
- If you're unfamiliar with GA4:
    - We add a `data-ga4-link` JSON to links we want to track with some data we want tracked
    - We also initialise a `ga4-link-tracker` JavaScript module on these links
    - When the link is clicked, the `ga4-link-tracker` JS will use the JSON values, as well as create it's own values, which are then sent to GA4. Therefore we only need to add a small amount of templating data to get these links tracked.
    - We also add a `data-ga4-global-bar` attribute, which can be used in a future `govuk_publishing_components` gem release to track every time the banner is rendered
- Trello card: https://trello.com/b/rm8pjAAk/ga4-gtm-implementation-ga4-team-user-experience-govuk

### DIFF
--- a/app/views/components/_global_bar.html.erb
+++ b/app/views/components/_global_bar.html.erb
@@ -16,15 +16,22 @@
   title_classes = %w(global-bar-title)
   title_classes << "js-call-to-action" if title_href
   title_classes << "govuk-link" if title_href
+
+  ga4_data = {
+    event_name: "navigation",
+    type: "global bar",
+    section: title,
+  }.to_json
+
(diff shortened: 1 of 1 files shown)


richardTowers (approved):
Thanks for the explanation in the PR message! Looks good as far as I can see.

PR #3360, opened 2024-06-19
Remove Universal Analytics from static

AshGDS:
This includes some new code in surveys.js to start the code if usage consent is given, as the previous code would only start the survey code if the UA client id existed.


andysellick (approved):
Looks good to me 👏

AshGDS:
Thanks @andysellick

PR #3369, opened 2024-07-01
Timed updates

KludgeKML:
https://trello.com/c/upXxaeBR

## WHAT

Add in a helper method (before_update_time?) which can be used to make banner (or other layout) changes at a fixed time. Using this method rather than rolling a new one each time we have to make timed updates will prevent problems related to time zone settings (such as comparing a Time object - which does support time zones - with a Date object - which does not - causing an update to be an hour out during BST)
 
## WHY

Currently all deployments in banner text (non-emergency and perhaps emergency too) are manually triggered using GitHub Actions. Attempts to schedule banner updates using Ruby date functions (by merging the changes early), have caused issues with time-sensitive features due to server time differences from UK time (GMT/BST). A unique banner text deployment failed to update at the expected time because the server time did not match UK time. Automating the deployment to account for this time difference between server and UK time is essential to ensure timely updates without requiring human intervention at midnight.

## TESTING

We've deployed a version of this branch to integration with a nearer time, and observed that the banner vanished at the correct time. 

## RELATED PRS
[Update dev docs to point to new helper method](https://github.com/alphagov/govuk-developer-docs/pull/4716)

### DIFF
--- a/app/helpers/timed_update_helper.rb
+++ b/app/helpers/timed_update_helper.rb
@@ -1,0 +1,5 @@
+module TimedUpdateHelper
+  def before_update_time?(year:, month:, day:, hour:, minute:)
+    Time.zone.now.before? Time.zone.local(year, month, day, hour, minute)
+  end
+end
--- a/Gemfile
+++ b/Gemfile
@@ -36,5 +36,6 @@ group :test do
   gem "mocha", require: false
   gem "shoulda-context"
   gem "simplecov"
+  gem "timecop"
   gem "webmock"
 end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -570,6 +570,7 @@ GEM
     terser (1.2.3)
       execjs (>= 0.3.0, < 3)
     thor (1.3.1)
+    timecop (0.9.10)
     timeout (0.4.1)
     tzinfo (2.0.6)
       concurrent-ruby (~> 1.0)
@@ -616,6 +617,7 @@ DEPENDENCIES
   simplecov
   sprockets-rails
   terser
+  timecop
   webmock
 
 RUBY VERSION
--- a/app/views/components/_global_bar.html.erb
+++ b/app/views/components/_global_bar.html.erb
@@ -1,10 +1,10 @@
 <%
-  show_global_bar ||= true # Toggles the appearance of the global bar
-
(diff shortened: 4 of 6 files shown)


app/helpers/timed_update_helper.rb
```diff
@@ -0,0 +1,5 @@
+module TimedUpdateHelper
+  def before_update_time?(year:, month:, day:, hour:, minute:)
```
leenagupte, line 2: Minor: I'm not sure about this name, but I can't think of anything better!

leenagupte (commented):
Thanks for adding tests, and for testing the "switchover" time 🎉 

Is this intention for this helper to be long lived or to be deleted after the global banner is taken down?

Could copy some of the detail from the Trello card into the commit so that the detail isn't lost when the card is eventually archived? Thanks.

app/helpers/timed_update_helper.rb
KludgeKML: Yeah, I also couldn't think of anything better. Let's sleep on it and see if a better name turns up in the morning.

KludgeKML:
> Thanks for adding tests, and for testing the "switchover" time 🎉
> 
> Is this intention for this helper to be long lived or to be deleted after the global banner is taken down?

I think I'd say longer-lived? Like, for as long as we're doing it this way rather than it being a publishable object, we should encourage people to use this method rather than roll their own and risk forgetting about timezones.

> Could copy some of the detail from the Trello card into the commit so that the detail isn't lost when the card is eventually archived? Thanks.

Yeah, I'll do that in the morning.

leenagupte:
> I think I'd say longer-lived? Like, for as long as we're doing it this way rather than it being a publishable object, we should **encourage people to use this method** rather than roll their own and risk forgetting about timezones.

That's great. If it's not part of the trello ticket, then can [updating the docs](https://docs.publishing.service.gov.uk/manual/global-banner.html) be added to it?

KludgeKML:
> That's great. If it's not part of the trello ticket, then can [updating the docs](https://docs.publishing.service.gov.uk/manual/global-banner.html) be added to it?

It is in the trello card. I'll pop in an update PR for govuk-developer-docs.

leenagupte (approved):
Thanks for updating the description. ⭐

PR #3383, opened 2024-07-17
Drop support for legacy browsers

AshGDS:
## What / Why
- Use type=module for this application's JS file...though I'm not sure what this file is used for, since the gem loads the `application.js` for static in the header layout.
- To coincide with https://github.com/alphagov/govuk_publishing_components/pull/4111
- Has [DO NOT MERGE] as we want the above PR live on `static` first before this is merged.
- To accurately test this you will need to checkout a local version the `govuk_publishing_components` branch linked above
- We will combine `es6-components.js` again as a separate task unless you disagree with this
- Trello card: https://trello.com/c/Hoa45SMD/141-turn-off-javascript-in-legacy-browsers

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -150,7 +150,7 @@ GEM
     govuk_personalisation (0.16.0)
       plek (>= 1.9.0)
       rails (>= 6, < 8)
-    govuk_publishing_components (41.1.1)
+    govuk_publishing_components (42.0.0)
       govuk_app_config
       govuk_personalisation (>= 0.7.0)
       kramdown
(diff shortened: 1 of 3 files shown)


MartinJJones (commented):
The change itself looks good to me, and happy to approve, you're right though, application.js is coming from the gem and this no longer appears to be used in static.

- _javascript.html.erb was originally added here and used in base.html - https://github.com/alphagov/static/pull/193/commits/e3b8d88bfab35ab6f218e0ccd1e5a39895e73190
- base.html removed in 2022 as part of this PR - https://github.com/alphagov/static/pull/2663/files#diff-fd217c7e3df07524f4631938560e80cf3540d16d25025a1f8758cc1885a84ec9L350

It could be worth looking into if we can remove this template, if so this would mean that this change can be merged in advance and not something that is directly blocking dropping support for older browsers.

Happy to go with either approach, I can always create an issue to come back to look at removing this template if preferred?

AshGDS:
@MartinJJones Thanks, yeah maybe it's better to have it as a separate issue just so these PRs do not get too bloated 👍

MartinJJones (approved):

PR #3413, opened 2024-08-21
Remove unused layouts

KludgeKML:
- remove four unused layouts: chromeless, gem_layout_explore_header, gem_layout_full_width_explore_header, gem_layout_no_emergency_banner.
- remove omit_emergency_banner code only used by no_emergency_banner layout.
- update test with removal of chromeless

Note: I've confirmed there are no mentions of these layouts in alphagov, and no requests for their templates in the logs in the last 14 days (last 5 days for gem_layout_no_emergency_banner, because that was when I [removed the hack in collections](https://github.com/alphagov/collections/pull/3743) that asked for it).

### DIFF
--- a/app/controllers/root_controller.rb
+++ b/app/controllers/root_controller.rb
@@ -22,16 +22,12 @@ class RootController < ApplicationController
     502
     503
     504
-    chromeless
     gem_layout
     gem_layout_account_manager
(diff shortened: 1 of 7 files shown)


leenagupte (approved):
Nice bit of clean up! 🥇

PR #3423, opened 2024-08-30
Upgrade Rails to 7.2.1

CodeSonia:
## What

Upgrade Static application to Rails 7.2.1

[Trello card](https://trello.com/c/EfS9DHDw/2784-update-our-applications-to-rails-72-l)

## Why
To keep the Static application up-to-date with the latest Rails version, incorporating recent bug fixes and security updates.


config/environments/development.rb
```diff
@@ -46,7 +46,7 @@
   # config.i18n.raise_on_missing_translations = true
 
   # Annotate rendered view with file names.
-  # config.action_view.annotate_rendered_view_with_filenames = true
+  config.action_view.annotate_rendered_view_with_filenames = true
```
beccapearce, line 49: Not a blocker, but would be nice to call out why we're enabling this

beccapearce (approved):
Looks good!
One non-blocking comment :)

config/environments/development.rb
CodeSonia: Good point! I left it uncommented as part of the upgrade suggestion, but I can definitely comment it out as I can see it become too noisy when debugging. The idea was to make it easier for developers to debug by adding comments in HTML to show which view template was responsible for rendering a specific section of a page. 

But I could see that becoming annoying/cluttered. So I'll uncomment this now and I'll also deal with the merge conflict that I can see that came about!


Thanks Becca for spotting this :)

beccapearce: I think you can leave it in and just have this comment as the explanation :)

CodeSonia, line 49: I left this line of code uncommented as part of the upgrade suggestion. The idea was to make it easier for developers to debug by adding comments in HTML to show which view template was responsible for rendering a specific section of a page.

If this becomes too noisy, we could comment this line of code!

PR #3428, opened 2024-09-09
Remove custom apple touch icons and redirects in static

MartinJJones:
## What

- Remove custom apple touch icons included in static
  - `apple-touch-icon.png` (120x120px)
  - `apple-touch-icon-152x152.png`
  - `apple-touch-icon-167x167.png`
  - `apple-touch-icon-180x180.png`
- Remove redirects for apple-touch-icons
- Remove references to apple-touch-icons from `publishing_api.rake`
- Update tests for icon redirects
  - Now only references `favicon.ico`
  - Uncommented test that was previously broken when checking for apple touch icons. It appears that in order for the tests to pass, there needs to be an entry in both `routes.rb` and `publishing_api.rake`, for example `favicon.ico` and `apple-touch-icon.png` pass (referenced in both files), `apple-touch-icon-152x152` does not pass (only referenced in routes.rb)

[Trello card](https://trello.com/c/vU3MYqzJ/2835-remove-legacy-apple-touch-icon-routes-and-redirect-code-placeholder)

## Why

- This change should help make maintenance of the crown logo easier, no need to create custom icons in different sizes, add more routes or run a rake task to update references to the icons in the content store
- The current solution to redirecting the icons is a bit complex and not clear how to maintain it, nearly half of the icons referenced in routes.rb and publishing_api.rake resulted in a 404 error when tested
- The custom apple touch icons are not required in static as we use the apple-touch-icon provided by govuk-frontend (Design System) across GOV.UK via the `link` tag in [layout_for_public.html.erb](https://github.com/alphagov/govuk_publishing_components/blob/main/app/views/govuk_publishing_components/components/_layout_for_public.html.erb#L85)

## Further info

The apple touch icon is primarily used on the iOS/iPadOS home screen.

I tested this change on Staging on older iOS devices to check the apple-touch-icon still works as expected when adding the GOV.UK website to the home screen, screenshots below.

### iPhone 7 - iOS 10.3
iOS-10 3

### iPhone 8 - iOS 11.4
iOS-11 4


KludgeKML (approved):
LVGTM! (Looks Very Good To Me)

Excellent work!

hannako (approved):
Thank you Martin for your deep dive into the archives! 🏅

PR #3533, opened 2024-12-11
Upgrade Ruby to version 3.3.6

unoduetre:

## What

Upgrade all our apps to use Ruby 3.3.6.

## Why

We want to keep our software up-to-date. Ruby 3.3.6 was released in November 2024, so we should upgrade to use it.

[Trello card](https://trello.com/c/dSLPBXkz/3095-upgrade-our-apps-to-use-ruby-336-l)


unoduetre:
As static is a separate app, other apps using it do not need to use the same version of Ruby.

PR #3546, opened 2024-12-27
Rails 8 upgrade

unoduetre:

## Why

[Trello ticket](https://trello.com/c/PhOD07yF/3127-rails-8-upgrades)


config/environments/test.rb
```diff
@@ -11,24 +11,20 @@
   # While tests run files are not watched, reloading is not necessary.
   config.enable_reloading = false
 
-  # Turn false under Spring and add config.action_view.cache_template_loading = true.
```
unoduetre, line 14: It's not necessary as [it's the opposite of config.enable_reloading](https://guides.rubyonrails.org/configuring.html#config-cache-classes) defined above.

unoduetre:
Please, see the comments in https://github.com/alphagov/frontend/pull/4556

config/environments/development.rb
```diff
@@ -43,20 +41,17 @@
   config.active_support.disallowed_deprecation_warnings = []
 
   # Raises error for missing translations.
-  # config.i18n.raise_on_missing_translations = true
+  config.i18n.raise_on_missing_translations = true
```
unoduetre, line 44: For consistency with other apps. This is for the development environment only.

config/environments/development.rb
```diff
@@ -43,20 +41,17 @@
...
-  # Uncomment if you wish to allow Action Cable access from any origin.
-  # config.action_cable.disable_request_forgery_protection = true
-
-  # Raise error when a before_action's only/except options reference missing actions
+  # Raise error when a before_action's only/except options reference missing actions.
   config.action_controller.raise_on_missing_callback_actions = true
 
   # Apply autocorrection by RuboCop to files generated by `bin/rails generate`.
-  # config.generators.apply_rubocop_autocorrect_after_generate!
+  config.generators.apply_rubocop_autocorrect_after_generate!
```
unoduetre, line 56: For consistency with other apps. This is for the development environment only.

app/views/root/_gem_base.html.erb
```diff
@@ -46,7 +46,6 @@
   global_bar: global_bar,
   homepage: homepage_blue_navbar,
   logo_link: logo_link,
-  product_name: product_name,
```
unoduetre, line 49: This line is duplicated below, so I'm removing it. It caused a warning from Ruby.

config/environments/development.rb
hannako: Static doesn't have any translations, and isn't likely to get any, so I'm not sure there's much point to this change.

config/environments/production.rb
```diff
@@ -49,45 +44,40 @@
...
+  config.log_formatter = ::Logger::Formatter.new
+
   # "info" includes generic and useful information about system operation, but avoids logging too much
   # information to avoid inadvertent exposure of personally identifiable information (PII). If you
   # want to log everything, set the level to "debug".
   config.log_level = ENV.fetch("RAILS_LOG_LEVEL", "info")
 
-  # Use a different cache store in production.
+  # Prevent health checks from clogging up the logs.
+  config.silence_healthcheck_path = "/up"
```
hannako, line 65: Same comment as I made on Frontend, ie we're not using `/up`

hannako (commented):
This looks good. Have you tested on integration against a frontend app using Rails 8 and Rails 7 ?

config/environments/development.rb
unoduetre: Yes, but for the reasons of consistency and just in case those are added. It's only about the development env anyways, so doesn't impact production in any way. Setting this option to true makes it safer in case translations are added and there are no downsides, if they're not.

config/environments/production.rb
unoduetre: Fixed.

unoduetre:
> This looks good. Have you tested on integration against a frontend app using Rails 8 and Rails 7 ?

I let it run for a a day on integration and staging with frontend being upgraded to Rails 8 and other apps being Rails 7.

hannako (approved):

config/environments/development.rb
unoduetre: (Also as we plan to merge all the frontend apps in the future, I suppose it would make sense to make the config as similar as possible.)

PR #3570, opened 2025-01-27
Use banner gem for emergencies

KludgeKML:
Ultimately the point of the gem is to allow us to retire static, but it will be useful to replace banner handling in static with the gem's code so that we move all possible config into one place. This PR allows static to use the gem for emergency banner handling rather than its own code, global banner handling will be moved in in a subsequent PR.

Note that we're leaving in the integration tests for emergency banners for now, since they should just work (they don't care whether the banner is locally sourced), and they give us a little extra confidence.

https://trello.com/c/EOsNWrSu/294-update-static-to-use-the-gem-for-emergency-banner-handling

Tested on Integration, screenshots:

## Homepage
Screenshot 2025-01-27 at 17 00 30

## Browse page
Screenshot 2025-01-27 at 17 00 37


app/views/root/_gem_base.html.erb
```diff
@@ -41,7 +28,7 @@
   account_nav_location: account_nav_location,
   **(defined?(blue_bar) ? {blue_bar: blue_bar,} : {}),
   draft_watermark: draft_environment,
-  emergency_banner: emergency_banner.presence,
+  emergency_banner: render("govuk_web_banners/emergency_banner", homepage:),
```
andysellick, line 31: Just curious... how does the text for the banner get passed to it?

andysellick (approved):
Looks fine to me 👍

app/views/root/_gem_base.html.erb
KludgeKML: The emergency banner work by a shared Redis cluster. There's a UI in Whitehall that writes a record to the cluster (including the text), and when you call this partial, the partial makes a call to Redis to check if the record has any values. If yes, the partial renders the component. If not, the partial renders empty. . (call is cached for 1 minute so we're not constantly hammering the redis instance).

PR #3591, opened 2025-02-19
Remove unused homepage_blue_navbar

hannako:
[To simplify the rollout of the homepage redesign work](https://github.com/alphagov/static/commit/3af25dff72646ba4f8fd6cc6b635f759afc48eff
), `homepage_blue_navbar` was set to `false` on the old homepage template. Whilst it was set to `true` on the new (currently live) homepage. 

This field is safe to remove - it no longer does anything


MartinJJones (approved):
Nice work, the change looks good to me 👍

PR #3598, opened 2025-02-27
Remove `show_explore_header` from view templates

MartinJJones:
## What

Remove `show_explore_header` from view templates

[Trello](https://trello.com/c/qE79TFYj/3303-remove-legacy-header-variants-from-layoutforpublic)

## Why

Removed `show_explore_header` from the view templates as this option was removed from the `layout_for_public` component the `govuk_publishing_components` gem.

Depends on:
- [x] https://github.com/alphagov/govuk_publishing_components/pull/4643

### Further info

It is not possible to remove [layout-header.scss](https://github.com/alphagov/static/blob/main/app/assets/stylesheets/application.scss#L29) from static as the super navigation menu relies on the logo styles provided by the design system.

It may be possible to remove [layout-header.js](https://github.com/alphagov/static/blob/main/app/assets/javascripts/application.js#L20) from static, however this will likely require further testing to ensure it does not break anything, I will create a separate card for this.

### DIFF
--- a/app/views/root/_gem_base.html.erb
+++ b/app/views/root/_gem_base.html.erb
@@ -9,7 +9,6 @@
   product_name ||= nil
   navigation_items ||= []
   one_login_navigation_items ||= {}
-  show_explore_header = show_explore_header === false ? false : true
   show_cross_service_header ||= false
   show_account_layout ||= false
   account_nav_location ||= nil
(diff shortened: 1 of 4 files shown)


jon-kirwan (approved):
Looks good to me!

PR #3599, opened 2025-02-28
Remove user survey banners

jon-kirwan:
## What

Remove user survey banners

https://trello.com/c/4tJWLAuf/3333-remove-custom-user-survey-banners-from-static

## Why

Code simplification.
> The survey was removed nearly 3 years ago, but the code is still included on production. https://github.com/alphagov/static/pull/2772
Removing the code would provide a small performance improvement to our end users, it would also help avoid copying code to other applications when moving away from Slimmer / Static.


MartinJJones (approved):
Nice work, I deployed to Integration to test it out further and the changes look good to me.

It looks like the shipped CSS and JS size from static will also be quite a bit smaller following this change, could be worth adding a note to SpeedCurve once the change is live 👍

